PR #80, opened 2018-05-19
Implemented support for response files.

### LINKED ISSUE #30: Add support for response files

TomServo3k:
Fixes #30


JoeRobich:
I just wanted to take a second to thank everyone for response file support. In dotnet-format we went back and forth about whether we needed an additional configuration file to support users with complicated configurations. In the end we deciding that response files were sufficient for these scenarios and it has allowed us to keep things simple. Thanks so much for this feature.

PR #345, opened 2018-12-21
Support aliases of param in Xml doc

wojtpl2:
Thanks to this PR you can create aliases of `<param>` in XML doc. E. g.:

```xml
<param name="verbose" alias="v">Show verbose output</param>
```

Output for DragonFruit application now looks like:
```ini
C:\Work\command-line-api\samples\DragonFruit (support_param_aliases_in_xml_doc -> origin)
λ dotnet run -- -h
DragonFruit:
  DragonFruit simple example program

Usage:
  DragonFruit [options]

Options:
  -v, --verbose    Show verbose output
  --flavor         Which flavor to use
  --count          How many smoothies?
  --version        Display version information
```

I wanted to support multiple aliases but they generate a hint in Visual Studio:
image

### DIFF
--- a/samples/DragonFruit/Program.cs
+++ b/samples/DragonFruit/Program.cs
@@ -10,7 +10,7 @@ namespace DragonFruit
         /// <summary>
         /// DragonFruit simple example program
         /// </summary>
-        /// <param name="verbose">Show verbose output</param>
+        /// <param name="verbose" alias="v">Show verbose output</param>
         /// <param name="flavor">Which flavor to use</param>
         /// <param name="count">How many smoothies?</param>
         static int Main(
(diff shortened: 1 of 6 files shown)


KathleenDollard:
I'm sorry that we've been so slow responding. That's partially because extensibility has been a bit fluid. 

DragonFruit was a playground for ideas (thus the non-serious name) and many of those ideas will be either extensibility points or extensions, including the XML Comment extraction. 

I put up a PR last night that is a proposed extensibility story for the app model layer. This approach takes each concept we need (name, alias, description, options, subcommands, etc) and uses a strategy pattern to find results via multiple strategies. I haven't yet done XML Doc strategies, but I think that is where this makes sense. 

We are still working on layering, and it may make most sense to have all of the XML Doc stuff in an independent layer/package. 

That's a long way of saying - I don't think we should merge this, but I would like to work with you to fit this and other work you might be interested into our longer term thinking. Along those lines, I'd love your feedback on https://github.com/dotnet/command-line-api/pull/363

wojtpl2:
Thanks for the reply. I understand that this PR will be rejected. It was just a suggestion. 

I've read #363 an it looks great. Especially `DotnetHybrid`. I see that there is still a lot of work to do. I mean `System.CommandLine.JackFruit.Program`, connecting this with `csproj`, adding different strategies Xml Doc...

I would like to help you, but I don't want to duplicate tasks and work. This repo is young and everything changes dynamically. I see great potential in this project and I will be excited if I can help.

@KathleenDollard Do you have any task for me? It doesn't have to be easy 😄

PR #347, opened 2018-12-21
[WIP] extract rendering functionality to System.CommandLine.Rendering

jonsequitur:


src/System.CommandLine.Rendering/System.CommandLine.Rendering.csproj
```diff
@@ -0,0 +1,12 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
```
Keboo: Shouldn't this be `netstandard2.0`?

src/System.CommandLine.Rendering/SystemTerminal.cs
```diff
@@ -87,25 +58,11 @@ public void TryEnableVirtualTerminal()
             _virtualTerminalMode = VirtualTerminalMode.TryEnable();
         }
 
-        private void ResetConsole()
+        protected override void Dispose(bool disposing)
```
Keboo: Shouldn't we only dispose `_virtualTerminalMode` if `disposing` is true?

src/System.CommandLine/AssemblyInfo.cs
```diff
@@ -0,0 +1,7 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("System.CommandLine.Rendering")]
```
Keboo: Would love to try and restructure to remove this.

jonsequitur: Me too. I'm trying to figure that out now, since our build tools make it awkward to make this work both in CI (where there's a strong name) and locally (where there is not).

`ITerminal` seems like it potentially belongs in `System.CommandLine` but it still needs trimming down so that it doesn't drag other things along. Thoughts?

src/System.CommandLine.Rendering/SystemTerminal.cs
jonsequitur: Whether this is called via an explicit `Dispose` or finalizer, we want to make sure that the virtual terminal mode gets reset.

Keboo: I don't believe that is correct. There is not defined order that the GC will clean up managed objects. So if this is invoked from a finalizer the `_virtualTerminalMode` may have already been GC'd. In fact, looking at the code closer, this is probably an indication that we should implement the Disposable pattern on `VirtualTerminalMode` as well.

jonsequitur: Good point. I'll move the finalizer to `VirtualTerminalMode`.

jonsequitur:
@dotnet-bot test Ubuntu16.04 Release please

PR #355, opened 2018-12-29
[WIP] move rendering implementations into ITerminal implementations

jonsequitur:


src/System.CommandLine.Rendering/ConsoleExtensions.cs
```diff
@@ -26,5 +26,31 @@ public static OutputMode DetectOutputMode(this IConsole console)
                 return OutputMode.File;
             }
         }
+
+        public static ITerminal TryGetTerminal(
```
Keboo: Given that this doesn't follow the normal Try* signature pattern, I would suggest simply calling it `GetTerminal`

src/System.CommandLine.Rendering/ConsoleExtensions.cs
```diff
@@ -26,5 +26,31 @@ public static OutputMode DetectOutputMode(this IConsole console)
...
+            }
+            else if (preferVirtualTerminal &&
+                     VirtualTerminalMode.TryEnable() is VirtualTerminalMode virtualTerminalMode &&
+                     virtualTerminalMode.IsEnabled)
+            {
+                return new VirtualTerminal(
+                    console,
+                    virtualTerminalMode);
+            }
+            else
```
Keboo: `else`s not needed in this method

src/System.CommandLine.Rendering/AnsiRenderingSpanVisitor.cs
```diff
@@ -9,8 +9,8 @@ namespace System.CommandLine.Rendering
     internal class AnsiRenderingSpanVisitor : ContentRenderingSpanVisitor
     {
         public AnsiRenderingSpanVisitor(
-            IConsole console,
-            Region region) : base(console.Out, region)
+            IConsole terminal,
```
Keboo: Not sure this rename makes sense since the type is still `IConsole`

jonsequitur:
@dotnet test this please

Issue #358, opened 2019-01-02
Typo correction


wli3:
The following experience would be much better when there is a typo

```
λ t-rex --versoin
Do you mean "--version"? <-------------------------with a suggestion for typo
Unrecognized command or argument '--versoin'

t-rex:
  A command line testing tool for .NET

Usage:
  t-rex [options]

Options:
  --file                     .trx file(s) to parse
  --filter                   Only look at tests containing the specified text. "*" can be used as a wildcard.
  --format                   The format for the output. (Summary, JSON)
  --path                     Directory or directories to search for .trx files. Only the most recent .trx file in a given directory is
                             used.
  --hide-test-output         For failed tests, hide detailed test output. (Defaults to false.)
  --ansi-mode                Shows output formatted using ANSI sequences
  --virtual-terminal-mode    Shows output formatted using ANSI sequences
  --version                  Display version information
```

PR #360, opened 2019-01-03
WIP: IHelpBuilder injection

Keboo:
One possible solution to allow projects, such as System.CommandLine.Rendering, to provider alternate help output.


src/System.CommandLine/Command.cs
```diff
@@ -48,24 +46,6 @@ public class Command : Symbol, ICommand, ISuggestionSource, IEnumerable<Symbol>
...
-                helpBuilder = HelpBuilder;
-            }
-            else
-            {
-                helpBuilder = new HelpBuilder(console);
-            }
-
-            helpBuilder.Write(this);
-        }
-
```
jonsequitur: Moving this out of `Command` / `ICommand` makes very good sense.

src/System.CommandLine/CommandLineConfiguration.cs
```diff
@@ -20,7 +21,8 @@ public class CommandLineConfiguration
             bool enablePositionalOptions = false,
             ValidationMessages validationMessages = null,
             ResponseFileHandling responseFileHandling = ResponseFileHandling.ParseArgsAsLineSeparated,
-            IReadOnlyCollection<InvocationMiddleware> middlewarePipeline = null)
+            IReadOnlyCollection<InvocationMiddleware> middlewarePipeline = null,
+            Func<InvocationContext, IHelpBuilder> helpProvider = null)
```
jonsequitur: This should probably eventually be something a bit more specific, like an interface. @KathleenDollard is working on an app model including providers for this and other services.

Keboo: I think that would be a good idea. That is certainly pushing in the same direction that `Microsoft.Extensions.DependencyInjection.Abstractions` uses with factory interfaces that provide the actual services. 
I also considered that rather than making the creation of the `IHelpBuilder` lazy, it might be simpler to just take in an instance and pass both the `InvocationContext` to it rather than just the command.

jonsequitur: You can get to everything from the `InvocationContext` so it should add some flexibility relative to just passing the command. For example, things like this: https://github.com/dotnet/command-line-api/issues/358.

PR #363, opened 2019-01-04
WIP: Jackfruit - generalized AppModel approach

KathleenDollard:
This is a proposed design for app model creation. The intent is that this can support any AppModel someone can come up with. 

An AppModel is a set of orthogonal strategies - for example, how you retrieve help is entirely orthogonal to whether you get your option info from a type, method, or something else. All of these are simple functions that use the same infrastructure. 

The HybridModel in the test will give a general idea of how this works. 

The expectation is that the programmer can use defaults or add additional strategies. More basic strategies are needed, but I thought this enough to get started. 

I am very unsure of the right strategy for early bail out. I think on a non-IEnumerable the default should be to stop on first, and for an IEnumerable to union results from all strategies. But should a strategy be able to say "hey, stop after me." I put this in place, but it makes an awful tuple going everywhere and I'm no longer convinced it has sufficient value.


src/System.CommandLine/ParseResult.cs
```diff
@@ -72,7 +73,8 @@ private void AddImplicitOptionsAndCheckForErrors()
                         }
                     }
 
-                    if (command.Command.Argument.HasDefaultValue &&
+                    if (command.Command.Argument != null &&
```
jonsequitur: I believe we're disallowing or should disallow a null `Argument` property, because the distinction between that and `Argument.None` (which is the default) is unclear.

src/System.CommandLine/ParseResultExtensions.cs
```diff
@@ -9,6 +9,36 @@ namespace System.CommandLine
 {
     public static class ParseResultExtensions
     {
+        public static object GetDefaultValue(this Type type)
```
jonsequitur: This would overload the concept of "default value", which currently means the default for the argument, not for the type.

src/System.CommandLine/ParseResultExtensions.cs
```diff
@@ -9,6 +9,36 @@ namespace System.CommandLine
...
+                    var optionType = option.Argument.ArgumentType == null
+                                    ? typeof(bool)
+                                    : option.Argument.ArgumentType;
+                    return optionType.GetDefaultValue();
+                default:
+                    throw new InvalidOperationException("Internal: Unknown result type");
+            }
+        }
+
+        public static object GetValue(this ParseResult parseResult, Argument argument)
```
jonsequitur: I believe this should be called `GetValueOrDefault`. The naming is meant to follow familiar LINQ method conventions, e.g.:

* `Enumerable.FirstOrDefault` -> Returns the first item or the type default. 
* `Enumerable.First` -> Returns the first item or throws.

src/System.CommandLine.JackFruit.Tests/ReflectionBindingCommandHandlerTests.cs
```diff
@@ -0,0 +1,203 @@
...
+using System.IO;
+using System.CommandLine.Invocation;
+using System.Threading.Tasks;
+using System.Reflection;
+using System.CommandLine.JackFruit.Tests.MethodModel;
+using System.CommandLine.Builder;
+
+namespace System.CommandLine.JackFruit.Tests
+{
+    public class ReflectionCommandHandlerTests
```
jonsequitur: Class and type name don't match.

src/System.CommandLine.JackFruit.Tests/TestUtils.cs
```diff
@@ -0,0 +1,105 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using FluentAssertions;
+
+namespace System.CommandLine.JackFruit.Tests
+{
+    internal class TestUtils
```
jonsequitur: FluentAssertions lets you write methods like these more expressively as extension methods.

src/System.CommandLine.JackFruit/Binding.cs
```diff
@@ -0,0 +1,54 @@
...
+            ReflectionThing = reflectionThing;
+            ReturnType = returnType;
+        }
+
+        public object ReflectionThing { get; } // ParameterInfo, PropertyInfo or other
+        public Type ReturnType { get; }
+
+    }
+
+    public class SymbolBinding : BindingBase
```
jonsequitur: Separate files please.

src/System.CommandLine.JackFruit/Binding.cs
```diff
@@ -0,0 +1,54 @@
...
+        public ISymbolBase Symbol { get; }
+
+        public static SymbolBinding Create(ParameterInfo paramInfo, ISymbolBase optionOrArgument)
+            => new SymbolBinding(paramInfo, paramInfo.ParameterType, optionOrArgument);
+
+        public static SymbolBinding Create(PropertyInfo propertyInfo, ISymbolBase optionOrArgument)
+            => new SymbolBinding(propertyInfo, propertyInfo.PropertyType, optionOrArgument);
+    }
+
+    public class FuncBinding<TTarget> : BindingBase
```
jonsequitur: Separate files please.

src/System.CommandLine.JackFruit/Constants.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System.Reflection;
+
+namespace System.CommandLine.JackFruit.Reflection
+{
+    public class Constants
```
jonsequitur: This should be internal and probably declared closer to where it's used.

src/System.CommandLine.JackFruit/Implementations/Strategy.cs
```diff
@@ -0,0 +1,99 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+
+namespace System.CommandLine.JackFruit
```
jonsequitur: Split into separate files please.

src/System.CommandLine/ParseResult.cs
KathleenDollard: I'm fine with that and removed this check and tests pass. 

Do we need an issue to ensure that we never get this far, because it's a bad place for a user to crash? Or do you think that won't recur

src/System.CommandLine/ParseResultExtensions.cs
KathleenDollard: Understand the naming concern.  We could:

- rename it
- make it private where I need it 

I prefer the first because if someone needs this behavior (and we're tossing a lot of objects around with the reflection work) I'd like it done this way.

src/System.CommandLine.JackFruit.Tests/ReflectionBindingCommandHandlerTests.cs
```diff
@@ -0,0 +1,203 @@
...
+            command.Handler = handler;
+            var melonOption = new Option("melon", "", new Argument<string>());
+            var berryOption = new Option("berry", "", new Argument<bool>());
+            var mangoOption = new Option("mango", "", new Argument<int>());
+            var bananaOption = new Option("banana", "", new Argument<string>());
+            handler.AddBinding(SymbolBinding.Create(type.GetProperties()[0], melonOption));
+            handler.AddBinding(SymbolBinding.Create(type.GetProperties()[1], berryOption));
+            handler.AddBinding(SymbolBinding.Create(type.GetProperties()[2], mangoOption));
+            handler.AddBinding(SymbolBinding.Create(type.GetProperties()[3], bananaOption));
+            command.AddOptions(new Option[] { melonOption, berryOption, mangoOption, bananaOption });
```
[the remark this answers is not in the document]
KathleenDollard: I had not put the new test approach into these tests. That's done. We probably need more tests

src/System.CommandLine.JackFruit.Tests/TestUtils.cs
KathleenDollard: I haven't written them. Any samples?

jonsequitur: This is a somewhat complex one but it'll give you the idea: https://github.com/dotnet/command-line-api/blob/master/src/System.CommandLine.Tests/AssertionExtensions.cs

KathleenDollard: Thanks!

KathleenDollard:
Closing this as we are taking a different approach with Jon's "decomposed model binding PR"

A future PR will do AppModel work against the new binding.

PR #364, opened 2019-01-04
parse directives as key/value pairs

jonsequitur:
This changes parsed directives (`ParseResult.Directives`) from `IReadOnlyCollection<string>` to `IReadOnlyDictionary<string, string>`. The only allowed delimiter is `:`.  This is intended to enable things like this:

```console
> myApp [render=ansi] -x -y -z
```

Thoughts on the general approach and design?

### DIFF
--- a/src/System.CommandLine.Tests/DirectiveTests.cs
+++ b/src/System.CommandLine.Tests/DirectiveTests.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Collections.Generic;
 using FluentAssertions;
 using Xunit;
 
@@ -25,7 +26,7 @@ namespace System.CommandLine.Tests
 
             var result = option.Parse("[parse] -y");
 
-            result.Directives.Should().Contain("parse"); 
+            result.Directives.Keys.Should().Contain("parse");
             result.Tokens.Should().Contain("[parse]");
         }
 
@@ -36,7 +37,7 @@ namespace System.CommandLine.Tests
 
             var result = option.Parse("[parse] -y");
 
-            result.Directives.Should().Contain("parse");
+            result.Directives.Keys.Should().Contain("parse");
         }
 
         [Fact]
@@ -46,8 +47,8 @@ namespace System.CommandLine.Tests
 
             var result = option.Parse("[parse] [suggest] -y");
 
-            result.Directives.Should().Contain("parse");
-            result.Directives.Should().Contain("suggest"); 
+            result.Directives.Keys.Should().Contain("parse");
+            result.Directives.Keys.Should().Contain("suggest");
         }
 
         [Fact]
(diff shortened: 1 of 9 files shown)


src/System.CommandLine.Tests/DirectiveTests.cs
```diff
@@ -57,7 +58,61 @@ public void Directives_must_be_the_first_argument()
...
             var result = option.Parse("-y [suggest]");
 
-            result.UnmatchedTokens.Should().Contain("[suggest]"); 
+            result.UnmatchedTokens.Should().Contain("[suggest]");
+        }
+
+        [Theory]
+        [InlineData("key:value", "key", "value")]
+        [InlineData("key:value:more", "key", "value:more")]
+        [InlineData("key:", "key", "")]
```
Keboo: Minor thing, but I think it would be easier to follow if we left the braces as part of the inline data. Then it would match the other unit tests in this file as well.

src/System.CommandLine/Parser.cs
```diff
@@ -30,7 +30,7 @@ public Parser() : this(new RootCommand())
         {
             var rawTokens = arguments;  // allow a more user-friendly name for callers of Parse
             var lexResult = NormalizeRootCommand(rawTokens).Lex(Configuration);
-            var directives = new List<string>();
+            var directives = new Dictionary<string, string>();
```
Keboo: I don't think we want to use a `Dictionary` here. The order of items when enumerating a dictionary is explicitly undefined, so exposing them in this way causes the order to be lost. Perhaps keep it as a `List<KeyValue<,>>`?

jonsequitur: Interesting point. I realize that my assumption is that order should not be significant for directives. Did you have a use case in mind for preserving order?

(Related, I think that the token collections on the `ParseResult` do preserve order, though I'm not sure this is tested.)

Keboo: I am not sure that there is a use case with the order current directives in core. I am thinking of if someone puts in their own directives that have an ordering dependency. This is certainly an edge case.

Keboo: Agreed. I am less concerned about the middleware, are more just noting that anyone manually processing the `ParseResult` will (to some degree since as you point out your could go back to the Tokens to look up order) lose access to the order they were specified.

thargy:
Why not use OrderedDictionary under the hood and support both an IReadOnlyList and IReadOnlyDictionary? Or just a Dictionary of a struct that includes the string and ordinal index, then use Values.OrderBy to expose the existing enumeration. Or use the List of KVPs and create the dictionary dynamically?

src/System.CommandLine/Parser.cs
jeredm: I agree with Kevin that the order could matter if somebody adds their own directives. 

@jonsequitur What was your thought on the value the dictionary provides? The directives are parsed once, so I don't see a time trade off. `OrderdDictionary` could work, if there is another reason. Duplicate directives that contradict each other (e.g. render=ansi and render=notthat) would be processed twice in a collection, but in general I don't think we need to handle duplication.

jonsequitur: @jeredm Thanks! I just realized duplicates aren't tested, and will result in an exception with this change.

jonsequitur: It's come up a few times that directives with values would be useful, e.g. for render mode hints. It makes sense for the parsing rules for these values to be consistent across directives. The dictionary was the simplest choice. 

With a dictionary, one can test for presence of a directive independently from a value, if the handler cares about the value. Maybe a custom type would be better? I opted for strings because binding to other types get us into much more complex territory (validation, arity, etc.), although the argument could be made that directives should be derived from `ISymbol` and support all of this as well as completions.

jonsequitur:
@thargy Good suggestions, though I do think we want this property to be ordering-agnostic. Most handling of the `ParseResult` shouldn't depend on order. I added tests though to pin down the expectation that `ParseResult.Tokens` does maintain the original order of the arguments on the command line if someone needs it.

Given that, any concerns with `IReadOnlyDictionary<string, string>`?

Keboo:
@jonsequitur no objections from me.

KathleenDollard:
Are we certain we will never want directives to have multiple values? 

[Something=Json] [Something=Xml]

Any list approach with a complex item and appropriate hash/equals would handle this. Any dictionary approach would not. 

It's possible KeyValuePair hashes the Key because of its intended use. If that is the case, we should not use KeyValuePair.

The underlying idea of allowing the user to supply limited data to directives seems a good one.

thargy:
I’m with @kathleendollard on this one.  It’s not like these lists will ever be that long so a full scan wouldn’t be that slow.  I would err on allowing total flexibility and an enumeration wouldn’t worry about duplicate keys.

jonsequitur:
This is a great point and brings arity into the picture despite my attempt to avoid it :)

Some command line tools allow you to "overwrite" an option by respecifying it:

```console
> myapp -a 1 -a 2    # -a == 2
```
Others aggregate the values:
```console
> myapp -a 1 -a 2    # -a == [1,2]
```
The latter is intended to make programmatic composition of command lines more deterministic by removing some ordering dependencies.

It makes sense that we could allow both for options based on argument arity. (We don't currently do this. We aggregate and if you specify an option more than once when only one is allowed, we produce a parse error.)

The approach I took with directives is different from what we do with options. What @KathleenDollard  is suggesting though is how ASP.NET Core models headers. Directives are analogous to headers in several ways. This might be more familiar and flexible.

jeredm:
Not sure how we are tracking this, but a documentation dependency is here: https://github.com/dotnet/command-line-api/wiki/Syntax-Concepts-and-Parser

PR #369, opened 2019-01-12
Two bug fixes

jonsequitur:
This fixes two bugs:

* When a custom argument converter returned a failure, the default value was not being used.
* A custom converter on an option argument was accepting additional arguments after a failure, preventing these arguments from being accepted by the parent command.


src/System.CommandLine/Symbol.cs
```diff
@@ -68,15 +68,10 @@ public virtual string Name
             }
         }
 
-        public Command Parent { get; private set; }
+        public Command Parent { get; protected internal set; }
```
Keboo: Do we really want to expose the setter for this outside of core? Is there a use case for needing it yet?

Issue #366, opened 2019-01-07
make Argument.Name required 


jonsequitur:
Argument names should be required. They are now used for binding (#344) and will be used eventually for disambiguating positional arguments (#310).

This will also solve #362.

Issue #362, opened 2019-01-03
Auto generated usage text doesn't mention mandatory argument


rchande:
I created a command like so:
```
Command thing()
            {
                var run = new Command("thing", "Try a thing", argument: new Argument<string>());

                run.Handler = CommandHandler.Create<string, IConsole>((string thing, IConsole console) =>
                {
                    if (string.IsNullOrEmpty(thing))
                    {
                        console.Out.WriteLine("nothing");
                    }
                    else
                    {
                        console.Out.WriteLine(thing);
                    }
                });

                return run;
            }
```

If I dotnet run my app without providing an argument, the output is:
```
Required argument missing for command: thing

thing:
  Try a thing

Usage:
  MLS.Agent thing
```

Note that the usage text doesn't actually mention the missing argument. I would have expected something like

```
MLS.Agent thing [arg]
```

rchande:
cc @jonsequitur

jeredm:
@rchande Can you please include the code for your Main method? I am not sure if this is a bug because the argument is not given a name. If you set the Name property when you instantiate the Argument object, help should display the name of the argument. I am asking for more info because my output looks different from yours and I am not sure if there is a change to help that is causing this difference. 

My help looks like this when I name the argument "hi":
```
Required command was not provided.

Usage:
  DotnetMetal [command]

Commands:
  thing <hi>    Try a thing
```

Here is the related code that makes me think the argument needs to be named:
https://github.com/dotnet/command-line-api/blob/5bace63365b71ce0c3001d50ec417e2201b37035/src/System.CommandLine/Help/Builders/HelpBuilder.cs#L449-L462

jonsequitur:
I'm inclined to see this as a bug because it's confusing. The fact that there is an argument should be indicated in help even if it wasn't named.

That said, we might want to make argument names mandatory because they have a functional role in binding, which would solve this issue as well.

Issue #310, opened 2018-11-15
Allow a command to accept multiple distinct arguments


tmds:
Some commands take multiple distinct arguments:

```
git remote add [<options>] <name> <url>
ln <target> <link_name>
```

This is typically used for required parameters.
The user doesn't have to specify the name when providing these values.

`System.CommandLine` currently only supports a single argument.
This argument may be of arity 1+.
The above UX is not possible.

That single argument is made available as a `string[]`.
This is different from `Options`, where options are bound based on their name to handler argument parameters (related https://github.com/dotnet/System.CommandLine/issues/297). This makes it much less convenient to consume these arguments, than it is to consume `Options`.

Suggested feature:
- Support multiple arguments per command
  * Commands with have subcommands can have at most 1 arguments
  * Only the last argument can have arity > 1
  * Only the last argument can be optional
    ~ or relaxed version:  If an argument is optional, all arguments succeeding must also be optional
- Support naming arguments for handler binding (cfr Options)

jonsequitur:
Yes, commands and options can both currently specify how many arguments are required, via argument arity, and a default value satisfies the requirement if the user doesn't specify an argument. (Help output doesn't reflect this currently.) I would expect this behavior to extend to this scenario. 

One consideration for the API design is we don't vary behavior based on the order in which child objects are added to their parents. Since the parsing of these arguments will necessarily be positional, the API will need to be explicit about their position.

KathleenDollard:
@jnm2 

It seems to me that the difference between options and arguments is whether the argument name (`--database`) is included. I assume in the example you posted, you'd have an opinion on which is better for end users. When there are multiple arguments, they need to be positional. 

I think this would be a valuable addition. 

And when @jonsequitur is back, I think we need a separate conversation about where/how required values are supported.

jnm2:
@KathleenDollard Yes, that is exactly how I was thinking of it. Thanks!

jonsequitur:
@jnm2 Your API suggestion was a good one. Thoughts? https://github.com/dotnet/command-line-api/pull/542/files#diff-fb1621b467c13fb6006c42c063e47bc2.

jonsequitur:
I've opened a separate issue to cover related help improvements: #543

jnm2:
@jonsequitur Wow... that's a lot to get my head around, but I'm glad my suggestion was useful and I'll look forward to future versions!

gitfool:
[...]

PR #371, opened 2019-01-14
Require argument name

jeredm:
Arguments must have a name when there is an arity greater than zero. This assists the help system with ensuring there are no unnamed mandatory arguments.

Fixes https://github.com/dotnet/command-line-api/issues/366

We may want to change `Argument{T}` to include a constructor that includes a `Name` to make to make creating an `Argument` easier. This would change would conflict with the some of the defaulting we have in the `Argument{T}` constructor.


src/System.CommandLine/CommandLineConfiguration.cs
```diff
@@ -36,6 +36,17 @@ public class CommandLineConfiguration
 
             foreach (var symbol in symbols)
             {
+                foreach (var childSymbol in symbol.Children)
```
Keboo: I don't believe this is correct. The `symbols` [that are passed in](https://github.com/dotnet/command-line-api/blob/master/src/System.CommandLine/Builder/CommandLineBuilder.cs#L33) will likely just be the root command. The `Children` will only be its direct children. We need to iterate down through all of them to do the validation.

jeredm: @Keboo Just to clarify this, starting at `symbol.Children` is correct. We just need to make sure to recursively check the children below.

jeredm:
I agree that naming the argument after the parent symbol could be confusing. For example, '--name <name>    Specifies the name option' doesn't inform me what argument I am supposed to supply to name. Let me know how you guys want to proceed. I can make the fix the @Keboo proposed above and then add a default "argument" with a test. We will need to decide what to do about all of the changes to the unit tests. I don't think it provides value by having to set `Name` everywhere and may confuse the tests purpose.

src/System.CommandLine/CommandLineConfiguration.cs
Keboo: Correct, just needs to be recursive.

KathleenDollard:
In help, parameters are capitalized, such that I prefer the first to the second here:

```bash
Usage:
> foo --name <NAME>
> foo --name <ARGUMENT>
```

I think people should always give a name, and we should make that quite easy. 

In addition to help, we also may use the name in binding. In this case, the name is important only when the argument appears on the command, and primarily if/after we allow multiple/positional arguments.

jonsequitur:
If we force people to give a name, then it will end up being repeated information most of the time. It's interesting that it's only used in binding for command arguments. This makes me wonder if we should adjust the interface a bit, also taking into account #310, which also applies only to commands.

jeredm:
@jonsequitur @Keboo The check for a required `Argument.Name` is now recursive. I am waiting on a decision with how/if we want to provide a default name to finish the rest. I am also good if we close #366 and create a new issue for providing a default.

jonsequitur:
I'd prefer to provide a default for now in order to undo some of the test churn here, which feels like unnecessary ceremony.

I agree with Kathleen that defaulting to the parent option name is best. This also makes the binding, given the current convention, less ambiguous, since the argument name and option name will be the same, which binds by default to a parameter having a matching name.

jeredm:
This should fix #366 and #362 . Symbols will now default argument names with the upper case alias where a arity larger than 0 when the argument is set.

src/System.CommandLine/CommandLineConfiguration.cs
```diff
@@ -38,6 +38,17 @@ public class CommandLineConfiguration
 
             foreach (var symbol in symbols)
             {
+                foreach (var childSymbol in ((ISymbolSet)symbol.Children).FlattenBreadthFirst(o => o.Children))
+                {
+                    if (childSymbol.Argument.Arity.MaximumNumberOfArguments != 0)
+                    {
+                        if (string.IsNullOrEmpty(childSymbol.Argument.Name))
+                        {
+                            throw new ArgumentException("Name must be set for command arguments.");
```
jonsequitur: This would be extra helpful if it specified which command's argument needs a name.

src/System.CommandLine/CommandLineConfiguration.cs
```diff
@@ -38,6 +38,19 @@ public class CommandLineConfiguration
...
             foreach (var symbol in symbols)
             {
+                foreach (var childSymbol in ((ISymbolSet)symbol.Children).FlattenBreadthFirst(o => o.Children))
+                {
+                    if (childSymbol.Argument.Arity.MaximumNumberOfArguments != 0)
+                    {
+                        if (string.IsNullOrEmpty(childSymbol.Argument.Name))
+                        {
+                            throw new ArgumentException(
+                                $"Name must be set for arguments with an arity above zero. "
```
Keboo: This string should be moved into `ValidationMessages`

src/System.CommandLine/CommandLineConfiguration.cs
```diff
@@ -38,6 +38,19 @@ public class CommandLineConfiguration
...
+                                + $"The argument missing a name has the alias '{childSymbol.Aliases.FirstOrDefault()}'.");
```
Keboo:
```suggestion
                                + $"The symbol with the argument missing a name has the alias '{childSymbol.Aliases.FirstOrDefault()}'.");
```

src/System.CommandLine/Symbol.cs
```diff
@@ -45,7 +46,21 @@ public abstract class Symbol : ISymbol
...
-        public Argument Argument { get; set; }
+        public Argument Argument 
+        { 
+            get
+            {
+                return _argument;
+            }
+            set
+            {
+                if (value != null && value.Arity.MaximumNumberOfArguments > 0 && string.IsNullOrEmpty(value.Name))
```
Keboo:
```suggestion
                if (value?.Arity.MaximumNumberOfArguments > 0 && string.IsNullOrEmpty(value.Name))
```

src/System.CommandLine/CommandLineConfiguration.cs
```diff
@@ -38,6 +38,19 @@ public class CommandLineConfiguration
 
             foreach (var symbol in symbols)
             {
+                foreach (var childSymbol in ((ISymbolSet)symbol.Children).FlattenBreadthFirst(o => o.Children))
+                {
+                    if (childSymbol.Argument.Arity.MaximumNumberOfArguments != 0)
+                    {
+                        if (string.IsNullOrEmpty(childSymbol.Argument.Name))
```
Keboo: These two if-statements could be combined to reduce nesting.

src/System.CommandLine/CommandLineConfiguration.cs
```diff
@@ -38,6 +38,19 @@ public class CommandLineConfiguration
 
             foreach (var symbol in symbols)
             {
+                foreach (var childSymbol in ((ISymbolSet)symbol.Children).FlattenBreadthFirst(o => o.Children))
```
Keboo: Why the `ISymbolSet` cast here? It should not be needed.

```suggestion
                foreach (var childSymbol in symbol.Children.FlattenBreadthFirst<Symbol>(o => o.Children))
```

src/System.CommandLine/ValidationMessages.cs
```diff
@@ -44,6 +44,10 @@ protected ValidationMessages()
                 ? $"Required argument missing for command: {symbolResult.Token}"
                 : $"Required argument missing for option: {symbolResult.Token}";
 
+        public virtual string RequiredArgumentNameMissing(string symbolAlias) =>
+            $"Name must be set for arguments with an arity above zero. "
+            + $"The symbol with the argument missing a name has the alias '{symbolAlias}'.";
```
jonsequitur: The string concatenation is unnecessary.

src/System.CommandLine/Symbol.cs
```diff
@@ -45,7 +46,21 @@ public abstract class Symbol : ISymbol
...
 
-        public Argument Argument { get; set; }
+        public Argument Argument 
+        { 
+            get
+            {
+                return _argument;
+            }
+            set
+            {
```
Keboo: One last thing I realized I missed previously. We have the assumption that the `Argument` will never be null. I believe we should check for `null` and in that case use `Argument.None` like what is done in the ctor.

src/System.CommandLine/Symbol.cs
```diff
@@ -58,7 +58,7 @@ public Argument Argument
                 {
                     value.Name = _aliases.First().ToUpper();
                 }
-                _argument = value; 
+                _argument = value ?? Argument.None;
```
jonsequitur: :+1:

PR #373, opened 2019-01-14
Add command to print complete script

wli3:
A small step to have auto complete script registration

### DIFF
--- a/src/System.CommandLine.Suggest.Tests/SuggestionShellScriptHandlerTest.cs
+++ b/src/System.CommandLine.Suggest.Tests/SuggestionShellScriptHandlerTest.cs
@@ -12,9 +12,9 @@ namespace System.CommandLine.Suggest.Tests
         public void When_shellName_does_not_support_it_should_print_error()
         {
             var console = new TestConsole();
-            Action a = () => SuggestionShellScriptHandler.Handle(console, "fish");
+            Action a = () => SuggestionShellScriptHandler.Handle(console, "notsupported");
             a.Should().Throw<SuggestionShellScriptException>().And.Message.Should()
-                .Contain("fish shell is not supported.");
+                .Contain("notsupported shell is not supported.");
         }
 
         [Fact]
--- a/src/System.CommandLine.Suggest.Tests/SuggestionShellScriptHandlerTest.cs
+++ b/src/System.CommandLine.Suggest.Tests/SuggestionShellScriptHandlerTest.cs
@@ -1,0 +1,36 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using FluentAssertions;
+using Xunit;
+
+namespace System.CommandLine.Suggest.Tests
+{
+    public class SuggestionShellScriptHandlerTest
+    {
+        [Fact]
+        public void When_shellName_does_not_support_it_should_print_error()
+        {
+            var console = new TestConsole();
+            Action a = () => SuggestionShellScriptHandler.Handle(console, "fish");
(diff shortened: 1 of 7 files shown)


src/System.CommandLine.Suggest/SuggestionShellScriptHandler.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        private static void PrintToConsoleFrom(IConsole console, string scriptName)
+        {
+            var assemblyLocation = Assembly.GetAssembly(typeof(SuggestionShellScriptHandler)).Location;
+            var directory = Path.GetDirectoryName(assemblyLocation);
+            console.Out.Write(File.ReadAllText(Path.Combine(directory, scriptName)));
+        }
+    }
+
+    public class SuggestionShellScriptException : Exception
```
jonsequitur: Separate file please.

src/System.CommandLine.Suggest.Tests/SuggestionShellScriptHandlerTest.cs
```diff
@@ -0,0 +1,36 @@
...
+{
+    public class SuggestionShellScriptHandlerTest
+    {
+        [Fact]
+        public void When_shellName_does_not_support_it_should_print_error()
+        {
+            var console = new TestConsole();
+            Action a = () => SuggestionShellScriptHandler.Handle(console, "fish");
+            a.Should().Throw<SuggestionShellScriptException>().And.Message.Should()
+                .Contain("fish shell is not supported.");
```
jonsequitur: This test will break when we add fish shell support.

wli3: change to something like "nonexistshell"?

wli3: changed to "notsupported"

src/System.CommandLine.Suggest/SuggestionShellScriptHandler.cs
wli3: separated

wli3:
@dotnet-bot  test Windows_NT Debug @dotnet-bot test Windows_NT Release

wli3:
Flaky tests?

wli3:
@jonsequitur finally passed

src/System.CommandLine.Suggest.Tests/SuggestionShellScriptHandlerTest.cs
```diff
@@ -0,0 +1,36 @@
...
+namespace System.CommandLine.Suggest.Tests
+{
+    public class SuggestionShellScriptHandlerTest
+    {
+        [Fact]
+        public void When_shellName_does_not_support_it_should_print_error()
+        {
+            var console = new TestConsole();
+            Action a = () => SuggestionShellScriptHandler.Handle(console, "notsupported");
+            a.Should().Throw<SuggestionShellScriptException>().And.Message.Should()
```
jonsequitur: This shows a stack trace in stderr, no?

jonsequitur: No, it's better to avoid using an exception for input validation or control flow. Writing the error to stderr is preferable.

wli3: I don't think if an exception go all the way up is considered control flow, catch it or show up in happy path is. Exception is a good abstraction for when something is wrong and the program cannot resolve by itself. Stop all and here is the reason why. It is especially useful for command line app, there is nothing to retry or revert.

Issue #370, opened 2019-01-13
Make CommandLineBuilderExtensions.GetTerminal() public


wjk:
Title says it all, really. My app needs to use a custom (non-System.CommandLine-compatible) argument parser. However, I want to use the `ITerminal` functionality for outputting ANSI-based color output sequences. However, `SystemConsoleTerminal` doesn't declare compatibility with ANSI sequences. `VirtualTerminal` does, but it's an internal class.

The function mentioned in the title does exactly what I need (auto-detects ANSI support, and returns an appropriately configured `ITerminal`), but I can't access it without reflection because it's private. I could send a PR for this if desired. Thanks!

jonsequitur:
I opened #374 which includes a change to move `GetTerminal` to `Terminal` and make it public. 

Just so you're aware, the API in `System.CommandLine.Rendering` is early and more unstable than `System.CommandLine.Experimental`. We'd appreciate your thoughts.

wjk:
I find `System.CommandLine.Rendering` to be very stable (albeit poorly documented). I don't understand how to use some of the more complex stuff, but the ANSI output works great.

One thing I would recommend adding, however, is adding a built-in way to get the ANSI codes for `StyleSpan` and `ColorSpan`. I tried using these as a way to store color/bold data using instance properties, but I had to implement translation to the static properties on the `Ansi` class by hand. Other than that, it looks great. Keep up the good work!

jonsequitur:
> One thing I would recommend adding, however, is adding a built-in way to get the ANSI codes for StyleSpan and ColorSpan.

Can you expand on this? Do you mean getting them from the terminal's current state?

wjk:
No. I mean I have to do something like this in my code to translate between `ColorSpan` instances and a code I can pass to `IStandardStreamWriter.Write()`:

```cs
public void SetForeground(ForegroundColorSpan color)
{
    if (!IsRedirected && OutputMode == OutputMode.Ansi)
    {
        if (color.RgbColor != null)
        {
            RgbColor rgb = color.RgbColor;
            Writer.Write(Ansi.Color.Foreground.Rgb(rgb.Red, rgb.Green, rgb.Blue).EscapeSequence);
        }
        else if (!string.IsNullOrEmpty(color.Name))
        {
            string name = color.Name;
            AnsiControlCode code;

            if (name == nameof(Ansi.Color.Foreground.Black)) code = Ansi.Color.Foreground.Black;
[...]

PR #374, opened 2019-01-14
start to differentiate VT vs non-VT behavior in TestTerminal

jonsequitur:
This addresses #370.

I also added `TestTerminal.VirtualTerminalMode` in order to start being able to better test ANSI vs non-ANSI output across terminal types.

### DIFF
--- a/src/System.CommandLine.Rendering/TestTerminal.cs
+++ b/src/System.CommandLine.Rendering/TestTerminal.cs
@@ -36,7 +36,7 @@ namespace System.CommandLine.Rendering
 
         private void OnCharWrittenToOut(char c)
         {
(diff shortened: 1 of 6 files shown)


src/System.CommandLine.Rendering/TestTerminal.cs
```diff
@@ -200,6 +207,8 @@ public virtual ConsoleColor ForegroundColor
             }
         }
 
+        public bool VirtualTerminalMode { get; set; } = true;
```
Keboo: Minor deal here, but the name `VirtualTerminalMode` does not really indicate a bool. Perhaps something like `IsVirtualTerminalModeEnabled`?

PR #375, opened 2019-01-14
Proposed .editorconfig change for var usage

Keboo:
@jonsequitur and @KathleenDollard are there any opinions about `var` usage. 
The default editor config that we put it suggests using var everywhere, even when the type is not apparent. I would suggest that we reverse that and only use `var` when the type is apparent. 

Thoughts?

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -63,7 +63,7 @@ dotnet_naming_symbols.constant_fields.required_modifiers          = const
 # var preferences 
 csharp_style_var_for_built_in_types = true:silent 
 csharp_style_var_when_type_is_apparent = true:silent 
-csharp_style_var_elsewhere = true:silent 
+csharp_style_var_elsewhere = false:silent 
 # Expression-bodied members 
 csharp_style_expression_bodied_methods = false:silent 
 csharp_style_expression_bodied_constructors = false:silent 


jonsequitur:
I have a very slight preference for `var` everywhere because type declarations for variables create a little bit of refactoring friction, but I'm ok with the change if people find that it improve readability.

jonsequitur:
@dotnet-bot test Ubuntu16.04 Debug

KathleenDollard:
I don't feel strongly about this, but I am fine with the changes. It's more familiar. 

I'm only concerned about the cleanup plan. The change will cause tons of merge conflicts if we change outside code we are active in.

KathleenDollard:
FWIW, what I recommend on stage is that people ONLY make these kinds of changes in chunks planned well ahead with a dire warning to have all code checked in before. At that point, you can make the change everywhere in a single commit.

tannergooding:
For reference: Only using `var` when it is "apparent" is the recommended coding style for the .NET repos. It is specified as rule 10 here: https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md

> We only use var when it's obvious what the variable type is (e.g. var stream = new FileStream(...) not var stream = OpenStandardInput()).

However, it is ultimately up to each repo and some do differ. Roslyn for example uses `var` everywhere: https://github.com/dotnet/roslyn/blob/master/.editorconfig#L133. Except for in the in compiler, where they only use it where apparent: https://github.com/dotnet/roslyn/blob/master/src/Compilers/.editorconfig#L8

PR #377, opened 2019-01-16
WIP: Adding typo correction.

### LINKED ISSUE #358: Typo correction

Keboo:
Seemed like a fun little thing to mock out. 
Opening this PR just to get feedback on if people want this finished or not.

Fixes #358

### DIFF
--- a/src/System.CommandLine/Invocation/InvocationExtensions.cs
+++ b/src/System.CommandLine/Invocation/InvocationExtensions.cs
@@ -153,7 +153,8 @@ namespace System.CommandLine.Invocation
         public static CommandLineBuilder UseParseDirective(
             this CommandLineBuilder builder)
         {
-            builder.AddMiddleware(async (context, next) => {
+            builder.AddMiddleware(async (context, next) =>
+            {
                 if (context.ParseResult.Directives.Contains("parse"))
                 {
                     context.InvocationResult = new ParseDirectiveResult();
@@ -170,7 +171,8 @@ namespace System.CommandLine.Invocation
         public static CommandLineBuilder UseSuggestDirective(
             this CommandLineBuilder builder)
         {
-            builder.AddMiddleware(async (context, next) => {
+            builder.AddMiddleware(async (context, next) =>
+            {
                 if (context.ParseResult.Directives.Contains("suggest"))
                 {
                     context.InvocationResult = new SuggestDirectiveResult();
@@ -179,6 +181,24 @@ namespace System.CommandLine.Invocation
                 {
                     await next(context);
                 }
+            }, CommandLineBuilder.MiddlewareOrder.Preprocessing);
+
+            return builder;
+        }
+
+        public static CommandLineBuilder UseTypoCorrections(
+            this CommandLineBuilder builder, int maxLevenshteinDistance = 3)
+        {
+            builder.AddMiddleware(async (context, next) =>
+            {
+                if (context.ParseResult.UnmatchedTokens.Any() &&
+                    context.ParseResult.CommandResult.Command.TreatUnmatchedTokensAsErrors)
+                {
+                    var typoCorrection = new TypoCorrection(maxLevenshteinDistance);
+                    
+                    typoCorrection.ProvideSuggestions(context.ParseResult, context.Console);
+                }
+                await next(context);
             }, CommandLineBuilder.MiddlewareOrder.Preprocessing);
 
             return builder;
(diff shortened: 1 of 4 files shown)


src/System.CommandLine/Invocation/InvocationExtensions.cs
```diff
@@ -184,6 +186,24 @@ public static CommandLineBuilder CancelOnProcessTermination(this CommandLineBuil
             return builder;
         }
 
+        public static CommandLineBuilder UseTypoCorrections(
+            this CommandLineBuilder builder, int maxLevenshteinDistance = 3)
```
jeredm: @Keboo Where did the default of 3 come from? I'm not saying it is a bad default. I am just curious.

Keboo: Mostly from [here](https://www.xkcd.com/221/) and the simplest thing to get working. There is probably some better heuristics to use rather than just picking a number out of a hat.
I considered using something like:
```C#
string unmatchedToken = ...;
int maxLevenshteinDistance = (int)Math.Ceiling(unmatchedToken.Length / 2.0);
```
Open to other suggestions too.

The other part I am not completely happy with in single character aliases. You can imaging have a command with options `-a`, `-b`, and `-c` and then the user types `-d`. All of those would show as suggestions since they all have the same distance.

src/System.CommandLine/Invocation/TypoCorrection.cs
```diff
@@ -0,0 +1,132 @@
...
+        public TypoCorrection(int maxLevenshteinDistance)
+        {
+            _maxLevenshteinDistance = maxLevenshteinDistance;
+        }
+
+        public void ProvideSuggestions(ParseResult result, IConsole console)
+        {
+            foreach (var token in result.UnmatchedTokens)
+            {
+                string suggestions = string.Join(", or ", GetPossibleTokens(result.CommandResult.Command, token).Select(x => $"'{x}'"));
```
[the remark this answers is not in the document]
Keboo: Completely agree. The message also needs to be moved out so it can be localized.

src/System.CommandLine/Invocation/TypoCorrection.cs
```diff
@@ -0,0 +1,132 @@
...
+
+namespace System.CommandLine.Invocation
+{
+    internal class TypoCorrection
+    {
+        private readonly int _maxLevenshteinDistance;
+
+        public TypoCorrection(int maxLevenshteinDistance)
+        {
+            _maxLevenshteinDistance = maxLevenshteinDistance;
```
colombod: should this check for >= 0 distance?

Keboo: Yes, assuming we keep this as an incoming parameter, that should be checked. Ideally it would be nice to eliminate the fixed value and use some heuristics based on the unmatched token.

src/System.CommandLine/Invocation/TypoCorrection.cs
```diff
@@ -0,0 +1,132 @@
...
+        }
+
+        public void ProvideSuggestions(ParseResult result, IConsole console)
+        {
+            foreach (var token in result.UnmatchedTokens)
+            {
+                string suggestions = string.Join(", or ", GetPossibleTokens(result.CommandResult.Command, token).Select(x => $"'{x}'"));
+                console.Out.Write($"'{token}' was not matched, did you mean {suggestions}?");
```
jonsequitur: ...matched. Did you mean...

src/System.CommandLine/Invocation/InvocationExtensions.cs
Keboo: Certainly. I think there is an outstanding question of what is expected. I think for single character tokens it might be reasonable to either not provide any suggestions or do something different (perhaps only return suggestions for things that start with the same character as the token). I am not really sure what would be expected or desired.

jonsequitur: I wonder if there should be more consistency of behavior here with tab suggestions. There, we do substring matching, which is a different kind of fuzzy matching. If someone types a single letter and hits tab, they get back all of the symbols containing that letter. That could work for typo suggestions as well.

Similarly, if someone types `myapp --nfo⇥`, maybe suggestions should allow tabbing through the typo-corrected variants?

In other words, is typo correction just a different interaction model for the existing `ISuggestionSource` abstractions, which should be expanded to consider misspellings in addition to the existing substring matching?

Issue #383, opened 2019-01-22
Multiple value of Option


wojtpl2:
I'm trying to switch [BenchmarkDotNet](https://github.com/dotnet/BenchmarkDotNet) (https://github.com/dotnet/BenchmarkDotNet/issues/1016) to `System.CommandLine`. 

But I have one problem. Currently BenchmarkDotNet uses multiple value of argument e.g.:
`dotnet run -c Release -- --runtimes net472 netcoreapp2.1` 
or 
`--filter *ClassA* *ClassB*`

Currently `System.CommandLine` doesn't support this. When I configure option:
```c#
   Option Filter() => new Option(new[] { "-r", "--runtimes" },
        "Full target framework moniker for .NET Core and .NET. For Mono just 'Mono', for CoreRT just 'CoreRT'. First one will be marked as baseline!",
        new Argument<string[]>());
```
I have to repeat an alias for each parameter value, like this:
`--runtimes net472 --runtimes netcoreapp2.1`

The worst scenario, instead:
`--filter *ClassA* *ClassB* --runtimes net472 netcoreapp2.1 --exporters GitHub StackOverflow RPlot CSV JSON HTML XML --attribute Attribute1 Attribute2` 
we should type this:
`--filter *ClassA* --filter *ClassB* --runtimes net472 --runtimes netcoreapp2.1 --exporters GitHub --exporters StackOverflow --exporters RPlot --exporters CSV --exporters JSON --exporters HTML --exporters XML --attribute Attribute1 --attribute Attribute2`
This example doesn't contain other attributes that can have one value. 

There is unit test that shows it doesn't work https://github.com/dotnet/command-line-api/blob/4f3b654f9d7116004d6c3f98ec20bf19c14e967d/src/System.CommandLine.Tests/ParserTests.cs#L570-L578

wojtpl2:
If you would like to I can prepare PR which will change this behavior.

jonsequitur:
I believe this is a duplicate of #368 and (with the difference of the delimiting character) #37.

If you would like to prepare a PR that would be great. 

Just as a heads-up, this is related to #310.

wojtpl2:
It is not duplicate. I don't want to split value. I want to pass many values with space between them. It is default behaviour of CommandLineParser library.

jonsequitur:
> I don't want to split value. I want to pass many values with space between them. 

[...]

PR #384, opened 2019-01-22
 Support multiple values of option without respecified option name.

wojtpl2:
Fix for #383

### DIFF
--- a/src/System.CommandLine.Tests/ParserTests.cs
+++ b/src/System.CommandLine.Tests/ParserTests.cs
@@ -406,7 +406,7 @@ namespace System.CommandLine.Tests
         }
 
         [Fact]
-        public void When_a_Parser_root_option_is_not_respecified_then_the_following_token_is_unmatched()
+        public void When_a_Parser_root_option_is_not_respecified_but_limit_is_not_reached_then_the_following_token_is_used_as_value()
         {
             var parser = new Parser(
                 new Option(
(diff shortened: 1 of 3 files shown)


wojtpl2:
In this PR I support:
`--option value1 value2` 
I support also arity of option.

src/System.CommandLine.Tests/ParserTests.cs
```diff
@@ -424,6 +424,43 @@ public void When_a_Parser_root_option_is_not_respecified_then_the_following_toke
...
+
+            result["vegetables"]
+                .Arguments
+                .Should()
+                .BeEquivalentTo("carrot");
+
+            result
+                .UnmatchedTokens
+                .Should()
+                .BeNullOrEmpty();
```
jonsequitur: This can be `BeEmpty()`.

src/System.CommandLine/SymbolResult.cs
```diff
@@ -80,13 +80,7 @@ internal SymbolResult TryTakeArgument(Token token)
             }
 
             if (!OptionWasRespecified)
-            {
-                if (Symbol is IOption)
-                {
-                    // Options must be respecified in order to accept additional arguments. This is not the case for command.
-                    return null;
```
jonsequitur: I love it when a PR just removes the code that made it not work. 🥇

PR #386, opened 2019-01-23
[debug] directive - resume automatically when debugger is attached

jonsequitur:


jonsequitur:
We could parameterize the directive to enable both scenarios.

Keboo:
Thoughts on making it an or condition. Block until a debuggers is attached or a key is pressed?

jonsequitur:
That could work although it might be a little tricky since `Console.ReadKey` isn't cancellable.

Keboo:
True, but i imagine some loop checking `Console.KeyAvailable` and only read if there is a key there would be what we want.

jonsequitur:
@dotnet-bot test this please

PR #387, opened 2019-01-24
Default command arg value bug fix

jonsequitur:
This fixes a bug where a default argument value on a command will overwrite the value specified on the command line (which is initially correctly parsed).


jonsequitur:

PR #390, opened 2019-01-25
Reflection binding

KathleenDollard:
This PR contains a new binder that builds on and combines the existing MethodBinder and TypeBinder. For comparison and to have this PR not be even more gigantic, work to remove classes that are replaced by the combined binder will be in a separate PR. This work is closely related to the AppModel/JackFruit work, but that will be in a future PR that accommodates the merged version of the ReflectionBinder.

Methods cannot exist without the context of a type (and an application) and thus it's logical to combine these two binders. 

In addition to supporting the upcoming AppModel/JackFruit and combining MethodBinder, ConstructorBinder and TypeBinder, goals for the work in this PR are:

* Provide an explicit and common binding system that can accommodate future binding sources and targets

* Recognizes and supports rich context for method invocation - setting of properties for example

* Provide alternate sources for property/parameter binding (arbitrary calculations like retrieving of environmental variables)

* Improve support for constructors with parameters

* Allow binding of symbols to parameters earlier in the process - to the parser, not via parse result (which allows the following)

* Make binding for the base pattern testable without running the handler method

```C#
    const string commandLine = "command --age 425 --name Gandalf";
    var command = new Command("command");
    command.AddOption(new Option("--name", "", new Argument<string>()));
    command.AddOption(new Option("--age", "", new Argument<int>()));
    var handler = CommandHandler.Create<string, int>((name, age) => { }, command);
    command.Handler = handler;

    // Example of testing the binding
    var arguments = handler.Binder
                    .GetInvocationArguments(GetInvocationContext(commandLine, command));
    arguments.Should().BeEquivalentSequenceTo("Gandalf", 425);
```

* Guarantee that the actual invocation of the method will not fail - all checks are complete and can provide good messages

* Allow for future checks of parameter/symbol binding like optionally ensuring symbols are only bound once (there are good reasons this will sometimes not be true)

Binding for construction generally occurs based on a delegate, and supports the context where the delegate will execute. Order of construction is:

1. Set static properties on the containing type
2. Gather constructor parameter values
3. Create instance
4. Set instance properties
5. Gather invocation parameters
6. Invoke method

Individual values are set if there is a higher precedence item that can contribute. Current precedence is:

1. Default for type (zero for integer)
2. Default from C# defaults (parameter default)
3. Value from binding (usually entered by user, but may be Symbol default, etc)


System.CommandLine.sln
```diff
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.27428.2037
```
jonsequitur: Please revert this.

samples/JackFruit/JackFruit.csproj
```diff
@@ -0,0 +1,17 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
jonsequitur: Did you mean to include this file?

src/System.CommandLine.Tests/CommandHandlerTests.cs
```diff
@@ -142,23 +143,42 @@ void Execute(string name, int Age)
...
+                //age.Should().Be(425);
+            }, command);
+
+
+            var arguments = handler.Binder
+                            .GetInvocationArguments(GetInvocationContext(commandLine, command));
+            arguments.Should().BeEquivalentSequenceTo("Gandalf", 425);
+
+            command.Handler = handler;
+            // Can't also call InvokeAsync because adding version a second time crashes. Probably fix as bug.
```
jonsequitur: This sounds like a significant bug that should be identified and fixed before merging.

src/System.CommandLine.Tests/CommandHandlerTests.cs
```diff
@@ -142,23 +143,42 @@ void Execute(string name, int Age)
...
-            command.Handler = CommandHandler.Create<string, int>((name, age) =>
+            var handler = CommandHandler.Create<string, int>((name, age) =>
             {
-                wasCalled = true;
-                name.Should().Be("Gandalf");
-                age.Should().Be(425);
-            });
-
-            await command.InvokeAsync("command --age 425 --name Gandalf", _console);
+                //wasCalled = true;
```
jonsequitur: This change means this test is testing something fairly different than it was before. I think the old test is still valid, as it's testing invocation with binding along the way. We still need invocation tests even if the new binding approach makes binding-only tests more viable.

Let's roll back this change and put your version in a new test class focused on binding.

src/System.CommandLine/Binding/Binding.cs
```diff
@@ -0,0 +1,38 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.CommandLine.Invocation;
+
+namespace System.CommandLine.Binding
+{
+
+    public delegate void BindingSetter(InvocationContext context, object target, object value);
```
jonsequitur: Our delegate naming convention (for better or worse) has so far been to make them more method-like, e.g. `Suggest`, `ConvertArgument`, and `ValidateSymbol`. Maybe `SetBinding` and `GetBinding`? Does that change the meaning?

src/System.CommandLine/Binding/BindingSet.cs
```diff
@@ -0,0 +1,46 @@
...
+using System.Collections.Generic;
+using System.CommandLine.Invocation;
+using System.Linq;
+
+namespace System.CommandLine.Binding
+{
+    // This should be named Binder, but that name was taken at present
+    public class BindingSet 
+    {
+        private List<Binding> _bindings { get; } = new List<Binding>();
```
jonsequitur: This can be a field.

src/System.CommandLine/Binding/BindingSide.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace System.CommandLine.Binding
+{
+    public abstract class BindingSide
+    {
+        public BindingSide(BindingGetter get, BindingSetter set)
```
jonsequitur: Null checks please.

src/System.CommandLine/Binding/IBinder.cs
```diff
@@ -0,0 +1,15 @@
...
+using System.CommandLine.Invocation;
+
+namespace System.CommandLine.Binding
+{
+    public interface IBinder
+    {
+        object GetTarget(InvocationContext context);
+        void SetTarget(object target);
+        void AddBinding(Binding binding);
+        void AddBinding(BindingSide targetSide, BindingSide parserSide);
```
jonsequitur: Overloads on an interface can generally be reduced to the base case, with the other(s) becoming extension methods.

src/System.CommandLine/Binding/ServiceBindingSide.cs
```diff
@@ -0,0 +1,16 @@
...
+namespace System.CommandLine.Binding
+{
+    public class ServiceBindingSide : BindingSide
+    {
+        private ServiceBindingSide(BindingGetter getter, BindingSetter setter)
+           : base(getter, setter)
+        { }
+
+        public static ServiceBindingSide Create(Type serviceType)
+             => new ServiceBindingSide((c, t) => c.ServiceProvider.GetService(serviceType),  null);
```
jonsequitur: You should pass `t` rather than `serviceType` to the invocation inside the lambda.

src/System.CommandLine/IArgument.cs
```diff
@@ -3,16 +3,12 @@
 
 namespace System.CommandLine
 {
-    public interface IArgument
+    public interface IArgument : ISymbolBase
```
jonsequitur: This needs more discussion.

KathleenDollard: I think it is important that we have a "thing that can provide a value" type. I am extremely open on all other issues around it. Although the current design could uptake the remaining entirely independent (by having two BindingSides).

And this conversation can be related to the multiple argument conversation. 

I selected "ISymbolBase" as a way to call out this conversation, and I'm very open on the resolution. Would you like me to open an issue on this specific question as a venue for conversation?

jonsequitur: Yes, I think this deserves an issue.

System.CommandLine.sln
jonsequitur: People using different versions of Visual Studio will cause this to change. It's just noise in the diff.

KathleenDollard: Right. Completely agree on it going back. Just trying to figure out why it changed/how to avoid in the future.

src/System.CommandLine.Tests/CommandHandlerTests.cs
KathleenDollard: I copied these tests and have them in a format that tests more completely via Reflection Binding

src/System.CommandLine.Tests/Binding/DataOnlyUsageTests.cs
```diff
@@ -0,0 +1,154 @@
...
+using System.CommandLine.Binding;
+using FluentAssertions;
+using System.Linq;
+using System.Threading.Tasks;
+using Xunit;
+using System.CommandLine.Builder;
+
+namespace System.CommandLine.Tests.Binding
+{
+    // Copyright (c) .NET Foundation and contributors. All rights reserved.
```
wli3: nit: extra header

src/System.CommandLine/Binding/ServiceBindingSide.cs
KathleenDollard: I am renaming the lambda parameters because this wasn't clear. t is not the service type.

KathleenDollard: I did this the other places that had the (c,t) for the bind

KathleenDollard:
OK, fixed most, a few comments. Two known naming questions:

Delegates: BindingGetter/Setter not right. Bind/BindDefault questionable (on the default). Needs brainstorming here. 
Should what is currently called "target" be called "model"?

jonsequitur:
I think "model" is clearer than "target".

src/System.CommandLine.Tests/CommandHandlerTests.cs
```diff
@@ -110,7 +111,7 @@ void Execute(string name, int age)
             }
 
             var command = new Command("command");
-            command.AddOption(new Option("--name", argument: new Argument { Arity = ArgumentArity.ExactlyOne }));
+            command.AddOption(new Option("--name", argument: new Argument<string>() { Arity = ArgumentArity.ExactlyOne }));
```
jonsequitur: Please revert.

src/System.CommandLine/Binding/IBinder.cs
```diff
@@ -0,0 +1,14 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.CommandLine.Invocation;
+
+namespace System.CommandLine.Binding
+{
+    public interface IBinder
+    {
+        object GetTarget(InvocationContext context);
+        void SetTarget(object target);
```
jonsequitur: The implication here is that the `IBinder` is stateful. Is this the case?

src/System.CommandLine/Binding/ParameterCollection.cs
```diff
@@ -0,0 +1,55 @@
...
+            }
+            _arguments[pos] = value;
+        }
+
+        public object GetParameter(ParameterInfo parameterInfo)
+        {
+            var pos = Array.IndexOf(_parameters, parameterInfo);
+            if (pos < 0)
+            {
+                throw new InvalidOperationException("Unexpected Parameter");
```
jonsequitur: The exception will be more helpful if more contextual information (e.g. the `parameterInfo`) is provided.

src/System.CommandLine/Binding/ReflectionBinder.cs
```diff
@@ -0,0 +1,426 @@
...
+        private object _explicitlySetTarget;
+        private Type Type { get; }
+        private  MethodInfo InvocationMethodInfo { get;  set; }
+        // I really hate the location of these. I think we need a reflection binding set that incorporates these
+        private ParameterCollection InvocationParameterCollection { get;  set; }
+        private ParameterCollection ConstructorParameterCollection { get;  set; }
+
+        private readonly BindingSet ConstructorBindingSet = new BindingSet();
+        private readonly BindingSet InvocationBindingSet = new BindingSet();
+        private bool isBoundToCommand;
```
jonsequitur: For consistency, re: the above declarations, please use `_` prefixes and make them fields if they are `private`.

src/System.CommandLine/Binding/ReflectionBinder.cs
```diff
@@ -0,0 +1,426 @@
...
+            switch (source)
+            {
+                case ParameterInfo parameterInfo:
+                    AddBinding(parameterInfo, option);
+                    break;
+                case PropertyInfo propertyInfo:
+                    AddBinding(propertyInfo, option);
+                    break;
+                default:
+                    throw new InvalidOperationException("Unexpected source type");
```
jonsequitur: Please include the unexpected type in the exception message.

src/System.CommandLine/Binding/ReflectionBinder.cs
```diff
@@ -0,0 +1,426 @@
...
+            switch (source)
+            {
+                case ParameterInfo parameterInfo:
+                    AddBinding(parameterInfo, argument);
+                    break;
+                case PropertyInfo propertyInfo:
+                    AddBinding(propertyInfo, argument);
+                    break;
+                default:
+                    throw new InvalidOperationException("Unexpected source type");
```
jonsequitur: Please include the unexpected type in the exception message.

src/System.CommandLine/Binding/ReflectionBinder.cs
```diff
@@ -0,0 +1,426 @@
...
+            switch (source)
+            {
+                case ParameterInfo parameterInfo:
+                    AddBinding(parameterInfo, valueFunc);
+                    break;
+                case PropertyInfo propertyInfo:
+                    AddBinding(propertyInfo, valueFunc);
+                    break;
+                default:
+                    throw new InvalidOperationException("Unexpected source type");
```
jonsequitur: Please include the unexpected type in the exception message.

src/System.CommandLine/Binding/ReflectionBinder.cs
```diff
@@ -0,0 +1,426 @@
...
+                    InvocationBindingSet.AddBinding(ParameterBindingSide.Create(
+                                parameterInfo, InvocationParameterCollection), parserBindingSide);
+                    return;
+                case ConstructorInfo constructorInfo:
+                    ConstructorParameterCollection = ConstructorParameterCollection
+                                            ?? new ParameterCollection(constructorInfo);
+                    ConstructorBindingSet.AddBinding(ParameterBindingSide.Create(parameterInfo, ConstructorParameterCollection), parserBindingSide);
+                    return;
+                default:
+                    throw new InvalidOperationException("Unexpected parameter location");
```
jonsequitur: Can this exception message be made more helpful?

src/System.CommandLine/Binding/ReflectionBinder.cs
```diff
@@ -0,0 +1,426 @@
...
+
+        private static (ICommand, ISymbolBase) FindMatchingSymbol(string name, ICommand command)
+        {
+            if (command.Argument.Name == name)
+            {
+                return (command, command.Argument);
+            }
+            var options = command
+                          .Children
+                          .OfType<Option>()
```
jonsequitur: Can this be `IOption`?

src/System.CommandLine/Invocation/InvocationContext.cs
```diff
@@ -14,6 +14,15 @@ public sealed class InvocationContext : IDisposable
         private readonly Lazy<IDictionary<string, object>> _items = new Lazy<IDictionary<string, object>>(() => new Dictionary<string, object>());
         private IConsole _console;
 
+        // This is a total hack until we figure out where to retrieve this information from 
+        public static IReadOnlyCollection<Type> AvailableServiceTypes
```
jonsequitur: Let's avoid making this public.

Issue #393, opened 2019-01-27
InvokeAsync crashes when there is a version option


KathleenDollard:
The version option may already exist in two cases:

- The author has a different version, but wants to use the rest of the defaults and doesn't want to think about it
- During testing, etc, an invocation context is created to retrieve arguments or types, and then invoking async creates another invocation context

It should not crash when the second option is added.

jonsequitur:
Can you provide a repro?

PR #394, opened 2019-01-27
Don't add version if it already exists

KathleenDollard:
Fixes [#393](https://github.com/dotnet/command-line-api/issues/393)


src/System.CommandLine/Builder/CommandLineBuilderExtensions.cs
```diff
@@ -46,6 +46,11 @@ public static class CommandLineBuilderExtensions
         public static CommandLineBuilder AddVersionOption(
             this CommandLineBuilder builder)
         {
+            if (builder.Command.Children.GetByAlias("--version") != null)
```
jonsequitur: I think I'd prefer to support this as a more general capability of the `CommandLineBuilder`. 

And just to be pedantic, if the method is going to be idempotent, maybe `UseVersionOption` is a better name?

KathleenDollard:
Thanks done

Issue #397, opened 2019-01-28
Add a generic extension helper method for AddOption


KathleenDollard:
I find we're writing this code a lot:

```c#
       var command = new Command("command");
            command.AddOption(
                new Option("--name",
                           argument: new Argument<string>()));
            command.AddOption(
                new Option("--age",
                           argument: new Argument<int>()));
            command.Handler = CommandHandler.Create<string, int>(Execute);
```

We can simplify it to the following with an extension method:

```c#
            var command = new Command("command");
            command.AddOption<string>("--name");
            var option = command.AddOption<int>("--age");
            option.Description = "My amazing age option";
           command.Handler = CommandHandler.Create<string, int>(Execute);
```

We can talk about what parameters - but I think we should start by simplifying the simple case. 

This differs from earlier work on the builder, because it writes directly to the Command per our new approach.

jonsequitur:
Reposting here from my PR comment:

I think this deserves more discussion. I like that it's more terse, but it implies the existence of an `Option<T>`. Does that then imply there should be a `Command<T>`? It might be confusing to have to dig into the `Option<T>.Argument` property to do things like set the arity or add a validator. But then digging into the `Command.Argument` to set validators that deal with multiple options is already there and is confusing.

A potential complication is the introduction of multiple arguments of different types. I've been noodling on:

```cs
var x = new Command("move");
x.AddArgument(new Argument<DirectoryInfo[]>());
x.AddArgument(new Argument<DirectoryInfo>());
```

Would you want types like `Command<T1,T2>` or `Option<T1, T2>` to be possible?

KathleenDollard:
Would the multiple arguments of different types apply to both options and commands?

Agree we should talk more about this. As this is strictly a convenience method, maybe there is a naming fix like "AddOptionForType<T>()"

For multiple arguments and multiple argument types, I wonder if we could leave the common case as now and create an aggregate argument that could support either a tuple (not named as reflection level) or a struct (prefer due to naming issues). This would push this research into a specific place, since it is not the common case.

jonsequitur:
> Would the multiple arguments of different types apply to both options and commands?

So far we're thinking just commands. I haven't found examples of CLIs that do this with options.

I like the tuple idea but reflection binding may hit a wall there. It will likely need to be strictly positional binding, which might be fine for #310.

PR #398, opened 2019-01-28
Generic add option method

KathleenDollard:
First cut at [#397](https://github.com/dotnet/command-line-api/issues/397)


src/System.CommandLine/CommandExtensions.cs
```diff
@@ -74,5 +74,13 @@ public static void ConfigureFrom(this Command command, IOptionBuilder optionBuil
             string commandLine,
             IReadOnlyCollection<char> delimiters = null) =>
             new Parser(command).Parse(commandLine);
+
+        public static Option AddOption<T>(this Command command, params string[] aliases)
```
jonsequitur: I think this deserves more discussion. I like that it's more terse, but it implies the existence of an `Option<T>`. Does that then imply there should be a `Command<T>`? It might be confusing to have to dig into the `Option<T>.Argument` property to do things like set the arity or add a validator. But then digging into the `Command.Argument` to set validators that deal with multiple options is already there and is confusing.

A potential complication is the introduction of multiple arguments of different types. I've been noodling on:

```cs
var x = new Command("move");
x.AddArgument(new Argument<DirectoryInfo[]>());
x.AddArgument(new Argument<DirectoryInfo>());
```

Would you want types like `Command<T1,T2>` or `Option<T1, T2>` to be possible?

PR #402, opened 2019-01-31
ParseResult.RawInput falls back to Environment.CommandLine

jonsequitur:


src/System.CommandLine.Tests/SuggestionTests.cs
```diff
@@ -144,7 +143,7 @@ public void When_one_option_has_been_specified_then_it_and_its_siblings_will_sti
         }
 
         [Fact]
-        public void When_a_subcommand_has_been_specified_then_its_siblings_will_not_be_suggested()
+        public void When_a_subcommand_has_been_specified_then_its_sibling_options_will_not_be_suggested()
```
Keboo: I am not sure this rename is correct. The options are not siblings to the specified subcommand. They are children of the siblings to the specified subcommand (nieces and nephews). Did you mean to have sibling options on the root command?

src/System.CommandLine/ParseResult.cs
```diff
@@ -48,7 +50,7 @@ public class ParseResult
 
         public IReadOnlyCollection<string> UnmatchedTokens { get; }
 
-        internal string RawInput { get; }
+        internal string RawInput => _rawInput ?? Environment.CommandLine;
```
Keboo: Is there a reason to do this check in the property's getter rather than just doing it once in the ctor? Is it possible for `Environment.CommandLine` to change during the life of the application?

jonsequitur: I made it lazy only because it might not be called. It's only used for suggestions.

Keboo: Gotcha. That makes sense.

jonsequitur:
@dotnet-bot test Windows_NT Release

PR #408, opened 2019-02-09
replace earlier binding infrastructure

jonsequitur:
This is attempt #3 at a generalized model binding infrastructure with lessons learned from the previous two iterations. If we think this is a good approach, this is intended to replace the following types and their related code:

* `TypeBinder`
* `MethodBinder`
* `ReflectionBinder`

This is intended to address the following scenarios in a way that is flexible enough to be leveraged within app models, e.g. DragonFruit:

* Binding input from various sources (`ParseResult`, service provider, arbitrary inputs such as environment variables, etc.) to arbitrary types (including constructor parameters, property setters, and handler parameters)
* Binding in the absence of invocation
* Easier testing of binding operations

It also shifts the code for building options based on a method signature, which DragonFruit uses, into `System.CommandLine.DragonFruit`. Some of this may later move under the app model work but as it's highly convention-based, we decided that it does not belong in the core library.


src/System.CommandLine/Binding/ModelBinder.cs
```diff
@@ -0,0 +1,65 @@
...
+        public ModelBinder(Type modelType) : this(ModelDescriptor.FromType(modelType))
+        {
+        }
+
+        public IReadOnlyCollection<BoundValue> GetConstructorArguments(BindingContext context)
+        {
+            return GetValues(context, ConstructorDescriptor.ParameterDescriptors);
+        }
+
+        public object CreateInstance(BindingContext context)
```
KathleenDollard: Nit, consider GetInstance

src/System.CommandLine/Binding/ModelBinder.cs
```diff
@@ -0,0 +1,65 @@
...
+            SetProperties(bindingContext, instance);
+        }
+
+        private ConstructorDescriptor ConstructorDescriptor => ModelDescriptor.ConstructorDescriptors.Single();
+
+        private void SetProperties(
+            BindingContext context,
+            object instance)
+        {
+            var boundValues = GetValues(
```
KathleenDollard: This looks like code that could get out of alignment. Can we consider getting a tuple here and using it below. It's the If skipping a value on line 58 that makes this code look problematic. 

Still thinking about GetValues vs. the calling code being responsible for GetValue.

src/System.CommandLine/Binding/ModelBinder{T}.cs
```diff
@@ -0,0 +1,41 @@
...
+        public void BindMemberFromOption<TValue>(
+            Expression<Func<TModel, TValue>> property,
+            IOption option)
+        {
+            NamedValueSources.Add(
+                property.MemberTypeAndName(),
+                new SpecificSymbolValueSource(option));
+        }
+
+        public void BindMemberFromCommand<TValue>(
```
KathleenDollard: Does this mean BindMemberFromArgument? How will a tight coupling between commands and arguments here work with multiple arguments? I do not believe people think about commands having values as much as having arguments.

src/System.CommandLine/Binding/ModelBinder{T}.cs
```diff
@@ -0,0 +1,41 @@
...
+
+namespace System.CommandLine.Binding
+{
+    public class ModelBinder<TModel> : ModelBinder
+    {
+        public ModelBinder() : base(ModelDescriptor.FromType<TModel>())
+        {
+        }
+
+        public void BindMemberFromOption<TValue>(
```
KathleenDollard: We need to be able to add options and arguments without a generic class, or we force any code doing generalized AppModel stuff to use ugly reflection hacks. This could be provided a couple of ways, but creating an Add on BindingModel (not generic) would be nice.

src/System.CommandLine/Binding/ModelDescriptor.cs
```diff
@@ -0,0 +1,49 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace System.CommandLine.Binding
+{
+    public class ModelDescriptor
```
KathleenDollard: AppModels will need to set just some properties. I think this will be a need for some other cases as well. How would you pass the list of PropertyDescriptors to this class?

KathleenDollard: won't ModelDescriptor need to be a ValueDescriptor for properties that are complex types?

src/System.CommandLine/Binding/SpecificSymbolValueSource.cs
```diff
@@ -0,0 +1,31 @@
...
+        public SpecificSymbolValueSource(ISymbol symbol)
+        {
+            Symbol = symbol;
+        }
+
+        public ISymbol Symbol { get; }
+
+        public bool TryGetValue(IValueDescriptor valueDescriptor, BindingContext bindingContext, out object value)
+        {
+            var optionResult = bindingContext.ParseResult.FindResultFor(Symbol);
```
KathleenDollard: I'm missing something here as I expected this to also handle command arguments.

src/System.CommandLine/Binding/SymbolBindingSide.cs
```diff
@@ -19,7 +19,7 @@ public static SymbolBindingSide Create(Option symbol)
         public static SymbolBindingSide Create(Argument argument)
             => new SymbolBindingSide(argument);
 
-        public ISymbolBase Symbol { get; }
+        public IValueDescriptor Symbol { get; }
```
KathleenDollard: Confused. It thought a ValueDescriptor was a parameter or property, not a symbol.

src/System.CommandLine/Binding/PropertyDescriptor.cs
```diff
@@ -0,0 +1,30 @@
...
+            _propertyInfo = propertyInfo;
+        }
+
+        public string Name => _propertyInfo.Name;
+
+        public Type Type => _propertyInfo.PropertyType;
+
+        public bool HasDefaultValue => false;
+
+        public object GetDefaultValue() => null;
```
KathleenDollard: People may be confused by this - properties now appear to have default values. Not sure if we can work with them.

src/System.CommandLine/Binding/ModelBinder{T}.cs
KathleenDollard: Ignore the above because this isn't the place for it, but to clarify the confusion. 

Our implementation is at odds with the ways I believe people think about the problem, and reflects an implementation detail. I do not believe people say "What's the value of that command?" Nor "What's the value of the argument on that option?" I think they would say "What's the value of the argument?" or "What's the value of the option?" and that is how Help seems laid out to me. I think this is going to be a rough cognitive load and reflects an implementation detail, and I know I will keep making this mistake whenever I am thinking about the real problem (the CLI). 

But this has nothing to do with the problem we are trying to solve today.

src/System.CommandLine/Binding/PropertyDescriptor.cs
KathleenDollard: Consider returning the type default instead of null.

jonsequitur: This isn't intended to being limited the capabilities of reflection. The intention is to be able to annotate the model descriptor so that, for example, if you know that a given type *behaviorally* has a default, you can define that in the model.

src/System.CommandLine/Binding/SymbolBindingSide.cs
jonsequitur: This is a refactoring artifact. This class is going to be removed in this PR in any case.

src/System.CommandLine/Binding/SpecificSymbolValueSource.cs
jonsequitur: It does. The variable is misnamed but it's operating on symbols generally.

src/System.CommandLine/Binding/ModelDescriptor.cs
jonsequitur: Possibly. Complex types will be build up using their constituent constructor arguments and property setters, so there may not be a need for this.

jonsequitur: Do you mean set some properties that were not discovered via reflection?

src/System.CommandLine/Binding/ModelBinder{T}.cs
jonsequitur: I'll add overloads for additional ways of configuring the bindings.

src/System.CommandLine/Binding/ModelBinder.cs
jonsequitur: Can you explain how you think about the distinction here?

src/System.CommandLine/Binding/ModelBinder.cs
jonsequitur: I'm not sure I follow. 

Would a cast to `PropertyDescriptor` rather than the `if` check address the concern?

src/System.CommandLine/Binding/PropertyDescriptor.cs
jonsequitur: Good catch.

src/System.CommandLine/Binding/ModelBinder.cs
KathleenDollard: On second look, Create is fine

src/System.CommandLine/Binding/ModelBinder{T}.cs
jonsequitur: It does mean bind from the argument. The concern is a good one. I think conceptually "What is the value for this option" is more intuitive but less precise than "What is the value for this option's argument"? And yes, the value for a command will become more ambiguous with #310.

jonsequitur:

jonsequitur:

PR #416, opened 2019-02-22
fix type-based arity assumption for command arguments with max arity > 1

jonsequitur:
The added test under `TypeConversionTests` was the repro for a bug by which the default arity for the type rather than the specified arity for the command argument was being used, resulting in an exception. This fixes it and tests the expected behavior.

### DIFF
--- a/src/System.CommandLine.Tests/Binding/TypeConversionTests.cs
+++ b/src/System.CommandLine.Tests/Binding/TypeConversionTests.cs
@@ -409,6 +409,33 @@ namespace System.CommandLine.Tests.Binding
                   .BeNull();
         }
 
+        [Theory]
+        [InlineData("c -a o c c")]
+        [InlineData("c c -a o c")]
+        [InlineData("c c c")]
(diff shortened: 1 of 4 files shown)


src/System.CommandLine.Tests/Binding/TypeConversionTests.cs
```diff
@@ -409,6 +409,33 @@ public void The_default_value_of_a_command_with_no_arguments_is_an_empty_collect
                   .BeNull();
         }
 
+        [Theory]
+        [InlineData("c -a o c c")]
+        [InlineData("c c -a o c")]
+        [InlineData("c c c")]
+        public void When_command_has_arity_greater_than_one_it_captures_arguments_before_and_after_option(string commandLine)
```
Keboo: "When_Command_argument_has_arity..."

jonsequitur: I'll get this in the next PR.

Issue #418, opened 2019-02-23
How to bind to nullable value type parameters


gitfool:
I'm enjoying using this library, despite it still being in the experimental stages, but I'm not sure how to deal with the latest changes.

Specifically, when [bumping](https://github.com/gitfool/BoardGameGeek.Dungeon/commit/6161e8de4d0f8d8e2aef69e13ea0c052a4f5c995) System.CommandLine.Experimental from 63807-01 to 63822-01, it no longer binds a nullable int parameter and it always gets passed as `0`, even though I specify it should default to a value. See the full source at [BoardGameGeek.Dungeon/Application/CommandLine.cs](https://github.com/gitfool/BoardGameGeek.Dungeon/blob/master/Application/CommandLine.cs).

If I change the year option to:
```c#
var yearOption = new Option(new[] { "-y", "--year" }, "Year to analyze.", new Argument<int?>());
```
I'd expect it to bind to `null` if not specified on the command-line, or the value if specified, but it still always gets passed as `0` in both cases, so the binding no longer works at all.

It does work again if I replace `int?` everywhere with `int`; but this used to work, albeit with no guarantees of stability, and it seems reasonable to want to bind to a nullable value type.

jonsequitur:
Thanks. This is a bug introduced by the latest binding refactor (which added support for binding to complex types, e.g. `CommandHandler.Create<MyComplexType>). I'm working on a fix for it.

jonsequitur:
This should be fixed now with #419.

gitfool:
Wow, confirmed as fixed! Thanks a lot, @jonsequitur!

Issue #399, opened 2019-01-28
Add Parent Command to Argument


KathleenDollard:
Options know the command they are attached to. 

Arguments do not. 

This is a problem for binding when a value may be bound to a parent command's argument. 

An example would be a parameter for an invoke method for `dotnet add package` where a ProjectName can be retrieved from the `add` command when the `package` command is invoked. This is a problem for both parameter and property binding. 

The work around would be to create a more complex binding that specified the command in either all cases or in cases where it wasn't the same command as invocation (properties are still aligned with that pattern at present. That seems like a lot more work than adding the parent to arguments. 

Which means arguments and options are alike in yet one more way :)

Reflection binding will not be able to bind values to parent commands until we have a plan for this. Since I don't think this is supported today in binding, I don't think this blocks the reflection binding PR. 

I'm happy to work on this, as long as it's the right direction, or let someone else step in.

PR #419, opened 2019-02-23
various binding improvements

jonsequitur:
There are a number of changes here:

* During binding, halt recursive binding of constructor parameters if the type can be bound from the `ParseResult`. This fixes a regression of the convention for single-string constructor parameter binding (e.g. `DirectoryInfo`).
* Fix a couple of bugs related to binding `Nullable<T>` (#418)
* When a `Command.Argument` has a nullable `ArgumentType`, default the arity to `ZeroOrOne`.
* Add property `Argument.Parent` (#399)


jonsequitur:

jonsequitur:

Issue #297, opened 2018-11-07
Make available the command arguments as a handler argument


tmds:
All parameters of the DragonFruit Main are currently added as Options.
Maybe there could be a way to make a parameter, the Argument of the Command.

jonsequitur:
Another approach that could work here would be for the binder to match properties or parameters by name using the argument name.

```csharp
var command = new Command("the-command", new Argument<bool> { Name = "x" } )
{
    new Option("-y", new Argument<string>("-y"))
};
command.Handler = CommandHandler.Create((bool x, string y) => {  /*  */ })
```

This decouples syntax from method invocation: the method doesn't have to care whether the parameter came from an option or an argument. It does require arguments to be given names, though.

@KathleenDollard @MarkMichaelis Thoughts?

KathleenDollard:
I think this is a great direction (naming arguments and allowing access to them). 

Fro the perspective of the end user typing at the terminal - these are positional, correct?

jonsequitur:
#310 will introduce positional parsing for arguments. 

For this issue, I believe supporting binding arguments by name during invocation will work regardless of the parser rules.

KathleenDollard:
On @tmds original question - the AppModel/JackFruit work currently supports both attributes and a convention of ending in Args. I'd like to chat more about how this could work best after we get the reflection binder finished.

jonsequitur:
Support for this was added by #420.

PR #420, opened 2019-02-25
DragonFruit argument support

jonsequitur:
This addresses #297.

### DIFF
--- a/src/System.CommandLine.DragonFruit/CommandLine.cs
+++ b/src/System.CommandLine.DragonFruit/CommandLine.cs
@@ -96,6 +96,13 @@ namespace System.CommandLine.DragonFruit
             object target = null) =>
             command.ConfigureFromMethod(method, () => target);
 
+        private static readonly string[] _argumentParameterNames =
+        {
+            "arguments",
+            "argument",
+            "args"
+        };
+
         public static void ConfigureFromMethod(
             this Command command,
             MethodInfo method,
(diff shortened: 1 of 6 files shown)


src/System.CommandLine.DragonFruit/CommandLine.cs
```diff
@@ -116,6 +116,18 @@ public static class CommandLine
                 command.AddOption(option);
             }
 
+            if (method.GetParameters()
```
Keboo: This ends up with the parameter being included as both an argument and an option. Is that what was intended? I think at a minimum we should have another test around this behavior (either including these parameters as options or not).

src/System.CommandLine.DragonFruit/CommandLine.cs
```diff
@@ -116,6 +116,18 @@ public static class CommandLine
...
+                      .FirstOrDefault(p => p.Name == "arguments" || 
+                                           p.Name == "args" ||
+                                           p.Name == "argument") is ParameterInfo argsParam)
```
Keboo: Can we change the method signature to take in an optional collection of names to match on for arguments, and fall back to using this set if that collection is not specified. That would also give (albeit non-obvious) method for disabling/configuring this behavior.

src/System.CommandLine.DragonFruit/CommandLine.cs
jonsequitur: Ah, good point. I'll fix that.

src/System.CommandLine.DragonFruit/CommandLine.cs
jonsequitur: This functionality is only available in DragonFruit which is so far not configurable, so I'd prefer to wait on making this more flexible until we decide whether that's something to pursue.

PR #429, opened 2019-03-02
Parameterize suggest directive

### LINKED ISSUE #376: correctly handle completions in the middle of a command line
When the cursor position is in the middle of the input rather than at the end, suggestions are often incorrect. 

While the `position` argument is present in the `dotnet-suggest` CLI and in the `ISuggestionSource` interface, it is not always passed through or acted upon correctly.

jonsequitur:
In order to fix #376, the shim shell scripts will need to pass along the unsplit command line to `dotnet-suggest` in order to preserve whitespace, and also pass along the cursor position.


src/System.CommandLine.Suggest.Tests/SuggestionDispatcherTests.cs
```diff
@@ -22,15 +21,54 @@ private static RegistrationPair CurrentExeRegistrationPair()
...
+
+            
+        }
+
+        [Fact]
+        public async Task InvokeAsync_executes_suggestion_command_for_executable()
+        {
+            string receivedTargetExeArgs = null;
+
+            var args = PrepareArgs($@"get -p 12 -e ""{CurrentExeFullPath()}"" -- {_currentExeName} add");
```
Keboo: Use deconstruction here? Saves on the weird lookings `args.args` below.
```suggestion
            var (args, _) = PrepareArgs($@"get -p 12 -e ""{CurrentExeFullPath()}"" -- {_currentExeName} add");
```

src/System.CommandLine/Parser.cs
```diff
@@ -250,6 +247,15 @@ internal IReadOnlyCollection<string> NormalizeRootCommand(IReadOnlyCollection<st
             }
 
             return args;
+
+            bool FirstArgMatchesExeName() =>
+                firstArg != null &&
+                (
+                    firstArg.Equals(commandName, StringComparison.OrdinalIgnoreCase) ||
+                    firstArg.Equals($"{commandName}.exe", StringComparison.OrdinalIgnoreCase)
+                    ||
```
Keboo: Minor thing here. Mixing white space for or operators. Consider always putting them on the end of the line or putting them on their own lines.

src/System.CommandLine/RootCommand.cs
```diff
@@ -33,9 +33,14 @@ public override string Name
             set => throw new NotSupportedException("The root command's name cannot be changed.");
         }
 
+        private static readonly Lazy<string> executablePath = 
+            new Lazy<string>(() => Assembly.GetEntryAssembly().Location);
+  
         private static readonly Lazy<string> executableName =
-            new Lazy<string>(() => Path.GetFileNameWithoutExtension(Assembly.GetEntryAssembly().Location));
+            new Lazy<string>(() => Path.GetFileNameWithoutExtension(executablePath.Value));
```
Keboo: So what is the point of these `Lazy<string>` here? Because of the static readonly properties below, these get evaluated during the static constructor, effectively mitigating the delayed evaluation.

jonsequitur: Good catch. I've changed them to be late-evaluated properties.

PR #441, opened 2019-03-09
remove dotnet-suggest version suffix; allow disabling Arcade SDK for local builds

jonsequitur:


src/System.CommandLine.Rendering.Tests/System.CommandLine.Rendering.Tests.csproj
```diff
@@ -22,4 +22,12 @@
     <ProjectReference Include="..\System.CommandLine\System.CommandLine.csproj" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(DisableArcade)' == '1'" >
```
wli3: nit: usually in msbuild people test '$(DisableArcade)' != 'true'

wli3: and everything else is false

PR #442, opened 2019-03-09
Response file and tokenization refactor

jonsequitur:
Rather than reuse `ParseError` to track tokenization errors, this introduces a `TokenizationError` class.

I've also added support for referencing one response file from within another.

### DIFF
--- a/src/System.CommandLine/StringExtensions.cs
+++ b/src/System.CommandLine/StringExtensions.cs
@@ -51,7 +51,7 @@ namespace System.CommandLine
             CommandLineConfiguration configuration)
         {
             var tokenList = new List<Token>();
-            var errorList = new List<ParseError>();
+            var errorList = new List<TokenizeError>();
 
             ISymbol currentSymbol = null;
             var foundEndOfArguments = false;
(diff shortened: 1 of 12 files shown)


src/System.CommandLine/StringExtensions.cs
```diff
@@ -192,10 +167,59 @@ internal static string RemovePrefix(this string option)
...
+                    var message = configuration.ValidationMessages
+                                               .ErrorReadingResponseFile(filePath, e);
+
+                    errorList.Add(
+                        new TokenizeError(message));
+                }
+            }
+        }
+
+        private static bool IsResponseFileReference(this string arg)
```
Keboo: I like this but it then means that the code using it has to understand the check and properly substring the `arg`. Perhaps change it up to something like
```suggestion
        private static string GetResponseFileReference(this string arg)
```
and simply return null if the arg is not a response file?

src/System.CommandLine/TokenizeError.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace System.CommandLine
+{
+    internal class TokenizeError
+    {
+        public TokenizeError(string message)
```
Keboo: Null check here?

src/System.CommandLine/TokenizeResult.cs
```diff
@@ -5,17 +5,18 @@
...
 namespace System.CommandLine
 {
-    internal class LexResult
+    internal class TokenizeResult
     {
-        public LexResult(
+        public TokenizeResult(
             IReadOnlyCollection<Token> tokens,
-            IReadOnlyCollection<ParseError> errors)
+            IReadOnlyCollection<TokenizeError> errors)
```
Keboo: Null checks here?

PR #456, opened 2019-03-13
Use Build ID as patch version for dotnet suggest

wli3:
So dotnet suggest version will be something like
1.2.176 . build ID is from Azure Devops


wli3:
@jonsequitur now it should be good

PR #459, opened 2019-03-14
update PackageLicenseUrl to PackageLicenseExpression

jonsequitur:


wli3:
we should also update arcade, arcade tries to add the license url

PR #464, opened 2019-03-17
perf tuning: reduce use of dynamic in binding

jonsequitur:


src/System.CommandLine/SuccessfulArgumentParseResult{T}.cs
```diff
@@ -1,15 +0,0 @@
-﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
-// Licensed under the MIT license. See LICENSE file in the project root for full license information.
-
-namespace System.CommandLine
-{
-    public class SuccessfulArgumentResult<T> : SuccessfulArgumentResult
```
wli3: It shows up in profiler? Passing an object is uglier than <T>

src/System.CommandLine/ArgumentConverter.cs
```diff
@@ -108,16 +99,16 @@ public static ArgumentResult ParseMany<T>(IReadOnlyCollection<string> arguments)
 
             if (successfulParseResults.Length == arguments.Count)
             {
-                dynamic list = Activator.CreateInstance(typeof(List<>).MakeGenericType(itemType));
```
wli3: remove dynamic is already good tho

src/System.CommandLine/SuccessfulArgumentParseResult{T}.cs
jonsequitur: One of the usages of `dynamic` was to allow calling the `Value` property when type `T` is not known at compile time. At this stage we're accepting that boxing will happen during binding, and there aren't solid use cases for API users to deal with `ArgumentResult` directly, so the generic type isn't very useful.

Issue #467, opened 2019-03-20
DragonFruit EntryPointDiscoverer Shoud not only look for `Program` type


evilz:
`Main` method is not forced to be in `Program` in dotnet.
Maybe the where should be removed ?  

https://github.com/dotnet/command-line-api/blob/master/src/System.CommandLine.DragonFruit/EntryPointDiscoverer.cs#L19-L20

jonsequitur:
Fixed by #480

Issue #478, opened 2019-03-27
CommandLine.ExecuteAssemblyAsync reference broken in generated Main()


clmcgrath:
After installing dragonfruit on a brand new project and adding a few parameters 
the AutoGeneratedProgram.g.cs breaks with a missing reference error  at CommandLine.ExecuteAssemblyAsync()
Currently targeting netcoreapp2.2

jonsequitur:
Thanks. Could you provide more detailed repro steps and a code sample?

clmcgrath:
i created i brand new project and it works now 
but in my template project it is breaking 
i will post a public repo for you

clmcgrath:
https://github.com/DigitalParadox/DigitalParadoxCLI.Template/tree/develop

clmcgrath:
It seems to be a conflict with commandlineparser after removing that package it seems to go away , and after reinstalling it it comes back

Keboo:
@clmcgrath good catch on the namespace collision. I have submitted a PR to fix it. This fix wont completely fix your small sample project though. When using DragonFruit it expects that you will write a strongly typed Main method rather than a simple one with string arguments (`static async Task Main(string[] args)`). I suspect this is just an artifact of the sample project and not what you are actually doing in your actual CLI.

clmcgrath:
correct ,  ya I just installed it again to regenerate the problem 
my actual use case has strongly typed args

PR #480, opened 2019-03-27
DragonFruit EntryPointDiscoverer Shoud not only look for `Program` type #467

Kosat:
- Removes code that assumed that entry point (Main method) always resided in Program class. See changes in `EntryPointDiscoverer.cs` as proposed by @evilz in the topic comment to the issue.
- Adds support for `<StartupObject>` project property See: [Common MSBuild project properties](https://docs.microsoft.com/en-us/visualstudio/msbuild/common-msbuild-project-properties?view=vs-2017) 
  Although, user won't be able to use the *(Project Properties)*->*Application*->*Startup object:* drop-down in Visual Studio UI, he/she will be able to add this property to the project file or to the `Directory.Build.props` file manually. I believe it would be nice for `command-line-api` to respect this setting.
 ``` XML
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
     ...
    <TargetFramework>netcoreapp2.1</TargetFramework>
    <StartupObject>RenderingPlayground.Program</StartupObject>
  </PropertyGroup>
```

### CODE REVIEW

File: samples/RenderingPlayground/RenderingPlayground.csproj [hunk 1]
```diff
@@ -1,10 +1,11 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>netcoreapp2.1</TargetFramework>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <LangVersion>latest</LangVersion>
+    <!--<StartupObject>RenderingPlayground.Program</StartupObject>-->
```
Keboo: I like this approach but I would propose having a common name between MSBuild and the C#. Perhaps `EntryPointFullTypeName`?

src/System.CommandLine.DragonFruit/CommandLine.cs
```diff
@@ -21,19 +21,22 @@ public static class CommandLine
         /// </summary>
         /// <param name="entryAssembly">The entry assembly</param>
         /// <param name="args">The string arguments.</param>
+        /// <param name="entryPoint">Explicitly defined entry point</param>
         /// <returns>The exit code.</returns>
         public static async Task<int> ExecuteAssemblyAsync(
             Assembly entryAssembly,
-            string[] args)
+            string[] args,
+            string entryPoint)
```
Keboo: I would propose changing this parameter name to match the name of the MSBuild property.

src/System.CommandLine.DragonFruit/EntryPointDiscoverer.cs
```diff
@@ -11,46 +11,67 @@ namespace System.CommandLine.DragonFruit
...
             var candidates = new List<MethodInfo>();
-            foreach (var type in assembly
-                                 .DefinedTypes
-                                 .Where(t =>
-                                            string.Equals("Program", t.Name, StringComparison.OrdinalIgnoreCase))
-                                 .Where(t => t.GetCustomAttribute<CompilerGeneratedAttribute>() == null))
+
+            if (!string.IsNullOrWhiteSpace(entryPoint))
+            {
+                var typeInfo = assembly.GetType(entryPoint, false, false).GetTypeInfo();
```
Keboo:
```suggestion
                var typeInfo = assembly.GetType(entryPoint, false, false)?.GetTypeInfo();
```

src/System.CommandLine.DragonFruit/targets/System.CommandLine.DragonFruit.targets
```diff
@@ -24,25 +24,31 @@
...
+        using System.CommandLine.DragonFruit;
+        using System.Runtime.CompilerServices;
+        using System.Threading.Tasks;
+
+        [CompilerGenerated]
+        internal class AutoGeneratedProgram
+        {
+          public static async Task<int> Main(string[] args)
+          {
+            return await CommandLine.ExecuteAssemblyAsync(
```
Keboo: This should be changed to fully qualify the method invocation. See #478 
```suggestion
            return await System.CommandLine.DragonFruit.CommandLine.ExecuteAssemblyAsync(
```

src/System.CommandLine.DragonFruit/targets/System.CommandLine.DragonFruit.targets
```diff
@@ -24,25 +24,31 @@
...
-  }
-}
-        ]]>
-      </_GeneratedProgramFileContent>
+      <PropertyGroup>
+         <_GeneratedProgramFileContent >
+
+<![CDATA[
+        // <auto-generated>This file was created automatically</auto-generated>
+        using System.CommandLine.DragonFruit;
```
Keboo: This can be removed. See #478 
```suggestion
```

samples/RenderingPlayground/RenderingPlayground.csproj
Kosat: Kindly, do you mean - " ... propose __NOT_ having a common name" ?
I was also concerned that it may cause problems to `AutoGeneratedProgram.Main()` bootstrap entry point but after testing `<StartupObject>` for a while I observed no any issues with it. So, I think we should give it a chance and if something goes wrong change it to `EntryPointFullTypeName`.  What do you think?

Keboo: I meant have this name match the parameter name being passed into `ExecuteAssemblyAsync`. The casing should obviously be different between them, but I believe the names should match. 
If someone were to enter a type name that did not exist, they would get the error `$"Could not find entry point class '{entryPoint}'."`. This exception message makes sense in the context of the `ExecuteAssemblyAsync` method, but does not indicate that the issue is the `StartupObject` inside of the project file. If these names matched it would be more obvious where the incorrect value were specified.

Issue #469, opened 2019-03-21
add benchmarks


jonsequitur:
A small benchmark suite would be helpful so that we can measure work on performance improvements.

Kosat:
Hello @jonsequitur ,

If I understand correctly, you would like to add a few __BenchmarkDotNet__ benchmarks into a new project , let it be `src/System.CommandLine.Benchmarks`. So that contributors would locally check their changes for performance degradation as explained [here](https://github.com/dotnet/performance/blob/master/docs/benchmarking-workflow-corefx.md#preventing-regressions).

Kindly, can I take this task?

jonsequitur:
That approach sounds good. We'd absolutely appreciate the help.

Thanks @Kosat!

billwert:
hey @jonsequitur - I think we should host these in the dotnet/performance repo as we do the [ml.net](https://github.com/dotnet/performance/tree/master/src/benchmarks/real-world/Microsoft.ML.Benchmarks). What do you think? That way we can run them often for you and you'll have trends (currently internal only, we're working on a reporting story for external.)

billwert:
hi @Kosat! @jonsequitur and I chatteda about this yesterday. For now we think it is best to put the benchmarks in a place in this repo to enable highly iteratitve workflows for contributors here. Once we've reached some point of stability we'll look at putting this in the perf repo and lab to guard against regression.

If you're not familiar with it you might be able to use the [nuget support built into BDN](https://benchmarkdotnet.org/articles/samples/IntroNuGet.html). Wanted to call it out if you hadn't seen it.

PR #486, opened 2019-04-06
add benchmarks #469

Kosat:
I've added the new project `System.CommandLine.Benchmarks` containing BenchmarkDotNet benchmarks.
I've also had to make some small changes to `CommandLine` class to be able to explicitly specify the path to xml doc file in some benchmarks.

Also, I'd like to note that from my observations BenchmarkDotNet benchmarks tend to be very sensitive to antivirus or any other processes (e.g. browser) interfering with the benchmark process. And this is not an issue with my performance tests only; I had tried running some `dotnet/performance` benchmarks locally and observed the same issues with them. Maybe it is just a problem with my machine, but it looks like that the idea of running benchmarks before and after changes and comparing them with `ResultsComparer` with a small threshold (e.g. 2-3%) on a dev machine is prone to producing a lot of false positives.

So, I believe that the most value from these tests will only come after integrating them with CI/CD and monitoring them with BenchView or else, like it is done in `dotnet/performance`.

FYI @billwert


jonsequitur:
@Thanks, @Kosat! This looks great.

Agreed that the benchmarks will be more accurate in a controlled environment which is why we plan to include these eventually in `dotnet/performance`. For now though it's easier to iterate on them here, make sure we have the right set of benchmarks in place to see long-term trends, and then we can move them over. 

@billwert Any comments on this benchmark suite?

src/System.CommandLine.Benchmarks/CommandLine/Perf_Parser_CustomScenarios.cs
```diff
@@ -0,0 +1,37 @@
...
+                                                   Arity = ArgumentArity.ExactlyOne
+                                               }));
+            rootCommand.AddCommand(nestedCommand);
+
+            _testParser = new Parser(rootCommand);
+            _testSymbolsAsString = "root_command nested_command -opt1 321";
+        }
+
+        [Benchmark]
+        public ParseResult Scenario1_Parse() => _testParser.Parse(_testSymbolsAsString);
```
billwert: Can this have a better name? Perf_Parser_CustomScenarios.OneOptWithNestedCommand or something. In the performance reporting system we use the method names as the display names in reporting. Assuming these tests wind up there some day it'd be nice if they didn't change names.

src/System.CommandLine.Benchmarks/CommandLine/Perf_Parser_Directives.cs
```diff
@@ -0,0 +1,47 @@
...
+            "[directive1] [directive2] -opt",
+            "[directive1:1] [directive2:2] -opt",
+            "[directive1] [directive2] [directive2] -opt",
+            "[directive1:1] [directive2:2] [directive2:3] -opt"
+        )]
+        public string TestCmdArgs;
+
+        [Benchmark]
+        public IDirectiveCollection ParseDirectives()
+            => _testParser.Parse(TestCmdArgs).Directives;
```
billwert: It looks like [`Directives`](https://github.com/dotnet/command-line-api/blob/master/src/System.CommandLine/ParseResult.cs#L51) is just a getter. Is this test really about calling `Parse()` with this set of options? If so I would elide fetching the property as it's confusing.

src/System.CommandLine.Benchmarks/CommandLine/Perf_Parser_Directives.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        [Benchmark]
+        public string ParseDirectives_Diagram()
+            => _testParser.Parse(TestCmdArgs).Diagram();
```
billwert: Do you know what the time split between `Parse()` and `Diagram()` is? A quick glance at the code doesn't suggest that `Diagram()` will be cheaper than `Parse()` - I ask because when you mix things together like this you can run the risk of losing regressions in the smaller part beacuse the overall negative contribution isn't big enough. I took a crack at a sample for how this might work [here](https://gist.github.com/billwert/cef0e7eb0156d9db92ca1bf83c6c028b). Sadly `ToString` on a `ParseResult` just calls `Diagram`, so perhaps you can make a small wrapper class which holds the `ParseResult` and has a `ToString` suitable for benchmark naming on it?

src/System.CommandLine.Benchmarks/CommandLine/Perf_Parser_TypoCorrection.cs
```diff
@@ -0,0 +1,57 @@
...
+        [Arguments("--0134679", "3d")]
+        [Arguments("--013467", "4d")]
+        [Arguments("--1023456789", "1t")]
+        [Arguments("--1023546789", "2t")]
+        [Arguments("--1023546798", "3t")]
+        [Arguments("--1032546798", "4t")]
+        public async Task TypoCorrection(string input, string _)
+        {
+            var result = _testParser.Parse(input);
+            await _testParser.InvokeAsync(result, _nullConsole);
```
billwert: same question as before - do you think invoking the parser to find typos is relatively cheap to `Parse`?

src/System.CommandLine.Benchmarks/DragonFruit/Perf_CommandLine_EntryPoint.cs
```diff
@@ -0,0 +1,165 @@
...
+
+namespace System.CommandLine.Benchmarks.DragonFruit
+{
+    /// <summary>
+    /// This is an end-to-end benchmark that measures the performance of
+    /// <see cref="System.CommandLine.DragonFruit.CommandLine.ExecuteAssemblyAsync"/>
+    /// including the execution of Main method per se.
+    /// </summary>
+    [BenchmarkCategory(Categories.DragonFruit)]
+    [InvocationCount(3000)]
```
billwert: why do we need to specify this here? how is the harness doing a poor job picking an iteration count?

src/System.CommandLine.Benchmarks/DragonFruit/Perf_CommandLine_Help.cs
```diff
@@ -0,0 +1,57 @@
...
+using BenchmarkDotNet.Attributes;
+
+namespace System.CommandLine.Benchmarks.DragonFruit
+{
+    /// <summary>
+    /// This is an end-to-end benchmark that measures the performance of --help option.
+    /// </summary>
+    [BenchmarkCategory(Categories.DragonFruit)]
+    [InvocationCount(3000)]
+    public class Perf_CommandLine_Help
```
billwert: how does this differ from passing `--help` in `Perf_CommandLine_Help.cs`? Can we remove this or remove the call there?

src/System.CommandLine.Benchmarks/DragonFruit/Perf_CommandLine_EntryPoint.cs
billwert: oh, I see your BDN bug now.

billwert:
Hello! I left several comments on the diff. Looking forward to the conversation.

Regarding noise in benchmarks caused by environmental things, you're right. Our guidance is always to turn off as many things as you can when running tests. (I add the directory where I'm working to AV exclusion paths, for example.) The dedicated lab is the best place to do things, but with a bit of care you can get "good enough" results locally to give you confidence. I'd recommend writing a readme next to these tests to help contributors in this repo avoid the most common pitfalls. We don't have great guidance for this in the perf repo. We include this phrase: "Please try to avoid running any resource-heavy processes that could spoil the benchmark results while running the benchmarks." which hopefully gets people thinking the right way.

src/System.CommandLine.Benchmarks/CommandLine/Perf_Parser_CustomScenarios.cs
Kosat: Ok. Renamed it to `Perf_Parser_CustomScenarios.OneOptWithNestedCommand_Parse`.

src/System.CommandLine.Benchmarks/CommandLine/Perf_Parser_Directives.cs
Kosat: Fixed. Parsing is now being done separately at setup stage.
src/System.CommandLine.Benchmarks/CommandLine/Perf_Parser_ParseResult.cs

src/System.CommandLine.Benchmarks/CommandLine/Perf_Parser_Directives.cs
Kosat: Fixed. Parsing is now being done separately at setup stage.
src/System.CommandLine.Benchmarks/CommandLine/Perf_Parser_ParseResult.cs

Introduced `BdnParam<T>` wrapper class to substitute `T.ToString()` with a custom comment.
src/System.CommandLine.Benchmarks/Helpers/BdnParam.cs

src/System.CommandLine.Benchmarks/CommandLine/Perf_Parser_TypoCorrection.cs
Kosat: Also fixed it.
Now it looks like 
```C#
        [Benchmark]
        [ArgumentsSource(nameof(GenerateTestParseResults))]
        public async Task TypoCorrection(BdnParam<ParseResult> parseResult)
            => await _testParser.InvokeAsync(parseResult.Value, _nullConsole);
```

src/System.CommandLine.Benchmarks/DragonFruit/Perf_CommandLine_EntryPoint.cs
Kosat: 1. *why do we need to specify this here? how is the harness doing a poor job picking an iteration count?*

Yes, the automatically selected values for iteration count were too small. BND was persistently giving me `MinIterationTime` warnings which are a serious problem, so I had to manually increase the invocations count to the value to the point when I stopped getting those warnings. I hope this value will also be enough for other machines where people will be running those tests.
```
// * Warnings *
MinIterationTime
  Perf_CommandLine_EntryPoint.'ExecuteAssemblyAsync entry point search.': IterationTime=250.0000 ms, MaxIterationCount=20, MinIterationCount=15, WarmupCount=1 -> The minimum observed iteration time is 922.8077 us which is very small. It's recommended to increase it.
```

2. *oh, I see your BDN bug now.*
No. It's a separate issue.

src/System.CommandLine.Benchmarks/DragonFruit/Perf_CommandLine_Help.cs
Kosat: Removed duplicate `--help` test from `Perf_CommandLine_EntryPoint.cs`.

src/System.CommandLine.Benchmarks/CommandLine/Perf_Parser_ParseResult.cs
```diff
@@ -0,0 +1,53 @@
...
+                "[directive1:1] [directive2:2] -opt",
+                "[directive1] [directive2] [directive2] -opt",
+                "[directive1:1] [directive2:2] [directive2:3] -opt",
+            }
+            .Select(cmd => new BdnParam<ParseResult>(_testParser.Parse(cmd), cmd));
+
+        [Benchmark]
+        [ArgumentsSource(nameof(GenerateTestParseResults))]
+        public IDirectiveCollection ParseResult_Directives(BdnParam<ParseResult> parseResult)
+            => parseResult.Value.Directives;
```
billwert: Sorry, I think I was unclear about the guidance around this and `Diagram` below. As far as I can tell, the [code](https://github.com/dotnet/command-line-api/blob/c3e92bc33b5971d0070627b07e977210c07e597c/src/System.CommandLine/Parser.cs#L70-L82) that's interesting when you use `UseParseDirective` actually happens at the invocation of `Parse`. `Directives` just returns a pre-calculated collection on `ParseResult`, so testing this property getter is uninteresting. If testing calculation and parsing of directives is interesting, the previous test was great for that (removing `.Directives`.)

src/System.CommandLine.Benchmarks/CommandLine/Perf_Parser_ParseResult.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        [Benchmark]
+        [ArgumentsSource(nameof(GenerateTestParseResults))]
+        public string ParseResult_Diagram(BdnParam<ParseResult> parseResult)
+            => parseResult.Value.Diagram();
```
billwert: This is different from the previous `Directives` case in that `Diagram` is an extension method which does real work, so testing it in isolation from the cost of `Parse` is great. This is exactly what I was hoping to see.

src/System.CommandLine.Benchmarks/DragonFruit/Perf_CommandLine_EntryPoint.cs
billwert: understood.

src/System.CommandLine.Benchmarks/CommandLine/Perf_Parser_ParseResult.cs
Kosat: Sorry, I've totally overlooked it.  `=> parseResult.Value.Directives;` is indeed a nonsense. I returned `_testParser.Parse(input).Directives;` back.
Fixed it again.

Kosat:
> Great improvements! I clarified one thing around the cases where directives are used. Can you take a look? Once that's fixed I think this looks good.

Thank you!
I'm planning to somewhat change the entry point tests (Perf_CommandLine_EntryPoint.cs) in the next smaller PR. It will require me touching the library code some more and, as this PR is already quite big, I'd be better do it in a separate PR.

billwert:
LGTM.

Issue #497, opened 2019-04-23
 Operation is not supported on this platform


colombod:
System.CommandLine.ConsoleExtensions.ResetTerminalForegroundColor will throw System.PlatformNotSupportedException when used in platforms like mono wasm that do not support that api.

The get accessor for property  System.Console.IsOutputRedirected is the source of such exception.

This is happening when running on wasm and is related to [Console](https://github.com/mono/mono/blob/660ca3bc7094f8418590ba08d1c77c2e6472d50b/mcs/class/corlib/System/Console.cs#L905) implementation

PR #498, opened 2019-04-24
intercept and track the platform not supported exceptions

colombod:
trying to address #497 
The issue so far manifests only on mono when compiled with following symbols

-  MONOTOUCH
- MONODROID 
- XAMMAC 
- WASM 
[see](https://github.com/mono/mono/blob/af6d863e5248a9e67292f94daf6ff0e0cbb38454/mcs/class/corlib/System/Environment.cs#L960) for more details

### DIFF
--- a/src/System.CommandLine/ConsoleExtensions.cs
+++ b/src/System.CommandLine/ConsoleExtensions.cs
@@ -14,13 +14,12 @@ namespace System.CommandLine
                 ((dynamic)console).ForegroundColor = ConsoleColor.Red;
             }
 
-            var supported = Platform.SupportsOperation(() => Console.IsOutputRedirected,
-                    out var isOutputRedirected);
-            if (!supported)
+
+            if (!Platform.IsWasm && !Console.IsOutputRedirected)
             {
                 Console.ForegroundColor = ConsoleColor.Red;
             }
-            else if (!isOutputRedirected)
+            else if (Platform.IsWasm)
             {
                 Console.ForegroundColor = ConsoleColor.Red;
             }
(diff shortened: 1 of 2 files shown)


src/System.CommandLine/ConsoleExtensions.cs
```diff
@@ -13,7 +13,13 @@ internal static void SetTerminalForegroundRed(this IConsole console)
             {
                 ((dynamic)console).ForegroundColor = ConsoleColor.Red;
             }
-            else if (!Console.IsOutputRedirected)
+
+
+            if (!Platform.IsWasm && !Console.IsOutputRedirected)
```
loic-sharma: Nit: two empty lines above this one :)

src/System.CommandLine/Platform.cs
```diff
@@ -0,0 +1,28 @@
+﻿namespace System.CommandLine
+{
+    internal static class Platform
+    {
+        private static bool? _isWasm;
+        public static bool IsWasm
```
jonsequitur: Maybe call this something like `IsConsoleRedirectionSupported` or something? This isn't definitive proof that we're running under WASM, although I'm not aware of other cases where this would happen.

colombod: kind of agree but note that the not supported operation ahs nothing to do with redirection per se, that seems to be support, is to call IsOutputRedirected, or IsErrorRedirected or IsInputRedirected that is not supported.

amis92:
@colombod @jonsequitur I think during the rename from `IsWasm` to `IsConsoleRedirectionCheckSupported` the actual *meaning* of the value was inversed, but the code using it wasn't, which results in the current code being very much contradicting.

* `IsWasm = true` when the accessor `IsOutputRedirected` threw
* `IsConsoleRedirectionCheckSupported = true` is now when the exception is thrown, as well.

The code as it is works as expected, but any other person using this `Platform` API that doesn't check the actual implementation will use it exactly in the opposite way.

I'm not sure I should raise it as a separate issue, so I'm commenting here.

jonsequitur:
Nice catch! Please open a separate issue for this.

PR #507, opened 2019-05-04
XML comments for args param show in help

### LINKED ISSUE #485: DragonFruit XML comment documentation does not show up for arguments
When a DragonFruit `Main` method contains a parameter that will be interpreted by convention as an argument (e.g. `args`, `argument`, `arguments`), documentation for it is not shown.

Example:

```cs
    class Program
    {
        /// <summary>
        /// My example app
        /// </summary>
        /// <param name="args">The arguments</param>
        static void Main(int intOption = 42, string[] args = null)
        {
        }
    }
```

The above code will show the arguments help section as:

```console
Arguments:
  <args>
```

It should show this:

```console
Arguments:
  <args>     The arguments
```

jeredm:
When a XML parameter is set for "args" it will now show up in the help. Fixes #485


jonsequitur:
Thanks, @jeredm!

PR #510, opened 2019-05-05
Command.Argument -> Command.Arguments 

jonsequitur:
This is in preparation for enabling #310.

### DIFF
--- a/src/System.CommandLine.Tests/ParserTests.MultipleArguments.cs
+++ b/src/System.CommandLine.Tests/ParserTests.MultipleArguments.cs
@@ -1,7 +1,10 @@
 ﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Collections.Generic;
+using FluentAssertions;
 using Xunit;
+using Xunit.Abstractions;
 
 namespace System.CommandLine.Tests
 {
@@ -9,17 +12,66 @@ namespace System.CommandLine.Tests
     {
         public class MultipleArguments
         {
-            [Fact(Skip = "#310")]
-            public void Mulitple_positional_arguments_can_differ_by_arity()
+            private readonly ITestOutputHelper _output;
+
+            public MultipleArguments(ITestOutputHelper output)
             {
-                // FIX: (Mulitple_positional_arguments_can_differ_by_arity) 
+                _output = output;
+            }
+
+            [Fact(Skip = "wip")]
+            public void Multiple_arguments_can_differ_by_arity()
+            {
                 var command = new Command("the-command")
                 {
-                    // new Argument<string>()
+                    new Argument<string>
+                    {
+                        Arity = new ArgumentArity(3, 3),
+                        Name = "several"
+                    },
+                    new Argument<string>
+                    {
+                        Arity = ArgumentArity.ZeroOrMore,
+                        Name = "one"
+                    }
                 };
-                
-                // TODO-JOSEQU (Mulitple_positional_arguments_can_differ_by_arity) write test
-                Assert.True(false, "Test Mulitple_positional_arguments_can_differ_by_arity is not written yet.");
+
+                var result = command.Parse("1 2 3 4");
+
+                _output.WriteLine(result.ToString());
+
+                var several = result.CommandResult
+                                    .GetArgumentValueOrDefault<IEnumerable<string>>("several");
+
+                var one = result.CommandResult
+                                .GetArgumentValueOrDefault<string>("one");
+
+                several.Should()
+                       .BeEquivalentTo("1", "2", "3");
+                one.Should()
+                   .Be("4");
+            }
+
+            [Fact(Skip = "wip")]
+            public void Multiple_arguments_can_differ_by_type()
+            {
+                var command = new Command("the-command")
+                {
+                    new Argument<string>
+                    {
+                        Name = "several"
+                    },
+                    new Argument<int>
+                    {
+                        Name = "one"
+                    }
+                };
+
+                var result = command.Parse("1 2");
+
+                _output.WriteLine(result.ToString());
+
+                Assert.True(false, "Test Multiple_positional_arguments_can_differ_by_type is not written yet.");
             }
         }
     }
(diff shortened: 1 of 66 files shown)


src/System.CommandLine/ArgumentResult.cs
```diff
@@ -5,14 +5,19 @@ namespace System.CommandLine
 {
     public abstract class ArgumentResult
     {
-        private protected ArgumentResult()
+        protected internal ArgumentResult(IArgument argument)
```
Keboo: Given that this is an abstract class, can't this simply be `protected`

src/System.CommandLine/ArgumentResult.cs
```diff
@@ -5,14 +5,19 @@ namespace System.CommandLine
...
         {
+            Argument = argument;
         }
 
-        public static FailedArgumentResult Failure(string error) => new FailedArgumentResult(error);
+        public IArgument Argument { get; }
 
-        public static SuccessfulArgumentResult Success(object value) => new SuccessfulArgumentResult(value);
+        public string ErrorMessage { get; set; }
```
Keboo: It seems un-usual to have an ErrorMessage on the base class. Can this be moved to the `FailedArgumentResult` class?

src/System.CommandLine/Option.cs
```diff
@@ -13,14 +15,43 @@ public class Option : Symbol, IOption
...
+
+                AddArgumentInner(value);
+            }
+        }
+
+        IArgument IOption.Argument => Argument;
+
+        string IValueDescriptor.ValueName => Name;
+
+        Type IValueDescriptor.Type => _arguments.Single().ArgumentType;
```
Keboo: Should this be `=> Argument.ArgumentType;`?
The same for `HasDefaultValue` and `GetDefaultValue()`. 

Since the argument is optional when instantiating the class, I suspect we should not assume the `_arguments` collection has any items.

src/System.CommandLine/SymbolResult.cs
```diff
@@ -137,22 +131,188 @@ internal SymbolResult TryTakeArgument(Token token)
             }
         }
 
-        public ArgumentResult ArgumentResult
+        private readonly HashSet<IArgument> _argumentsUsingDefaultValue = new HashSet<IArgument>();
```
Keboo: This should probably move up with the other readonly fields.

src/System.CommandLine/SymbolResult.cs
```diff
@@ -137,22 +131,188 @@ internal SymbolResult TryTakeArgument(Token token)
...
+                return null;
+            }
+
+            ParseError CustomError()
+            {
+                foreach (var symbolValidator in argument.SymbolValidators)
+                {
+                    var errorMessage = symbolValidator(symbolResult);
+
+                    if (!String.IsNullOrWhiteSpace(errorMessage))
```
Keboo:
```suggestion
                    if (!string.IsNullOrWhiteSpace(errorMessage))
```

src/System.CommandLine/Token.cs
```diff
@@ -22,13 +22,7 @@ public Token(string value, TokenType type)
...
             Type == token.Type;
 
-        public override int GetHashCode()
-        {
-            var hashCode = 1574892647;
-            hashCode = hashCode * -1521134295 + EqualityComparer<string>.Default.GetHashCode(Value);
-            hashCode = hashCode * -1521134295 + Type.GetHashCode();
-            return hashCode;
-        }
+        public override int GetHashCode() => (Value, Type).GetHashCode();
```
Keboo: 👍

src/System.CommandLine/ConvertArgument.cs
```diff
@@ -3,5 +3,7 @@
 
 namespace System.CommandLine
 {
-    public delegate ArgumentResult ConvertArgument(SymbolResult symbolResult);
+    public delegate bool TryConvertArgument(SymbolResult symbolResult, out object value);
+
+    public delegate bool TryConvertArgument<T>(SymbolResult symbolResult, out T value);
```
Keboo: IMO I liked the delegate signature better when it was expected to return a result object, and that object handled the success/failure rather than going with the TryParse pattern

src/System.CommandLine.Tests/ParserTests.MultipleArguments.cs
```diff
@@ -0,0 +1,78 @@
...
+using System.Collections.Generic;
+using FluentAssertions;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace System.CommandLine.Tests
+{
+    public partial class ParserTests
+    {
+        public class MultipleArguments
```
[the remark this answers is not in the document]
jonsequitur: They're not ready. There's another PR in the works that will enable this but some of the core parser pieces need significant refactoring first.

src/System.CommandLine/ArgumentArity.cs
```diff
@@ -64,9 +74,9 @@ public ArgumentArity(int minimumNumberOfArguments, int maximumNumberOfArguments)
 
         public static IArgumentArity ExactlyOne => new ArgumentArity(1, 1);
 
-        public static IArgumentArity ZeroOrMore => new ArgumentArity(0, int.MaxValue);
+        public static IArgumentArity ZeroOrMore => new ArgumentArity(0, byte.MaxValue);
```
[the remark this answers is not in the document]
jonsequitur: Overflow issues with `int.MaxValue`. We could also use some arbitrary value in between, if that seems better.

src/System.CommandLine/ArgumentResult.cs
jonsequitur: Changing back to `private protected` for the moment. Ultimately I think this class needs to be rewritten so that it inherits `SymbolResult`.

src/System.CommandLine/ArgumentResult.cs
jonsequitur: It's pretty weird. The whole class has to be replaced. I can make this internal in the meantime.

src/System.CommandLine/ConvertArgument.cs
jonsequitur: The reason I made this change was that it felt a bit more discoverable and arguably more idiomatic for C#. This also helps to untangle a chicken/egg problem between parsing and conversion/validation. That's very much a work in progress though.

PR #512, opened 2019-05-07
add SymbolResult.Tokens, make SymbolResult.Arguments obsolete

jonsequitur:


src/System.CommandLine/CommandResult.cs
```diff
@@ -8,7 +8,10 @@ namespace System.CommandLine
 {
     public class CommandResult : SymbolResult
     {
-        public CommandResult(ICommand command, CommandResult parent = null) : base(command, command?.Name, parent)
+        public CommandResult(
+            ICommand command, 
+            Token token = null,
+            CommandResult parent = null) : base(command, token ?? command?.DefaultToken(), parent)
```
Keboo: If the passed in `command` is null, it will throw an augment null exception on the token, which is highly misleading. I know it looks funny, but perhaps something like this:
```C#
        public CommandResult(
            ICommand command, 
            Token token = null,
            CommandResult parent = null) : base(command ?? throw new ArgumentNullException(nameof(command)), token ?? command.DefaultToken(), parent)
        {
            Command = command;
        }
```

src/System.CommandLine/OptionResult.cs
```diff
@@ -7,9 +7,9 @@ public class OptionResult : SymbolResult
     {
         public OptionResult(
             IOption option,
-            string token = null,
+            Token token = null,
             CommandResult parent = null) :
-            base(option, token ?? option?.Token(), parent)
+            base(option, token ?? option?.DefaultToken(), parent)
```
Keboo: Same issue as the `CommandResult` above.

PR #513, opened 2019-05-08
fix parsing bug when minimum arity is greater than 1

jonsequitur:


src/System.CommandLine/Token.cs
```diff
@@ -15,6 +17,29 @@ internal Token(string value, TokenType type)
...
+            Value == token.Value &&
+            Type == token.Type;
+
+        public override int GetHashCode()
+        {
+            var hashCode = 1574892647;
+            hashCode = hashCode * -1521134295 + EqualityComparer<string>.Default.GetHashCode(Value);
+            hashCode = hashCode * -1521134295 + Type.GetHashCode();
+            return hashCode;
+        }
```
Keboo: This hashing function will not produce very randomly distributed hashes for `Tokens` with the same `Value`.
A few options:
```cs
public static int GetHashCode()
{
    unchecked
    {
        return ((int) Type * 397) ^ (Value != null ? Value.GetHashCode() : 0);
    }
}
```

Or leverage the [RyuJIT optimizations](https://source.dot.net/#System.Private.CoreLib/shared/System/Numerics/Hashing/HashHelpers.cs,12) that were done in `ValueTuple` 
```cs
public static int GetHashCode()
{
  return (Type, Value).GetHashCode();
}
```

jonsequitur: Fixing this in my other PR.

PR #515, opened 2019-05-10
fix #514

### LINKED ISSUE #514: Arity ZeroOrOne when argument not supplied can cause bound constructor to be called with null
When an arity of 0 is allowed for an argument that will be bound to a single-string constructor call (e.g. a path string passed to `FileInfo`) then if the argument is not supplied on the command line, null is passed and the constructor may throw (e.g. `ArgumentNullException`).

jonsequitur:


src/System.CommandLine/ArgumentConverter.cs
```diff
@@ -193,17 +204,27 @@ public static bool CanBeBoundFromScalarValue(this Type type)
...
-                .Where(c =>
-                {
-                    var parameters = c.GetParameters();
-                    return c.IsPublic &&
-                           parameters.Length == 1 &&
-                           parameters[0].ParameterType == parameterType;
-                })
-                .SingleOrDefault();
+            Type parameterType,
+            out (ConstructorInfo ctor, ParameterDescriptor parameterDescriptor) info)
```
Keboo: This would probably be better as two separate out parameters.

PR #525, opened 2019-05-15
Add alias attribute to XML doc comment params for DragonFruit

markrendle:
There is currently no way (at least that I can see) to add aliases when using the DragonFruit API.

This change lets you add an `alias` attribute to the `param` tags in your XML doc comments to set an alias for the parameter. It seemed like a non-invasive way of introducing the functionality, as opposed to, say, creating an `AliasAttribute` for parameters.

Example usage:
```csharp
    class Program
    {
        /// <summary>
        /// DragonFruit simple example program
        /// </summary>
        /// <param name="verbose" alias="v">Show verbose output</param>
        /// <param name="flavor" alias="f">Which flavor to use</param>
        /// <param name="count" alias="c">How many smoothies?</param>
        static int Main(
            bool verbose,
            string flavor = "chocolate",
            int count = 1)
        {
            if (verbose)
            {
                Console.WriteLine("Running in verbose mode");
            }
            Console.WriteLine($"Creating {count} banana {(count == 1 ? "smoothie" : "smoothies")} with {flavor}");
            return 0;
        }
    }
```

Help output:

```
$ dotnet run -- --help
DragonFruit:
  DragonFruit simple example program

Usage:
  DragonFruit [options]

Options:
  -v, --verbose            Show verbose output
  -f, --flavor <flavor>    Which flavor to use
  -c, --count <count>      How many smoothies?
  --version                Display version information
```

Oh, and I updated the SDK version in global.json because `2.2.204` just came out and I've uninstalled `2.2.203`.

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "tools": {
-    "dotnet": "2.2.203",
+    "dotnet": "2.2.204",
     "vs": {
       "version": "16.0"
     }
--- a/samples/DragonFruit/Program.cs
+++ b/samples/DragonFruit/Program.cs
@@ -10,9 +10,9 @@ namespace DragonFruit
         /// <summary>
         /// DragonFruit simple example program
         /// </summary>
-        /// <param name="verbose">Show verbose output</param>
-        /// <param name="flavor">Which flavor to use</param>
-        /// <param name="count">How many smoothies?</param>
+        /// <param name="verbose" alias="v">Show verbose output</param>
+        /// <param name="flavor" alias="f">Which flavor to use</param>
+        /// <param name="count" alias="c">How many smoothies?</param>
         static int Main(
             bool verbose,
             string flavor = "chocolate",
--- a/src/System.CommandLine.DragonFruit.Tests/CommandLineTests.cs
+++ b/src/System.CommandLine.DragonFruit.Tests/CommandLineTests.cs
@@ -50,7 +50,7 @@ namespace System.CommandLine.DragonFruit.Tests
                   .Contain("<args>    These are arguments")
                   .And.Contain("Arguments:");
             stdOut.Should()
-                  .Contain("--name <name>    Specifies the name option")
+                  .Contain("-n, --name <name>    Specifies the name option")
                   .And.Contain("Options:");
             stdOut.Should()
                   .Contain("Help for the test program");
--- a/src/System.CommandLine.DragonFruit.Tests/TestProgram.cs
+++ b/src/System.CommandLine.DragonFruit.Tests/TestProgram.cs
@@ -14,7 +14,7 @@ namespace System.CommandLine.DragonFruit.Tests
         /// <summary>
         /// Help for the test program
         /// </summary>
-        /// <param name="name">Specifies the name option</param>
+        /// <param name="name" alias="n">Specifies the name option</param>
         /// <param name="console"></param>
         /// <param name="args">These are arguments</param>
         public void TestMain(string name, IConsole console, string[] args = null)
--- a/src/System.CommandLine.DragonFruit.Tests/XmlDocReaderTests.cs
+++ b/src/System.CommandLine.DragonFruit.Tests/XmlDocReaderTests.cs
@@ -1,4 +1,4 @@
-// Copyright (c) .NET Foundation and contributors. All rights reserved.
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System.IO;
@@ -30,7 +30,7 @@ namespace System.CommandLine.DragonFruit.Tests
             <summary>
             Hello
             </summary>
-            <param name=""verbose"">Show verbose output</param>
+            <param name=""verbose"" alias=""v"">Show verbose output</param>
             <param name=""flavor"">Which flavor to use</param>
             <param name=""count"">How many smoothies?</param>
         </member>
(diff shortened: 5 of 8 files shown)


global.json
```diff
@@ -1,6 +1,6 @@
 {
   "tools": {
-    "dotnet": "2.2.203",
+    "dotnet": "2.2.204",
```
Keboo: Is this change intentional?

jonsequitur:
@KathleenDollard and I previously declined a PR by @wojtpl2 that was uncannily similar: https://github.com/dotnet/command-line-api/pull/345.

We clearly missed something.

We've had ongoing discussions around app models and where the line should be between the core API and the app models. The parser, syntax types, and invocation pipeline model the main concerns in this problem space so they go in the core. App models are for the (arguably) more opinionated parts that in many frameworks will influence the structure of your app, so we want them to be opt-in, and make the core API support creating them. DragonFruit is our first example.

An odd thing about DragonFruit is that it doesn't add new concepts. I've had the experience many times of people not even noticing it's there. They see the `Main` method but don't notice that the arguments aren't `string[]`. You don't even need a `System.CommandLine` `using` directive.

Now, I don't think this is necessarily valuable and clearly it's limiting. The question from the beginning has been how to add capabilities, and most of the thinking has gone towards the C# part of the API rather than the XML part. So far the XML part supports help but not the definition of your app's interface. I had personally leaned away from adding to the XML part because to me it felt less intuitive and less discoverable.

Then @markrendle and @wojtpl2 went and implemented the exact same feature. 😳

So I'll ask this: If this is a direction that makes sense, what is the most important missing capability after aliases, and where should _that_ one go?

global.json
markrendle: It brings it up to the current SDK version. I always install them as they are released and uninstall the previous ones, so I don't have 2.2.203 installed any more.

markrendle:
> The question from the beginning has been how to add capabilities, and most of the thinking has gone towards the C# part of the API rather than the XML part. So far the XML part supports help but not the definition of your app's interface. I had personally leaned away from adding to the XML part because to me it felt less intuitive and less discoverable.
> 
> Then @markrendle and @wojtpl2 went and implemented the exact same feature. 😳
> 
> So I'll ask this: If this is a direction that makes sense, what is the most important missing capability after aliases, and where should _that_ one go?

@jonsequitur 

So in this instance, my first instinct was to go with attributes on the parameters, but there are two issues with that:

1. As you say, it means bringing in a `using System.CommandLine.DragonFruit;` directive where one is not currently needed.
2. It makes the `Main` declaration line *really* ugly:

```csharp
public static void Main([Alias("s")] string source, [Alias("o")] string output, [Alias("p")] string packageName)
{
    // Automatically convert a .NET 4.x WCF service to a .NET Core 3.0 gRPC project
    // (Seriously, that's what my app does)
}
```

However, I do agree that it's strange having something in the doc comments actually changing functionality, and I'm not precious about this solution. I'm happy to rewind and open a new PR with the attributes if that's the preferred approach. I just want aliases in DragonFruit.

markrendle:
I guess another option might be like this:

```csharp
[Alias(nameof(source), "s")]
[Alias(nameof(output), "o")]
[Alias(nameof(packageName), "p")]
public static void Main(string source, string output, string packageName)
{
    // Automatically convert a .NET 4.x WCF service to a .NET Core 3.0 gRPC project
    // (Seriously, that's what my app does)
}
```

Slightly less ugly.

global.json
KathleenDollard: Set global.json to the lowest that will not break a contributor, since setting this higher will force a contributor to upgrade their SDK. 

There are few cases where requiring a specific patch is needed. If the one specified is not found, any higher patch (last 2 digits) will be used. 

@jonsequitur are we using anything specific to 2.2? To 2.2.200? 

2.2 is particularly concerning since it goes out of support sooner and would force an upgrade to 3.0.  2.1.300 may be a better choice for System.CommandLine to correspond to the lowest needed for global tools.

KathleenDollard:
Dragonfruit is an app model (not core, although we can discuss that with the BCL folks, I will be surprised if they'd take it) and magic. I don't mind it being more magic by having the XML changes. 

However, I think there are plenty of places where we will need an attribute for app models, and I think we should start collecting them. I don't know if they will make it into core, but they should at least be in a library easily accessed by all app model authors. We don't want to wake up one day with the same attribute in two app models one with a British and one with an American spelling ;-)

So, I actually think both could/maybe should be supported.

jonsequitur:
Another approach that's been discussed is to support a `Configure` method, which would let you have lower-level access to the parser configuration before the DragonFruit entrypoint is invoked.

Air code:

```cs
public static void Main(string source, string output, string packageName)
{
}

public static void Configure(CommandLineConfiguration configuration)
{
    configuration.RootCommand.Options["source"].AddAlias("s");
    // ...
}
```

(This doesn't currently work because the `CommandLineConfiguration` object is effectively immutable but we could change that if it seems useful.)

john-h-k:
Something to allow aliases would be very useful. Is there a way yet or is this the only open PR about it?

jonsequitur:
We're exploring ways to expand the DragonFruit model. We want to do it in a way that provides access to all of the functionality of `System.CommandLine` including things like custom validation and subcommands, which is why we're hesitant to put responsibilities on the XML part of the API.

I'm going to close this PR but we'd welcome issues or suggestions to help make DragonFruit more capable. https://github.com/dotnet/command-line-api/issues?q=is%3Aopen+is%3Aissue+label%3ADragonFruit

Issue #440, opened 2019-03-07
Integration with GenericHost


haefele:
Are there any plans on how to integrate this with the GenericHost from Microsoft.Extensions.Hosting?

I mean all the functionality the GenericHost provides like the hosting environment, logging, DI, configuration - would still be useful when creating a command-line API.

I saw that you can add your own registrations to the BindingContext.ServiceProvider - but I'm not quite sure if that's the right way to integrate the GenericHost with this library.

jonsequitur:
It's been brought up a couple of times but not looked at closely. The service registration in `BindingContext` is still a work in progress. How might you imagine the integration working?

loic-sharma:
@natemcmaster's fork of CommandLineUtils has integration with generic hosts: https://natemcmaster.github.io/CommandLineUtils/docs/advanced/generic-host.html

The highlights are:

1. There's an extension method on [`IHostBuilder`](https://github.com/natemcmaster/CommandLineUtils/blob/master/src/Hosting.CommandLine/HostBuilderExtensions.cs#L32) to run the command line app
2. Dependency injection works with the command handlers

couven92:
I concluded in #533 that adding the generic can neatly be done as an Invocation Middleware. This means that command-handlers can accept parameters of type `IHost` and continue from there. E.g. a handler can get the Host DI-container from the `IHost.Services` property. This becomes especially nice when using the `Host.CreateDefaultBuilder()` factory method coming in `Microsoft.Extensions.Hosting` version 3.0 (currently in preview).

Actually there are some things to be said for separating the command-line parser services from the host. Take a look at #533 where I have done that, but provided the ability to bridge that separation if necessary.

The approach is similar to how the generic host separates host configuration and app configuration. I.e. the generic host actually first builds a Configuration for the host, and then injects that into the application DI-container. I create the Host and start it, and then add it to the invocation binding context.

If you do the Generic Host as a command-line invocation middleware, you still get the benefits of being able to do `--help` short-circuiting different behaviours depending on commands, etc, and it does so without having to jump through too many hoops.

alexdresko:
@couven92 Now you just have to get your branch to pass the CI build. :)

haefele:
@couven92 That is basically what I ended up doing as well - just not as well organized. Good job! :smile:

couven92:
[...]

PR #533, opened 2019-05-31
Add CommandLine API extensions for Microsoft.Extensions.Hosting

couven92:
Adds a new project `System.CommandLine.Hosting` project that provides extension methods for `CommandLineBuilder` to use a CommandLine InvocationPipeline with the .NET Generic Host API provided by the `Microsoft.Extensions.Hosting` package.

This implements a CommandLine middleware that configures and builds a new `IHost`. The built `IHost` instance is then added to the `BindingContext` service provider and the host is started. When the `InvocationPipeline` completes, the host is stopped and disposed.

By default the Hosting middleware will simply create a new instance of the `HostBuilder` class and then run a user-supplied host-configuration delegate. Additionally, there is an overload that allows the caller to control how the `IHostBuilder` is created. The new `HostedConsoleApp` sample shows how the `Host.CreateDefaultBuilder()` (currently in the version 3.0 preview of `Microsoft.Extensions.Hosting`) can be used instead of the plain `HostBuilder` class.

Since the created host instance is added to the invocation pipeline binding context, any command handler can accept an `IHost` instance and resolve hosted services as shown in the `HostedConsoleApp` sample.

These API extensions are supplied in a separate project, because it supplies opt-in convenience methods. Pulling in all dependencies from `Microsoft.Extensions.Hosting` would be overkill for simple small console applications. The project could instead depend on `Microsoft.Extensions.Hosting.Abstractions` which has much fewer dependencies, but doing so would **require** callers to specify a `IHostBuidler` factory delegate, as `HostBuilder` lives in the other package.


src/System.CommandLine.Hosting/HostingExtensions.cs
```diff
@@ -0,0 +1,48 @@
...
+namespace System.CommandLine.Hosting
+{
+    public static class HostingExtensions
+    {
+        public static CommandLineBuilder UseHost(this CommandLineBuilder builder,
+            Func<string[], IHostBuilder> hostBuilderFactory,
+            Action<IHostBuilder> configureHost = null) =>
+            builder.UseMiddleware(async (invocation, next) =>
+            {
+                var argsRemaining = invocation.ParseResult.UnmatchedTokens.ToArray();
```
jonsequitur: >argsRemaining [](start = 20, length = 13)

What's an example of what these args would contain?

If these are coming from the command line input and filtered through the parser configuration, I wonder if this might be unpredictable.

jonsequitur:
Thanks, @couven92. 

Can you add some tests to illustrate and verify the usage?

jonsequitur:
#440

src/System.CommandLine.Hosting/HostingExtensions.cs
couven92: As shown in the added test case shown below unmatched tokens can be passed on to the HostConfiguration. If you include the `Microsoft.Extensions.Configuration.CommandLine` package.

https://github.com/dotnet/command-line-api/blob/9f1fa174542e5c6296bb7570f42cd1fb12df4a98/src/System.CommandLine.Hosting.Tests/HostingTests.cs#L91-L127

For example, you can thus pass `--environment Development` to set the Hosting Environment to `Development`. ASP.NET uses this to configure Kestrel from command-line arguments. However, the `AddCommandLine()` extension is really simple and only supports Key-Value-Pairs, so `System.CommandLine` is MUCH nicer to actual command-line argument parsing :)

The default on the `Command` constructor is `treatUnmatchedTokensAsError: true`, so normally the `UnmatchedTokens` property in `ParseResult` will be empty causing the host configuration to remain unaffected.

It is also worth pointing out that the normal `HostBuilder` constructor does not accept command-line arguments. That was only added in the 3.0 preview with the `Host.CreateDefaultBuilder` method. If have added test-cases showing how the unmatched command-line arguments can be passed into the Host configuration.

couven92:
@jonsequitur I have added a few illustrating test cases. :)

couven92:
Hmm... @jonsequitur I have no idea why the OSX CI build fails... It does not look like it's my fault though, can you help?

jonsequitur:
Thanks. I'll take a look.

src/System.CommandLine.Hosting/HostingExtensions.cs
jonsequitur: There are a couple of other approaches we might want to explore here. Requiring someone to relax what is usually an error check seems restrictive.

One option is to use unparsed tokens: https://github.com/dotnet/command-line-api/blob/4c7695193bb79d649cd3711bb00639a0c2c5b458/src/System.CommandLine.Tests/ParserTests.cs#L160 Similar to the `UnmatchedTokens` though this does potentially conflict with someone's preferences for their app's command line interface.

An approach that wouldn't constrain the other parts of someone's interface would be to use directives: https://github.com/dotnet/command-line-api/wiki/Syntax-Concepts-and-Parser#directives. The wiki doesn't cover the fact that directives are parameterizable. In the following example, "config" is the directive name. Everything after the colon is the directive value. Multiple directives having the same name will aggregate the values into a collection.

```console
> myapp [config:key=value]           --int-option 123
         ^----^ ^-------^
```

couven92: @jonsequitur Oh! I totally agree! I will change it it to use `UnparsedTokens` instead.

Using the a `config` directive sounds like a great idea. You could already do it yourself right now, by getting the `InvocationContext` from the `HostBuilderContext`. But making that first-class would be so much nicer and not break anyone.

I'll add that in ASAP :)

jonsequitur:
@couven92 The macOS build issue is fixed. You'll want to rebase onto the latest master.

couven92:
@jonsequitur As you suggested, I have added an `AddCommandLineDirectives` extension method for `IConfigurationBuilder` which takes a `ParseResult` and a `name`. The method will add all key-value-pairs (separated by `=`) to the configuration builder.

The `UseHost` extension method calls that method with `config` as the directive name, thus flowing all config-directives to the host configuration.

samples/HostedConsoleApp/Program.cs
```diff
@@ -0,0 +1,35 @@
...
+
+namespace HostedConsoleApp
+{
+    public static class Program
+    {
+        public static Task Main(string[] args) => new CommandLineBuilder(
+            new RootCommand
+            {
+                Handler = CommandHandler.Create<IHost, string>(Run),
+                TreatUnmatchedTokensAsErrors = false
```
jonsequitur: We shouldn't need this anymore, correct?

couven92: Remoed

src/System.CommandLine.Hosting/HostingExtensions.cs
```diff
@@ -0,0 +1,55 @@
...
+                });
+                hostBuilder.ConfigureServices(services =>
+                {
+                    services.AddSingleton(invocation);
+                    services.AddSingleton(invocation.BindingContext);
+                    services.AddSingleton(invocation.Console);
+                    services.AddTransient(_ => invocation.InvocationResult);
+                    services.AddTransient(_ => invocation.ParseResult);
+                });
+                hostBuilder.UseConsoleLifetime();
```
tmds: @couven92 @jonsequitur it would be nice if we could use the `CancellationToken` from https://github.com/dotnet/command-line-api/blob/a58aa11718049db7a9e49a65cd52db4bfb42e38f/src/System.CommandLine/Invocation/InvocationExtensions.cs#L19 here.

couven92: @tmds you can do that already. A handler taking both `IHost` and `CancellcationToken` as arugments would still gte the cancellation token from the `BindingContext` as you usually would already.

During the `AddMiddleware` method used in `UseHost` there is unfortunately no way to resolve services added to the `BindingContext`, since I cannot access its `ServiceProvider` property from within the `Hosting` project. And even we could, what would you have me do with the cancellation token? Use in the calls to `host.StartAsync()` and `host.StopAsync`?

I don't think it would be a great idea to add the cancellation token to the host service provider using `ConfigureServices`. If that was your intention, I'd much rather add the `CancellationTokenSource` obtained from the private field `_cts` in `InvocationContext`, but that might not be ideal either.

couven92: Discussion for this moved to issue https://github.com/dotnet/command-line-api/issues/549.

PR #535, opened 2019-06-01
initial Try .NET tutorial work, handler binding tutorial

jonsequitur:


README.md
```diff
@@ -18,6 +18,30 @@ System.CommandLine.Rendering       | [![Nuget](https://img.shields.io/nuget/v/Sy
 
 Daily builds are available if you add this feed to your nuget.config: https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json.
 
+## Interactive tutorials
+
+If you'd like to try out `System.CommandLine` using an interactive tutorial of its features and API, you can, using Try .NET.
+
+![binding](https://user-images.githubusercontent.com/547415/58752436-905aa880-8463-11e9-9ab7-c2a8136b0a93.gif)
+
+To use the tutorial, first clone this repo:
```
KathleenDollard:
```suggestion
To use the tutorial, first clone the System.CommandLine repo:
```

this was probably OK, but thought specific good

README.md
```diff
@@ -18,6 +18,30 @@ System.CommandLine.Rendering       | [![Nuget](https://img.shields.io/nuget/v/Sy
...
+Next, install the `dotnet try` global tool:
+
+```console
+> dotnet tool install -g dotnet-try
+```
+
+Finally, launch the `dotnet try` pointing to the tutorial directory:
+
+```console
+> dotnet try <PATH_TO_REPO>/samples/tutorial
```
KathleenDollard:
```suggestion
> dotnet try <PATH_TO_COMMAND_LINE_REPO>/samples/tutorial
```

Although not sure that is exactly right

samples/tutorial/Binding.md
```diff
@@ -0,0 +1,25 @@
+# Binding
+
+Parsing command line arguments is a means to an end. You probably don't really want to think about parsing the command line. You just want some arguments passed to a method. 
+
```
KathleenDollard: Do we want to mention type binding here?

samples/tutorial/Binding.md
```diff
@@ -0,0 +1,25 @@
...
+In C#, that method has always looked something like this:
```
KathleenDollard:
```suggestion
In C#, the application startup that contained the arguments has always looked something like this:
```

samples/tutorial/BindingToCommandHandlers.md
```diff
@@ -0,0 +1,62 @@
...
+```cs --source-file ./src/Binding/HandlerBindingSample.cs --project ./src/Binding/Binding.csproj --region FileSystemTypes 
+```
+
+## Anything with a string constructor
+
+But `FileInfo` and `DirectoryInfo` are not special cases. Any type having a constructor that takes a single string parameter can be bound in this way. Go back to the previous example and try using a `Uri` instead.
+
+## More complex types
+
+Binding also supports creating instances of more complex types than we used in the examples above. If you have a large number of options, this can be cleaner than adding more parameters to your handler. `System.CommandLine` has the default convention of binding `Option` arguments to either properties or constructor parameters by name. The name matching uses the same strategies that are used when matching parameters on a handler method.
```
KathleenDollard: First sentence is a bit awkward. Suggest

```suggestion
Binding also supports creating instances of more complex types. If you have a large number of options, this can be cleaner than adding more parameters to your handler. `System.CommandLine` has the default convention of binding `Option` arguments to either properties or constructor parameters by name. The name matching uses the same strategies that are used when matching parameters on a handler method. 
```

samples/tutorial/BindingToCommandHandlers.md
```diff
@@ -0,0 +1,62 @@
...
+## More complex types
+
+Binding also supports creating instances of more complex types than we used in the examples above. If you have a large number of options, this can be cleaner than adding more parameters to your handler. `System.CommandLine` has the default convention of binding `Option` arguments to either properties or constructor parameters by name. The name matching uses the same strategies that are used when matching parameters on a handler method. 
+
+In the next sample, the handler accepts an instance of `ComplexType`. Try removing its setters and uncommenting the constructor. Try adding properties, or changing the types or names of its properties.
+
+```cs --source-file ./src/Binding/HandlerBindingSample.cs --project ./src/Binding/Binding.csproj --region ComplexTypes 
+```
+
+## Dependency injection
```
KathleenDollard: Assuming we aren't doing full DI, consider softening this title - or give an example of using the programmer's own type

samples/tutorial/src/Binding/HandlerBindingSample.cs
```diff
@@ -0,0 +1,187 @@
...
+        {
+            #region MultipleArgs
+
+            var command = new RootCommand
+            {
+                new Option("--a-string") { Argument = new Argument<string>() },
+                new Option("--an-int") { Argument = new Argument<int>() }
+            };
+
+            command.Handler = CommandHandler.Create(
```
KathleenDollard: I wonder if this would be more clear with a method, even a local method, than the lambda

samples/tutorial/src/Binding/GetValueSample.cs
```diff
@@ -0,0 +1,24 @@
...
+        {
+            #region GetValueFromOptionArgument
+
+            var option = new Option("--an-int")
+            {
+                Argument = new Argument<int>()
+            };
+            var parseResult = option.Parse("--an-int 123");
+            var value = parseResult.FindResultFor(option).GetValueOrDefault();
+            value.PrintToConsole();
```
KathleenDollard: You don't consistently use PrintToConsole, so it feels a bit distracting. 

```
Console.WriteLine(FormatOutput(value));
```

would leave focus on the Ain work.

samples/tutorial/src/Rendering/AnsiSamples.cs
```diff
@@ -0,0 +1,81 @@
+using System;
```
KathleenDollard: Do we want to check these in prior to the tutorial?

samples/tutorial/BindingToCommandHandlers.md
jonsequitur: By "full DI" do you mean "from a container"? We can fully support passing someone's custom types today (which may or may not come from a DI container, we're not really opinionated about it), but it probably deserves a separate tutorial because configuring this has little to do with binding.

samples/tutorial/src/Binding/HandlerBindingSample.cs
jonsequitur: I was opting for an older construct on the assumption more people would be familiar with it.

samples/tutorial/BindingToCommandHandlers.md
KathleenDollard: Maybe skip in this tutorial. I think there are two topics: the automatically available types you can have just by asking, and the ability to add your own type.

jonsequitur: Agreed. Mainly I wanted to point out that the capability exists in this context.

Related: https://github.com/dotnet/command-line-api/pull/533

samples/tutorial/BindingToCommandHandlers.md
```diff
@@ -0,0 +1,62 @@
...
+Binding also supports creating instances of more complex types. If you have a large number of options, this can be cleaner than adding more parameters to your handler. `System.CommandLine` has the default convention of binding `Option` arguments to either properties or constructor parameters by name. The name matching uses the same strategies that are used when matching parameters on a handler method. 
+
+In the next sample, the handler accepts an instance of `ComplexType`. Try removing its setters and uncommenting the constructor. Try adding properties, or changing the types or names of its properties.
+
+```cs --source-file ./src/Binding/HandlerBindingSample.cs --project ./src/Binding/Binding.csproj --region ComplexTypes 
+```
+
+## Dependency injection
+
+Not everything you might want passed to your handler will necessarily come from parsed command line input. There are a number of types provided by `System.CommandLine` that you can bind to. The following example demonstratres injection of `ParseResult` and `IConsole`. Other types can be passed this way as well.
```
Keboo: Something that seems to be lacking from our documentation is a list of built-in binding parameters.

samples/tutorial/BindingToCommandHandlers.md
Keboo: I agree with softening the title here. Perhaps something more like "built-in parameter types"?

samples/tutorial/BindingToCommandHandlers.md
jonsequitur: Yeah, this definitely deserves more detailed documentation.

samples/tutorial/BindingToCommandHandlers.md
jonsequitur: "System.CommandLine types" ?

Keboo: Yea I like that.

samples/tutorial/src/Binding/HandlerBindingSample.cs
BillWagner: FWIW, we tend to prefer using the newer features where applicable. It encourages adoption for readers to see them.

When I read these samples, the lambda "feels" natural to me. In production code, would you use a lambda or a local (even local static) function? I don't know if seeing the lambda is because it looks familiar, or if it's "right".

jonsequitur: As to which is right, it depends, of course. :smile: I suppose for a smaller codebase the lambda seems simpler and more cohesive, but for larger codebases the static or local function helps break the code into more understandable pieces. In `dotnet try` our handlers are static methods in separate classes entirely. Among other things, this enables testing them directly.

Issue #458, opened 2019-03-14
Can CommandHandler.Create() go more than 7 parameters?


lorenh:
This works:
subcommand.Handler = CommandHandler.Create((int p1, int p2, int p3, ... int p7) => ...

This doesn't:
subcommand.Handler = CommandHandler.Create((int p1, int p2, int p3, ... int p8) => ...

Is it reasonable to ask to keep going a few more parameters since Func<> and Action<> support more?

Keboo:
My personal opinion is the `CommandHandler.Create` should probably include overloads to match the `Func<>` and `Action<>` arity.

jonsequitur:
Another approach that's worth considering: You can do bind complex objects.

```cs
command.Handler = CommandHandler.Create<MyType>(obj => /*   */ );
```

Constructor injection and property setters are both supported, with the same naming conventions as we use to match parameters.

AndyAyersMS:
Can you say a bit more about this? I have a command with 9 options ... what should I do?

jonsequitur:
Let's assume you have a command line with options `-a` through `-i`. You can create a handler like this:

```cs
command.Handler = CommandHandler.Create<MyType>(myType => /*   */ );
```

where `MyType` looks like either of the following:

```cs
public class MyType
{
    public string A { get; set; }
    public string B { get; set; }
    public string C { get; set; }
    public string D { get; set; }
    public string E { get; set; }
    public string F { get; set; }
    public string G { get; set; }
    public string H { get; set; }
    public string I { get; set; }
}

// or...

public class MyType
{
    public MyType(string a, string b, string c, string d, string e, string f, string g, string h, string i)
    {
        A = a;
        B = b;
        C = c;
        D = d;
        E = e;
        F = f;
        G = g;
        H = h;
        I = i;
    }

    public string A { get; }
    public string B { get; }
    public string C { get; }
    public string D { get; }
    public string E { get; }
    public string F { get; }
    public string G { get; }
    public string H { get; }
    public string I { get; }
}
```

AndyAyersMS:
Thanks, that works nicely.

granadacoder:
> My personal opinion is the `CommandHandler.Create` should probably include overloads to match the `Func<>` and `Action<>` arity.

I agree.  

Aka, 16.

https://docs.microsoft.com/en-us/dotnet/api/system.func-17?view=netcore-3.1

granadacoder:
[...]

PR #536, opened 2019-06-02
CommandHandler.Create() to handle more than 7 parameters.

xavierjohn:
CommandHandler.Create() to handle more than 7 parameters.

Issue #458


jonsequitur:
@SamWhiteNVR The recommended approach is to bind to a complex type rather than add more parameters. You can see an example here: https://github.com/dotnet/command-line-api/issues/458#issuecomment-495789697

granadacoder:
My opinion is that the command line should match the number available for the "regular Func".

Aka, 16.

https://docs.microsoft.com/en-us/dotnet/api/system.func-17?view=netcore-3.1

PR #537, opened 2019-06-03
Core parser refactor

jonsequitur:
The previous implementation of the core parser coupled parsing and syntax concerns with binding concerns. There are a few issues I'm planning to address with this refactor.

* `ArgumentResult` was not a `SymbolResult` while `CommandResult` and `OptionResult` were. `SymbolResult.Children` did not contain the `ArgumentResult` objects. This is confusing.
* `SymbolResult` coupled binding, parsing, and validation in ways that were hard to follow.
* Changing the parsing rules independently of the binding rules was therefore difficult.

I'm separating the parsing functionality with the goal of slimming down the `*Result` types. New parser types are in the `System.CommandLine.Parsing` namespace.


src/System.CommandLine/StringExtensions.cs
```diff
@@ -210,6 +200,74 @@ void ReadResponseFile(string filePath, int i)
...
+            }
+            else
+            {
+                args = new[]
+                       {
+                           commandName
+                       }.Concat(args).ToList();
+            }
+
+            return args.ToList();
```
Keboo: Both branches of the above if statement call `.ToList()`. You can likely remove them in favor of just this one.

src/System.CommandLine/SymbolExtensions.cs
```diff
@@ -15,14 +16,14 @@ internal static class SymbolExtensions
...
-             !string.IsNullOrWhiteSpace(symbol.Description) ||
+            (!String.IsNullOrWhiteSpace(symbol.Name) ||
+             !String.IsNullOrWhiteSpace(symbol.Description) ||
              symbol.Arguments().Any(a => a.ShouldShowHelp()));
 
         internal static bool ShouldShowHelp(
             this IArgument argument) =>
             argument != null &&
-            !string.IsNullOrWhiteSpace(argument.Name) &&
+            !String.IsNullOrWhiteSpace(argument.Name) &&
```
Keboo: Is there a reason for converting away from the keyword?

src/System.CommandLine/SymbolResult.cs
```diff
@@ -25,293 +28,220 @@ public abstract class SymbolResult
...
+                return ArgumentResults.SingleOrDefault() ??
+                       ArgumentResult.None(argument);
+            }
+        }
+
+        internal ArgumentResultSet ArgumentResults
+        {
+            get
+            {
+                // FIX: (ArgumentResults)
```
Keboo: Not sure I understand this comment

src/System.CommandLine/SymbolResult.cs
```diff
@@ -25,293 +28,220 @@ public abstract class SymbolResult
...
+            return _defaultArgumentValues.GetOrAdd(
+                argument,
+                a => a.GetDefaultValue());
         }
 
         internal bool UseDefaultValueFor(IArgument argument)
         {
-            return _argumentsUsingDefaultValue.Contains(argument);
-        }
+            // FIX: (UseDefaultValueFor)
```
Keboo: I am not sure I understand this comment

src/System.CommandLine/SymbolResult.cs
```diff
@@ -25,293 +28,220 @@ public abstract class SymbolResult
...
+                    return ArgumentResult.Success(argument, symbolResult.Arguments);
+            }
 
-                    default:
-                        return ArgumentResult.Success(argument, symbolResult.Arguments);
-                }
+            bool ShouldCheckArity()
+            {
+                return !(symbolResult is OptionResult optionResult) ||
+                       !optionResult.IsImplicit;
```
Keboo:
```suggestion
                return symbolResult is OptionResult optionResult && optionResult.IsImplicit;
```

src/System.CommandLine/CommandResultExtensions.cs
```diff
@@ -42,5 +43,39 @@ public static T GetValueOrDefault<T>(this CommandResult commandResult)
...
+        {
+            var children = commandResult.Children
+                                        .Where(o => valueDescriptor.ValueName.IsMatch(o.Symbol))
+                                        .ToArray();
+
+            SymbolResult symbolResult = null;
+
+            if (children.Length > 1)
+            {
+                throw new ArgumentException($"Ambiguous match while trying to bind parameter {valueDescriptor.ValueName} among: {String.Join(",", children.Select(o => o.Name))}");
```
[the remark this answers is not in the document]
jonsequitur: This will be done at some point but localizing exceptions messages is out of scope for the moment.

src/System.CommandLine/SymbolResult.cs
jonsequitur: This is just a reminder to myself. 

To explain what's in progress here, the `ArgumentResult` versus `ArgumentResult2` issue still has to be sorted out and I'm wondering whether the former needs to remain anywhere in the public API. What I have in mind is something like this:

* `ArgumentResult` and its subclasses get renamed (`ArgumentConversionResult`?) and made internal. Properties that expose it get hidden since the main APIs for retrieving argument results go through `Command` or `Option` anyway (in order to deal gracefully with default or missing values).
* `ArgumentResult2` gets renamed to `ArgumentResult`.

Thoughts?

src/System.CommandLine/SymbolResult.cs
jonsequitur: Removed.

src/System.CommandLine/SymbolResult.cs
Keboo: I like that

src/System.CommandLine/SymbolExtensions.cs
jonsequitur: Must be a format setting. This was unintentional.

PR #542, opened 2019-06-06
Multiple distinct arguments for commands

jonsequitur:
This addresses #310.


src/System.CommandLine/ArgumentConversionResult.cs
```diff
@@ -0,0 +1,23 @@
...
+        private protected ArgumentConversionResult(IArgument argument)
+        {
+            Argument = argument ?? throw new ArgumentNullException(nameof(argument));
+        }
+
+        public IArgument Argument { get; }
+
+        internal string ErrorMessage { get; set; }
+
+        internal static FailedArgumentConversionResult Failure(IArgument argument, string error) => new FailedArgumentConversionResult(argument, error);
```
KathleenDollard: Why two internal and one public?

jonsequitur: Looks like a historical artifact. The whole class is internal now in any case.

tmds:
Awesome! Thanks @jonsequitur !

PR #551, opened 2019-06-11
Issue546 cursor control Show/Hide

xavierjohn:
Create CursorControlSpan to Show and Hide cursor independent of the terminal type.

### DIFF
--- a/src/System.CommandLine.Rendering.Tests/Views/ScreenViewTests.cs
+++ b/src/System.CommandLine.Rendering.Tests/Views/ScreenViewTests.cs
@@ -35,7 +35,7 @@ namespace System.CommandLine.Rendering.Tests.Views
         [Fact]
         public void ScreenView_requires_a_console()
         {
-             Action nullRenderer = () => new ScreenView(_renderer, null);
+            Action nullRenderer = () => new ScreenView(_renderer, null);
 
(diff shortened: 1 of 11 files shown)


src/System.CommandLine.Rendering.Tests/Views/ScreenViewTests.cs
```diff
@@ -47,31 +47,29 @@ public void Render_hides_the_cursor()
 
             screen.Render();
 
-            _terminal.Out
-                     .ToString()
-                     .Should()
-                     .StartWith(Ansi.Cursor.Hide.EscapeSequence);
+            _terminal.Events
+                .Should()
+                .BeEquivalentSequenceTo(new TestTerminal.CursorHidden());
```
jonsequitur: Nice change.

src/System.CommandLine.Rendering/CursorControlSpan.cs
```diff
@@ -0,0 +1,32 @@
...
+        {
+            Name = name;
+            Row = row;
+            Column = column;
+        }
+        public override int ContentLength => 0;
+
+        public string Name { get; }
+        public int Row { get; }
+        public int Column { get; }
```
jonsequitur: Are `Row` and `Column` being used? They don't look applicable  to all of the `CursorControlSpan` variants.

Issue #549, opened 2019-06-10
Integrate CancellationToken from InvocationContext into UseHost method


couven92:
With the newly merged `System.CommandLine.Hosting` extensions, @tmds raised the question whether the `UseHost` method should use/pass on a `CancellationToken` from the `InvocationContext`. To better separate the [discussion](https://github.com/dotnet/command-line-api/pull/533#discussion_r291921200) started in a closed [Pull Request (#533)](https://github.com/dotnet/command-line-api/pull/533), I decided to copy the thread and continue the discussion here.

## Background

In cases where you add hosted services in the generic host (e.g. by calling `AddHostedService()`) these services might take a while to start or to shutdown (depending on the nature of the service). For that reason, `IHost` supports passing a `CancellationToken` to the `StartAsync` and `StopAsync` methods which indicate whether start should be aborted immediately, or if graceful shutdown should be circumvented.

Currently we do not pass any `CancellationToken` to either `StartAsync` nor `StopAsync`, so startup and shutdown is always as graceful as can be, but during that time there is now way for the user to terminate the operation unless they forcefully kill the process (i.e. pressing CTRL+C does not immediately shut down).

## Proposal

I see two possible approaches for this problem. The second approach has two minor variations.

1. We could decide that starting and stopping the host is the responsibility of the user, meaning that they should call `StartAsync` and `StopAsync` in their own user-defined Command handler themselves.
   In this case you'd define your handler using e.g. `CommandHandler.Create<IHost, CancellationToken>(Run)`. The `CancellationToken` would be resolved from the `BindingContext` of the invocation and work together with `CommandLine` middleware like `CancelOnProcessTermination`.
2. If the `UseHost` could access the `ServiceProvider` property on the `BindingContext` class, the middleware could simply resolve the `CancellationToken` from the invocation `BindingContext` in the same way the arguments to a command handler are resolved. However, the `ServiceProvider` property of `BindingContext` is currently internal, making it inaccessible for `System.CommandLine.Hosting`.
   1. We could mark the internals of `System.CommandLine` visible to `System.CommandLine.Hosting` in order to be able to access the internal `ServiceProvider` property of `BindingContext`.
   2. We could make the `ServiceProvider` property as `public` and re-type the property to `IServiceProvider` to prevent people from adding services.

[...]

PR #552, opened 2019-06-11
Integrate CancellationToken from InvocationContext into UseHost method

couven92:
Implementation of variant 2 of the 2<sup>nd</sup> approach described in issue #549.

* Add new instance method `InvokeContext.GetCancellationToken()` which resolves the `CancellationToken` from the service provider of the `BindingContext`.
* Resolve `CancellationToken` in `HostingExtensions.UseHost` to pass it to `host.StartAsync` and `host.StopAsync` to be able to forcefully shut down the application if needed.

### DIFF
--- a/src/System.CommandLine.Hosting/HostingExtensions.cs
+++ b/src/System.CommandLine.Hosting/HostingExtensions.cs
@@ -36,15 +36,17 @@ namespace System.CommandLine.Hosting
                 hostBuilder.UseConsoleLifetime();
                 configureHost?.Invoke(hostBuilder);
 
+                var invokeCancel = invocation.GetCancellationToken();
+
                 using (var host = hostBuilder.Build())
                 {
                     invocation.BindingContext.AddService(typeof(IHost), () => host);
 
-                    await host.StartAsync();
+                    await host.StartAsync(invokeCancel);
 
                     await next(invocation);
 
-                    await host.StopAsync();
+                    await host.StopAsync(invokeCancel);
                 }
             });
 
--- a/src/System.CommandLine/Invocation/InvocationContext.cs
+++ b/src/System.CommandLine/Invocation/InvocationContext.cs
@@ -13,6 +13,14 @@ namespace System.CommandLine.Invocation
 
         public BindingContext BindingContext { get; }
 
+        public CancellationToken GetCancellationToken()
+        {
+            var obj = BindingContext.ServiceProvider.GetService(typeof(CancellationToken));
+            if (obj is CancellationToken ct)
+                return ct;
+            return CancellationToken.None;
+        }
+
         public InvocationContext(
             ParseResult parseResult,
             IConsole console = null)


jonsequitur:
Sorry to arrive late to the discussion. 

I'd prefer to find an approach that keeps the service provider internal. It's deliberately an implementation detail. For example, it's not capable of the kind of dependency composition or lifetime support that people might expect. I think before we take this step, there should be some discussion of whether we want more complex DI support and what that would look like.

In the meantime, I would advocate a more limited approach, perhaps exposing the `CancellationToken` directly.

couven92:
@jonsequitur like https://github.com/dotnet/command-line-api/pull/552/commits/935a402524f2b793a88142fb5aed83966e129997 ? I added `InvokeContext.GetCancellationToken()` instead.

couven92:
and again I have no idea why the Windows Debug build suddenly failed...?

jonsequitur:
Thanks @couven92!

src/System.CommandLine.Hosting/HostingExtensions.cs
```diff
@@ -36,15 +36,17 @@ public static class HostingExtensions
                 hostBuilder.UseConsoleLifetime();
                 configureHost?.Invoke(hostBuilder);
 
+                var invokeCancel = invocation.GetCancellationToken();
```
tmds: What's the difference between calling this, and calling `invocation.AddCancellationHandling()`?

couven92: @tmds AddCancellationHandling() can only be called once. And it's called that one time in the InvokeContext constructor. It is also internal, and thus not callable from other assemblies.

couven92: calling it multiple times will throw an `InvalidOperationException`

tmds: > calling it multiple times will throw an InvalidOperationException

I know. I put the exception there.

The idea is: this is called once when the invocation context becomes cancellable, from then on-wards the user needs to use that CancellationToken.

>  And it's called that one time in the InvokeContext constructor.

It doesn't call `AddCancellationHandling`, but adds a factory method for when someone binds to a `CancellationToken`.
You can call `AddCancellationHandling` in `UseHost`, it shouldn't throw.

couven92: @tmds yeah, but adding Cancellation Handling is not a concern that `UseHost` should deal with. You'd call `CancelOnProcessTermination` to add that capability.

couven92: more importantly, `UseHost` would otherwise exclude usage for `CancelOnProcessTermination` or **ANY** other middlware that wanted to implement their own cancellation handling.

There is simply no need for `UseHost` to that, when it can simply get the CancellationToken (however that is configured) from the `InvocationContext`. `UseHost` wants to use `CancellationToken` it does not need to actually set it up (why would it need to?).

tmds: My understanding is the method added here is the same thing as:
```
public CancellationToken GetCancellationToken()
  => AddCancellationHandling();
```
so I don't understand what you mean by:

> more importantly, UseHost would otherwise exclude usage for CancelOnProcessTermination or ANY other middlware that wanted to implement their own cancellation handling.

> it does not need to actually set it up (why would it need to?).

The pattern is to distinguish between invocations that can be cancelled and those that do not. In the latter case we don't want `CancelOnProcessTermination` to ignore SIGINT/SIGTERM.

couven92: @tmds okay, yes I can see what you mean. However, you could still override the `CancellationToken` in the `BindingContext` service provider by calling `AddService<CancellationToken>()`. Not sure anyone would want to do that, but technically possible.

More importantly: You still cannot simply call `AddCancellationHandling`. The method is internal, an thus not accessible.

`GetCancellationToken` mimics what the BindingContext would actually do when it needed to fill in a `CancellationToken` value, so that is the more sensible way to do it (for a public method). Since `AddCancellationHandling` is internal, it is a mere implementation detail and cannot guarantee a correct value, whereas `BindingContext` actually IS the source of truth here.

tmds: > However, you could still override the CancellationToken in the BindingContext service provider by calling AddService<CancellationToken>(). Not sure anyone would want to do that, but technically possible.

This isn't a feature.

> More importantly: You still cannot simply call AddCancellationHandling. The method is internal, and thus not accessible.

We can make it public.

> GetCancellationToken mimics what the BindingContext would actually do when it needed to fill in a CancellationToken value, so that is the more sensible way to do it (for a public method). Since AddCancellationHandling is internal, it is a mere implementation detail and cannot guarantee a correct value, whereas BindingContext actually IS the source of truth here.

The ServiceProvider is for implementing the typed argument binding. I don't see the value of using it here, and it makes a complex `GetCancellationToken` implementation.

PR #553, opened 2019-06-12
Merge GetCancellationToken and AddCancellationHandling

tmds:
These functions do the same thing.

@couven92 @jonsequitur 

`AddCancellationHandling` is the point where the InvocationContext becomes cancellable, that is: the invocation can be canceled because it is using the cancellation token.

```
        /// <summary>
        /// Indicates the invocation can be cancelled.
        /// </summary>
        /// <returns>Token used by the caller to implement cancellation handling.</returns>
```

### DIFF
--- a/src/System.CommandLine.Hosting/HostingExtensions.cs
+++ b/src/System.CommandLine.Hosting/HostingExtensions.cs
@@ -33,20 +33,20 @@ namespace System.CommandLine.Hosting
                     services.AddTransient(_ => invocation.InvocationResult);
                     services.AddTransient(_ => invocation.ParseResult);
                 });
-                hostBuilder.UseConsoleLifetime();
                 configureHost?.Invoke(hostBuilder);
-
-                var invokeCancel = invocation.GetCancellationToken();
 
                 using (var host = hostBuilder.Build())
                 {
                     invocation.BindingContext.AddService(typeof(IHost), () => host);
 
-                    await host.StartAsync(invokeCancel);
+                    // Stop the host when the invocation gets cancelled.
+                    var cancellationToken = invocation.AddCancellationHandling();
+
+                    await host.StartAsync(cancellationToken);
 
                     await next(invocation);
 
-                    await host.StopAsync(invokeCancel);
+                    await host.StopAsync(cancellationToken);
                 }
             });
 
--- a/src/System.CommandLine.Hosting/HostingExtensions.cs
+++ b/src/System.CommandLine.Hosting/HostingExtensions.cs
@@ -3,6 +3,7 @@ using System.CommandLine.Invocation;
 using System.Linq;
 
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection.Extensions;
 using Microsoft.Extensions.Hosting;
 
 namespace System.CommandLine.Hosting
@@ -33,6 +34,7 @@ namespace System.CommandLine.Hosting
                     services.AddTransient(_ => invocation.InvocationResult);
                     services.AddTransient(_ => invocation.ParseResult);
                 });
+                hostBuilder.UseInvocationLifetime(invocation);
                 configureHost?.Invoke(hostBuilder);
 
                 using (var host = hostBuilder.Build())
@@ -53,5 +55,17 @@ namespace System.CommandLine.Hosting
         public static CommandLineBuilder UseHost(this CommandLineBuilder builder,
             Action<IHostBuilder> configureHost = null
             ) => UseHost(builder, null, configureHost);
+
+        public static IHostBuilder UseInvocationLifetime(this IHostBuilder host,
+            InvocationContext invocation, Action<InvocationLifetimeOptions> configureOptions = null)
+        {
+            return host.ConfigureServices(services =>
+            {
+                services.TryAddSingleton(invocation);
+                services.AddSingleton<IHostLifetime, InvocationLifetime>();
+                if (configureOptions is Action<InvocationLifetimeOptions>)
+                    services.Configure(configureOptions);
+            });
+        }
     }
 }
--- a/src/System.CommandLine.Hosting/HostingExtensions.cs
+++ b/src/System.CommandLine.Hosting/HostingExtensions.cs
@@ -41,14 +41,11 @@ namespace System.CommandLine.Hosting
                 {
                     invocation.BindingContext.AddService(typeof(IHost), () => host);
 
-                    // Stop the host when the invocation gets cancelled.
-                    var cancellationToken = invocation.GetCancellationToken();
-
-                    await host.StartAsync(cancellationToken);
+                    await host.StartAsync();
 
                     await next(invocation);
 
-                    await host.StopAsync(cancellationToken);
+                    await host.StopAsync();
                 }
             });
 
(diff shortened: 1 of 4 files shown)


couven92:
I still think this is suboptimal.

1. The naming of `AddCancellationHandling` is unfortunate and misleading. `GetCancellationToken` actually conveys what happens.
2. It is unfortunate that `AddCancellationHandling` can only be called once. What we really need for middleware is `GetCancellationToken`. If resolving the cancellation token from the `BindingContext` is problematic (why, reoslving requested values is exactly it purpose is it not?), can we instead modify `GetCancellationToken` to return `_cts.Token` instead? Maybe call `AddCancellationHandling` provisionally if `_cts` is `null`? That way, multiple calls to `GetCancellationToken` would not throw, but still return the correct value.

tmds:
> The naming of AddCancellationHandling is unfortunate and misleading. GetCancellationToken actually conveys what happens.

Naming is hard. GetCancellationToken seems like we are just getting something, while we are enabling cancellation. Suggestions for other names?

> It is unfortunate that AddCancellationHandling can only be called once. What we really need for middleware in cases like these is GetCancellationToken. If resolving the cancellation token from the BindingContext is problematic (why, resolving requested values is exactly it purpose is it not?), can we instead modify GetCancellationToken to return _cts.Token instead? Maybe call AddCancellationHandling provisionally if _cts is null? That way, multiple calls to GetCancellationToken would not throw, but still return the correct value.

The rationale for throwing is this: once the user gets a cancellation token, he must be using it, so there should be some sort of strategy on how cancellation is used throughout the entire invocation. Getting a CancellationToken at several different times seems like there is no such strategy.

If it is really needed, it can be changed.

couven92:
> The rationale for throwing is this: once the user gets a cancellation token, he must be using it

I don't see why anyone would care. Sure, if you call a method that returns a value back to you, you should probably use it. But I fail to understand why we'd force anyone to use it.

I do however question the rationale that a CancellationToken for an invocation pipeline only can be obtained once by a single middleware.

I can easily imagine scenarious where you'd have multiple middlewares in your pipline that all want to have a cancellation token. Currently that is not possible. I think it should. So I'd rather see the changes made to `GetCancellationToken` so that it could be invoked multiple times.

tmds:
Updated to allow multiple retrievals of the token without throwing.

couven92:
I'd also leave the call to host.UseConsoleLifetime. I imagine it would surprise people that there is no lifetime service available in the host when you run it.

tmds:
> I imagine it would surprise people that there is no lifetime service available in the host when you run it.

How would someone 'use' this lifetime?

couven92:
@tmds For example if your host contains a service this:

``` cs
public class Class1
{
    private readonly IApplicationLifetime lifetime;
    public Class1(IApplicationLifetime lifetime)
    {
        this.lifetime = lifetime;
    }

    public Task Foobar()
    {
        return Task.Delay(TimeSpan.MaxValue, lifetime.ApplicationStopping);
    }
}
```

jonsequitur:
If we want to ensure that the same cancellation is returned on repeat calls, maybe just a plain old property, which I think implies that it will be the same instance: `InvocationContext.CancellationToken` on repeated calls, although the behavior change that calling this implies would be less obvious.

A complete and honest name would be something like: `EnableCancellationAndGetCancellationToken`

@tmds Can you remind me, if I use cancellation handling and don't cancel the token explicitly, will the process exit when my handler method exits?

couven92:
@tmds Never mind. I was mistaked. A service relying on `IHostLifetime` would not be able to resolve the service. But `IHostLifetime` does not look like it does anything particularly useful, so that might be okay.

tmds:
> public Class1(IApplicationLifetime lifetime)

`IHostLifetime` and `IApplicationLifetime` are not the same.
`IHostLifetime` is what ConsoleLifetime implements and provides lifetime events for the host, and `IApplicationLifetime` is for the users to consume.

> @tmds Can you remind me, if I use cancellation handling and don't cancel the token explicitly, will the process exit when my handler method exits?

The process termination events cancel the token.
As long as no-one requested the token, they don't do anything and just let the process terminate.
When someone requests the token, the handlers stop the process from terminating, and trust the requester to cause the app to eventually terminate.

couven92:
yes, and funnily enough, `ConsoleLifetime` is added to the host anyways, so there's no need to actually call `UseConsoleLifetime()` unless you want to re-overwrite a previously added `IHostLifetime` service.

jonsequitur:
> When someone requests the token, the handlers stop the process from terminating, and trust the requester to cause the app to eventually terminate.

This part I get. Specifically, what must I do to cause the process to terminate? Is it sufficient to let my handler exit (e.g. I do a bunch of work in a loop, then I'm done, then it exits normally) or is the process going to live until I cancel the token?

tmds:
> what must I do to cause the process to terminate?

The cancellation is triggered externally, by the user pressing Control+C, or someone sending SIGINT/SIGTERM to the process.

couven92:
Hmm... looking at the source code of MS.Ext.Hosting and System.CommandLine the following can happen.

Pressing Ctrl+C or unexpected process termination (through `AppDomain.ProcessExit`) will by default cause the `IHost` to shut down. (Because `ConsoleLifetime` is added by default.) You can prevent that by adding an `IHostLifetime` service that does something else/nothing.

If you add `CancelOnProcessTermination()` to your invocation pipeline, pressing Ctrl+C will also cancel the cancellation token in the InvocationContext. This will subsequently lead to the host being shutdown non-gracefully. Unexpected process termination will wait for the handler to return before the process is terminated.

If you do NOT add  `CancelOnProcessTermination()` to your invocation pipeline the invocation context cancellation token is never cancelled.

So your handler should observe either the cancellation token from the invocation (by taking `CancellationToken` as an argument, or observe the `IHost` shutting down, by resolving the `IApplicationLifetime` service and observe the cancellation token from the `ApplicationStopping` property.

If your handler does not observe any cancellation tokens, Ctrl+C and unexpected termination will effectively be supressed and wait for the handler to return.

tmds:
> If you do NOT add CancelOnProcessTermination() to your invocation pipeline the invocation context cancellation token is never cancelled.

Your process will just die on SIGINT/SIGTERM.

> if your handler does not observe any cancellation tokens, Ctrl+C and unexpected termination will effectively be supressed and wait for the handler to return.

Yes, if you enable cancellation by getting the token, you need to also consume the CancellationToken.

couven92:
Yes, problem is: the lifetime of the host ALSO suppresses Ctrl+C which might be surprising

couven92:
So we should probably add our own lifetime into the host which observes the token from the invocation context

tmds:
> So we should probably add our own lifetime into the host which observes the token from the invocation context

That makes sense to me. I wish the Host wouldn't assume ConsoleLifetime by default which is the real issue.

couven92:
@tmds I have an idea for that... Just give me a moment 🙂

couven92:
@tmds You can cherry-pick https://github.com/couven92/dotnet-command-line-api/commit/ee2f01aa6979f1386e1a4d91c83877f11012324e (or just fast-forward your branch to the `cancel` branch on https://github.com/couven92/dotnet-command-line-api.git, I based it on your branch)

ref.: https://github.com/tmds/command-line-api/compare/host_cancellation...couven92:cancel

couven92:
I basically just implemented `InvocationLifetime` based on the `ConsoleLifetime` so that the behaviour is as similar as possible. (Most people will probably not even notice the difference).

PR #555, opened 2019-06-14
Add support for multiple arguments in DragonFruit

xavierjohn:
Add support for multiple mandatory arguments.
All arguments that ends with "Arguments",  "Argument",  "Args" are required.

Example:
`static void Main(string destinationArgument, string sourceArgument)`


jonsequitur:
The naming convention is an interesting idea. I think the same convention should be usable for options, so that you could do this:

```cs
static void Main(
int somethingCountOption = 5, // configures Option<int> named "--something-count"
FileInfo[] filesArgument = null  // configures Argument<FileInfo[]> named "files" 
)


```

Note I also made "option" and "argument" prefixes, which I think reads a little better.

Thoughts?

This also will need some tests.

src/System.CommandLine.DragonFruit.Tests/ConfigureFromMethodTests.cs
```diff
@@ -221,6 +221,19 @@ public void Options_are_not_built_for_infrastructure_types_exposed_by_method_par
                    .NotContain(o => o.Argument.ArgumentType == type);
         }
 
+        [Fact]
+        public void Parameters_with_Argument_suffix_is_treated_as_required_arguments()
```
jonsequitur: Whether or not the argument is required depends on its arity. But this is also orthogonal to the naming convention, so I'd rename the test to remove "required".

src/System.CommandLine.DragonFruit.Tests/ConfigureFromMethodTests.cs
```diff
@@ -221,6 +221,19 @@ public void Options_are_not_built_for_infrastructure_types_exposed_by_method_par
...
+        {
+            var parser = new CommandLineBuilder()
+             .ConfigureRootCommandFromMethod(GetMethodInfo(nameof(Method_having_two_arguments)))
+             .Build();
+
+            var rootCommandArgument = parser.Configuration.RootCommand.Arguments.ToList();
+            rootCommandArgument.Should().HaveCount(2);
+            rootCommandArgument[0].Type.Should().Be(typeof(string));
+            rootCommandArgument[1].Type.Should().Be(typeof(int));
```
jonsequitur: It's better to do this with a single assertion:

```cs
rootCommandArgument.Select(a => a.Type).Should().BeEquivalentSequenceTo(typeof(string), typeof(int));
```

PR #565, opened 2019-06-18
Support multiple arguments in DragonFruit

xavierjohn:
All arguments that ends with "Arguments",  "Argument",  "Args" are required.

Example:
`static void Main(string destinationArgument, string sourceArgument)`

### DIFF
--- a/src/System.CommandLine.DragonFruit.Tests/ConfigureFromMethodTests.cs
+++ b/src/System.CommandLine.DragonFruit.Tests/ConfigureFromMethodTests.cs
@@ -221,6 +221,19 @@ namespace System.CommandLine.DragonFruit.Tests
                    .NotContain(o => o.Argument.ArgumentType == type);
         }
 
+        [Fact]
+        public void Parameters_with_Argument_suffix_is_treated_as_arguments()
+        {
+            var parser = new CommandLineBuilder()
+             .ConfigureRootCommandFromMethod(GetMethodInfo(nameof(Method_having_two_arguments)))
+             .Build();
+
+            var rootCommandArgument = parser.Configuration.RootCommand.Arguments.ToList();
+            rootCommandArgument.Should().HaveCount(2);
+            rootCommandArgument.Select(a => a.Name).Should().BeEquivalentSequenceTo("destinationArgument", "sourceArgument");
+            rootCommandArgument.Select(a => a.Type).Should().BeEquivalentSequenceTo(typeof(string), typeof(int));
+        }
+
         internal void Method_taking_bool(bool value = false)
         {
             _receivedValues = new object[] { value };
@@ -273,6 +286,10 @@ namespace System.CommandLine.DragonFruit.Tests
         }
 
         internal void Method_having_FileInfo_array_args(string stringOption, int intOption, FileInfo[] args)
+        {
+        }
+
+        internal void Method_having_two_arguments(string destinationArgument, int sourceArgument)
         {
         }
 
--- a/src/System.CommandLine.DragonFruit/CommandLine.cs
+++ b/src/System.CommandLine.DragonFruit/CommandLine.cs
@@ -103,13 +103,6 @@ namespace System.CommandLine.DragonFruit
             object target = null) =>
             command.ConfigureFromMethod(method, () => target);
 
-        private static readonly string[] _argumentParameterNames =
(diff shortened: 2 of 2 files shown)


src/System.CommandLine.DragonFruit.Tests/ConfigureFromMethodTests.cs
```diff
@@ -221,6 +221,19 @@ public void Options_are_not_built_for_infrastructure_types_exposed_by_method_par
...
         }
 
+        [Fact]
+        public void Parameters_with_Argument_suffix_is_treated_as_arguments()
+        {
+            var parser = new CommandLineBuilder()
+             .ConfigureRootCommandFromMethod(GetMethodInfo(nameof(Method_having_two_arguments)))
+             .Build();
+
+            var rootCommandArgument = parser.Configuration.RootCommand.Arguments.ToList();
```
Keboo:
```suggestion
            var rootCommandArguments = parser.Configuration.RootCommand.Arguments.ToList();
```

src/System.CommandLine.DragonFruit.Tests/ConfigureFromMethodTests.cs
```diff
@@ -221,6 +221,19 @@ public void Options_are_not_built_for_infrastructure_types_exposed_by_method_par
                    .NotContain(o => o.Argument.ArgumentType == type);
         }
 
+        [Fact]
+        public void Parameters_with_Argument_suffix_is_treated_as_arguments()
```
Keboo: This is great, but I would also include tests for when the argument names end with "arguments" and "args"

jonsequitur: The thing here that will need more work is that the method parameter name will be, e.g., `fooArg` but the `Argument.Name` (as reflected in help) would be `foo`. This means binding will need customized behavior as well.

Thinking this through led me back to the possibility of introducing attributes, and then perhaps solving the alias problem (#525) at the same time. I've been reluctant to do this before but maybe it's a the more evolvable approach.

xavierjohn: Can we just have one suffix (‘Argument’ or ‘Arg’)?

jonsequitur: I think the question first is whether we would need both a naming convention and an attributed approach. They would be redundant within one app model. This really would expand DragonFruit into a new direction, which might be good, but would certainly become more complex.

xavierjohn: Both are good, right? Just like asp.net ... you can use Attribute or Convention.

jonsequitur: Yes, but a lot more complexity.

xavierjohn: I personally prefer By Convention.

dashesy:
Similar to [this PR](https://github.com/dotnet/command-line-api/pull/525) is there a way to separate arguments using Xml Doc, instead of parameter name?
Having an argument for a `name` named just `argument` is rather not elegant, I'd rather call it something like `name`

jonsequitur:
@dashesy The named method parameters are used to generate options, not arguments, so what's needed (and part of why this PR is on hold) is a convention or API that distinguishes these. By and large, argument names aren't important to the user except to make help more informative.

dashesy:
What I mean is that, currently this will be in the help

```
Arguments:
  <argument>    ImageTagger, ImageLabeler. (default = ImageTagger).
```

But I would like this

```
Arguments:
  <detectorType>    ImageTagger, ImageLabeler. (default = ImageTagger).
```

1. Finding the default argument and adding it to the help would have been nice too
2. Furthermore, whereas this is the only format now:

```
Usage:
  TesterApp [options] <argument>
```

I would like to have:

```
Usage:
  TesterApp <argument> [options]
```

If I specify `argument` before `options` in `Main` like 

```
static void Main(string argument='default', string option1='something')
```

to be different from

```
static void Main( string option1='something', string argument='default')
```

Issue #573, opened 2019-06-25
Ability for command handler to use existing object instance


robsonj:
Currently on `public static void ConfigureFromMethod(this Command command, MethodInfo method, Func<object> target)` a Func<object> is passed in order to provide an existing object target.

It seems that the object returned from this Func<> (actually the Func<> isn't even used) is not utilized. Instead a new default construct-able target is created. This causes issues when integrating into an IoC container, where we create the handler object ahead of time and would like the handler method on an existing instance (the target) to be called.

PR #574, opened 2019-06-26
Added support for binding MethodInfo with target object in CommandHandler.Create and Command.ConfigureFromMethod

xuhongxu96:
Sometimes the command handler is a member method of a class instance, and it has too many parameters to use the generic create methods.

`Create from MethodInfo` cannot be used because `MethodInfo` lacks the `Target` object, i.e. `this`.
So, adding `Create from Delegate` to `CommandHandler` will be helpful.


xuhongxu96:
I think this PR is related to the issue #573 .
However I didn't add overloading with `Delegate` parameter for `ConfigureFromMethod`.

xuhongxu96:
#573 should be also fixed now.

Issue #575, opened 2019-06-27
TryConvertArgument is invoked twice


couven92:
When building a command-line pipeline with an Argument that is instanciated with a `TryConvertArgument` delegate, the conversion delegate is invoked twice during pipeline invocation.

In addition to doing possible expensive argument parsing twice, this also leads to validation messages being output twice on parsing failure.

I have opened Pull Request #576 for adding unit tests validating the expected behaviour. These added unit tests currently fail.

### Example

``` csharp
public static Task<int> Main(string[] args)
{
    var parser = new CommandLineBuilder(
        new RootCommand
        {
            Handler = CommandHandler.Create<int>(Run)
        })
        .AddOption(new Option(new[] { "-v", "--value" })
        {
            Argument = new Argument<int>(TryConvertHexArgument)
        })
        .UseDefaults()
        .Build();

    return parser.InvokeAsync(args);

    static bool TryConvertHexArgument(SymbolResult result, out int value)
    {
        Console.WriteLine("Invoked conversion method: " + nameof(TryConvertHexArgument));

        const string hexPrefix = "0x";

        var arg = result.Token.Value ?? string.Empty;
        if (arg.StartsWith(hexPrefix, StringComparison.OrdinalIgnoreCase))
            arg = arg.Substring(hexPrefix.Length);
        return int.TryParse(arg, NumberStyles.HexNumber, CultureInfo.InvariantCulture, out value);
    }

    static void Run(int value) => Console.WriteLine($"{nameof(Run)}: {value}");
}
```

Invoking the program above:

``` 
$ dotnet run -- -value 0x42
Invoked conversion method: TryConvertHexArgument
Invoked conversion method: TryConvertHexArgument
Run: 66
```

As can be seen from the program output, the `TryConvertHexArgument` method is invoked twice. 

I suspect this happens once during parsing, and once during binding. I also suspect that this is true for the inferred and default argument conversions (e.g. `Argument<int>`, `Argument<FileInfo>`, etc.), but we do not notice, since the default converters are silent.

PR #576, opened 2019-06-27
Add Unit test for ensuring single argument conversion invocation

couven92:
Adds unit test to ensure that argument conversion methods are only invoked once.

At time of writing these unit tests fail. Track issue #575


couven92:
Partly included in #585

PR #579, opened 2019-07-02
Added synchronous API overloads 

### LINKED ISSUE #412: `Parser.InvokeAsync` forces async methods to use the API
I am adding command line parsing support for a tool and my only option to run the parse / invoke logic is to use `Parser.InvokeAsync`. Is it a deliberate decision to not have a `Parser.Invoke`?  For a simple tool I just want to run the thing synchronously and then immediately exit.

Keboo:
Fixes #412

This add a synchronous overload for InvocationPipeleine.Invoke. This change propagates up to Parser.Invoke in System.CommandLine and up to CommandLine.ExecuteAssembly in SystemCommnadLine.DragonFruit

### DIFF
--- a/src/System.CommandLine/Invocation/InvocationPipeline.cs
+++ b/src/System.CommandLine/Invocation/InvocationPipeline.cs
@@ -33,7 +33,7 @@ namespace System.CommandLine.Invocation
 
             InvocationMiddleware invocationChain = BuildInvocationChain(context);
 
-            Task.Run(() => invocationChain(context, invocationContext => Task.CompletedTask)).Wait();
+            Task.Run(() => invocationChain(context, invocationContext => Task.CompletedTask)).GetAwaiter().GetResult();
 
             return GetResultCode(context);
         }
--- a/src/System.CommandLine/Invocation/InvocationPipeline.cs
+++ b/src/System.CommandLine/Invocation/InvocationPipeline.cs
@@ -18,9 +18,28 @@ namespace System.CommandLine.Invocation
 
         public async Task<int> InvokeAsync(IConsole console = null)
         {
-            var context = new InvocationContext(parseResult,
-                                                console);
+            var context = new InvocationContext(parseResult, console);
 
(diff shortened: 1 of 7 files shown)


src/System.CommandLine/Invocation/InvocationPipeline.cs
```diff
@@ -18,9 +18,28 @@ public InvocationPipeline(ParseResult parseResult)
...
+            return GetResultCode(context);
+        }
+
+        public int Invoke(IConsole console = null)
+        {
+            var context = new InvocationContext(parseResult, console);
+
+            InvocationMiddleware invocationChain = BuildInvocationChain(context);
+
+            Task.Run(() => invocationChain(context, invocationContext => Task.CompletedTask)).Wait();
```
couven92: Do **NOT** use `Wait()`! If anything this should be

`invocationChain(context, invocationContext => Task.CompletedTask).GetAwaiter().GetResult();`

That's the most correct and least error-prone way to execute asyncs synchronously.

couven92: ref.: https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#avoid-using-taskresult-and-taskwait

couven92: I realize we are not in ASP.NET, so the issues David points out in regards to thread blocking can be ignored in this case.

However, `Task.Wait()` will throw an `AggregateException` if the invocation throws. `.GetAwaiter().GetResult()` will actually throw the original exception. 

This is also the reason why the compiler lowers `static async Task Main() {}` to `static void Main() => MainAsync().GetAwaiter().GetResult()`. If the async main throws, you still get the original exception thrown into your face, not a wrapping `AggregateException`.

Keboo: Fixed. Thank you.

PR #595, opened 2019-07-27
Hide hidden args from options and subcommand help

### LINKED ISSUE #589: Hidden arguments are displayed by HelpBuilder for subcommands
Build: 0.3.0-alpha.19317.1

When generating help for the parent command, hidden arguments are shown for subcommands.

> CommandLine:
>   Testing help builder.
> 
> Usage:
>   CommandLine [command]
> 
> Commands:
>   outer \<test name\>    command Test description.

"test name" is the name of a hidden argument for the "outer" subcommand. Since it's marked hidden, I don't believe it should be written out here.

I _believe_, but not certain (I only just started using this library) that the problem is in lines 333-346 of HelpBuilder.cs :

```if (ShouldShowHelp(symbol))
            {
                foreach (var argument in symbol.Arguments())
                {
                    if (!string.IsNullOrWhiteSpace(argument.Name))
                    {
                        var argumentDescriptor = ArgumentDescriptor(argument);
                        if (!string.IsNullOrWhiteSpace(argumentDescriptor))
                        {
                            option = $"{option} <{argumentDescriptor}>";
                        }
                    }
                }
            }
```

Icontech:
Added call to `ShouldShowHelp` method in `OptionFormatter` to evaluate whether an argument should be included in the Options or Subcommand help sections.
Added two test cases to verify this functionality.

A suggestion would be to reconsider the name of the `OptionFormatter` method, since it is used not only for options, but also for commands that may take arguments.

This fixes #589

### DIFF
--- a/src/System.CommandLine.Tests/Help/HelpBuilderTests.cs
+++ b/src/System.CommandLine.Tests/Help/HelpBuilderTests.cs
@@ -1010,6 +1010,36 @@ namespace System.CommandLine.Tests.Help
             _console.Out.ToString().Should().Contain(expected);
         }
 
+        [Fact]
+        public void Options_section_does_not_contain_hidden_argument()
+        {
+            var command = new Command("the-command", "Does things.");
+            var opt1 = new Option("option1")
+            {
+                Argument = new Argument<int>()
+                {
+                    Name = "the-hidden",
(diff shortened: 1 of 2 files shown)


jonsequitur:
Very nice. Thanks!

PR #597, opened 2019-07-29
Fix typo and extra whitespace in .target file

fdbeirao:
While investigating how DragonFruit does its awesome magic of re-writing the user's `main` method, I came across a typo and what I perceive as an unnecessary whitespace in the `<_GeneratedProgramFileContent>` element.


jonsequitur:
Thanks!

Issue #607, opened 2019-08-02
Argument help section shows hidden arguments


Icontech:
The `AddArguments` method in `HelpBuilder.cs` makes a call to `ShouldDisplayArgumentHelp` to determine whether any arguments exist that should be added to the Arguments help section. If this is the case, the command containing the arguments is passed to the `ArgumentFormatter` method. The problem is that this method creates help items for all arguments, not just the ones that should be visible in the help section.

There is a test case (`Arguments_section_does_not_contain_hidden_argument`) aiming to verify the functionality, but since it only uses one (hidden) argument, it fails to catch this bug.

Possible fix: Add a call to `ShouldShowHelp` for each argument in `ArgumentFormatter`, to determine whether a help item should be created.

PR #608, opened 2019-08-03
Remove hidden arguments from arguments section

Icontech:
Added call to `ShouldShowHelp` in `ArgumentFormatter` method to determine what arguments should be included in Arguments help section.
Updated existing test case accordingly.
This fixes issue #607.


jonsequitur:
Thanks for working through all of these details on help output!

Icontech:
No problem, I'm happy to contribute! :)

PR #644, opened 2019-09-27
Fix markdown formatting

Ebram-Tharwat:
Fix markdown formatting for documeting "Binding to a command handler"

### DIFF
--- a/samples/tutorial/BindingToCommandHandlers.md
+++ b/samples/tutorial/BindingToCommandHandlers.md
@@ -5,35 +5,40 @@
 
 The simplest way to bind command line input is to set the `Command.Handler` property. The `System.CommandLine` model binder will look at the options and arguments for the command and attempt to match them to the parameters of the specified handler method. The default convention is that parameters are matched by name, so in the following example, option `--an-int` matches the parameter named `anInt`. Matching ignores hyphens (and other option prefixes, such as `'/'`) and is case insensitive.
 
-```cs --source-file ./src/Binding/HandlerBindingSample.cs --project ./src/Binding/Binding.csproj --region MultipleArgs 
+```cs
+--source-file ./src/Binding/HandlerBindingSample.cs --project ./src/Binding/Binding.csproj --region MultipleArgs 
 ```
 
 ## Booleans (flags)
 
 If `"true"` or `"false"` is passed for an option having a `bool` argument, it is parsed and bound as expected. But an option whose argument type is `bool` doesn't require an argument to be specified. The presence of the option token on the command line, with no argument following it, results in a value of `true`. You can see various examples here:
 
-```cs  --source-file ./src/Binding/HandlerBindingSample.cs --project ./src/Binding/Binding.csproj --region Bool 
+```cs
+--source-file ./src/Binding/HandlerBindingSample.cs --project ./src/Binding/Binding.csproj --region Bool 
 ```
 
 ## Enums
 
 You can bind `enum` types as well. The values are bound by name, and the binding is case insensitive:
 
-```cs --source-file ./src/Binding/HandlerBindingSample.cs --project ./src/Binding/Binding.csproj --region Enum 
+```cs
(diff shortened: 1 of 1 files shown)


samples/tutorial/BindingToCommandHandlers.md
```diff
@@ -5,35 +5,40 @@
 
 The simplest way to bind command line input is to set the `Command.Handler` property. The `System.CommandLine` model binder will look at the options and arguments for the command and attempt to match them to the parameters of the specified handler method. The default convention is that parameters are matched by name, so in the following example, option `--an-int` matches the parameter named `anInt`. Matching ignores hyphens (and other option prefixes, such as `'/'`) and is case insensitive.
 
-```cs --source-file ./src/Binding/HandlerBindingSample.cs --project ./src/Binding/Binding.csproj --region MultipleArgs 
+```cs
```
sharwell: 📝 `cs` doesn't seem like the correct language selection for this block. With that said, I'm not sure what the _right_ language to use is either.

omajid: Shouldn't this be `shell` not `cs`? I am not even sure `cs` as a language keyword is recognized by github.

Ebram-Tharwat: @sharwell , I agree.
In that case I would suggest:
1- Use `console`, which is already being used in _README.md_.
2- Leave it empty, which give the same _look & feel_.

Ebram-Tharwat: I guess it should be [csharp](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L542)

omajid: Saw, your previous comment; I agree with `console`/`shell` (they are different in the yml, but I dont know what the difference is).

jonsequitur: This is a Try .NET code block so it's C# code. The intention is that the source code can be filled in and will display correctly here, but when run using `dotnet-try`, the code can be edited and run in the browser. In the future, filling in the code from the specified source location will be automatable (https://github.com/dotnet/try/issues/214).

I believe `cs` is acceptable. Anyhow, experimenting with it seems to produce the same coloring as `csharp`, but if `csharp` is more correct or common we can change that.

jonsequitur: So this would ideally be as follows in order to be both functionally correct for Try .NET while also looking correct in GitHub's Markdown file viewer:

````
```csharp  --source-file ./src/Binding/HandlerBindingSample.cs --project ./src/Binding/Binding.csproj --region MultipleArgs
var command = new RootCommand
{
    new Option("--a-string") { Argument = new Argument<string>() },
    new Option("--an-int") { Argument = new Argument<int>() }
};

command.Handler = CommandHandler.Create(
    (string aString, int anInt) =>
    {
        Console.WriteLine($"{aString}");
        Console.WriteLine($"{anInt}");
    });

await command.InvokeAsync("--an-int 123 --a-string \"Hello world!\" ");
```
````

Ebram-Tharwat: @jonsequitur, you have a valid point giving that the sample will include/focus on C# code.
However a one could argue that this code block doesn't have any C# code _yet!_.

Ebram-Tharwat:
I would like to highlight that this PR focus on _fixing_ rendering empty code block.
empty-block

samples/tutorial/BindingToCommandHandlers.md
jonsequitur: It does if you run it with `dotnet-try`, due to the code fence annotations.

jonsequitur:
The correct fix is to fill in the empty code blocks with the code referenced by the existing code fence annotations. The annotations themselves shouldn't be changed, other than maybe changing `cs` to `csharp`. The annotations are used to enable the Try .NET experience:

image

jonsequitur:
@Ebram-Tharwat Feel free to reopen if you're interested in working on the fix described above, but I understand it's a bit more work than you might have intended to sign up for.

PR #655, opened 2019-10-09
Add support for unbundling with value

### LINKED ISSUE #39: Support argument assignment immediately following option name
* Asignment of option values via ''
i.e. 'getuser.exe -nInigo`
Note, this is supported by POSIX

Question: 
- [ ] This would presumably be mutually exclusive when abbreviation/startswith is active?
- [ ] Does this only work for single character names or can a full name be used?

Alxandr:
Add support for parsing `-lfoo.so` as `-l foo.so` and `-vlfoo.so` as
`-v -l foo.so`.

Resolves #39

### DIFF
--- a/src/System.CommandLine.Tests/ParserTests.cs
+++ b/src/System.CommandLine.Tests/ParserTests.cs
@@ -402,6 +402,69 @@ namespace System.CommandLine.Tests
                   .BeEquivalentTo("-a", "-b", "-c");
         }
 
+        [Theory]
+        [InlineData("-abcvalue", null, "value")]
+        [InlineData("-abc value", null, "value")]
+        [InlineData("-abc=value", null, "value")]
+        [InlineData("-abc:value", null, "value")]
+        [InlineData("-abvalue", "value", null)]
+        [InlineData("-ab value", "value", null)]
+        [InlineData("-ab=value", "value", null)]
+        [InlineData("-ab:value", "value", null)]
+        [InlineData("-abvalue -cvalue", "value", "value")]
+        // a should always be set
+        // b should always be set but value is optional
+        // c is optional
+        public void Last_bundled_option_can_accept_arguments(string arguments, string bvalue, string cvalue)
+        {
+            var optionA = new Option("-a");
+            var optionB = new Option("-b")
+            {
+                Argument = new Argument<string>
+                {
+                    Arity = ArgumentArity.ZeroOrOne
+                }
+            };
+            var optionC = new Option("-c") 
+            {
+                Argument = new Argument<string>
+                {
+                    Arity = ArgumentArity.ExactlyOne
+                }
+            };
+
+            var command = new RootCommand
+            {
+                optionA,
+                optionB,
+                optionC
+            };
+
+            var result = command.Parse(arguments);
+            result.HasOption(optionA).Should().BeTrue();
+            result.HasOption(optionB).Should().BeTrue();
+
+            if (bvalue != null)
+            {
+                result.FindResultFor(optionB)
+                    .Tokens
+                    .Should()
+                    .ContainSingle(t => t.Value == bvalue);
+            }
+
+            if (cvalue != null)
+            {
+                result.FindResultFor(optionC)
+                    .Tokens
+                    .Should()
+                    .ContainSingle(t => t.Value == cvalue);
+            }
+            else
+            {
+                result.HasOption(optionC).Should().BeFalse();
+            }
+        }
+
         [Fact]
         public void Parser_root_Options_can_be_specified_multiple_times_and_their_arguments_are_collated()
         {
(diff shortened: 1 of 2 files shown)


src/System.CommandLine/StringExtensions.cs
```diff
@@ -173,13 +174,81 @@ bool CanBeUnbundled(string arg)
...
-                       arg.RemovePrefix()
-                          .All(CharacterIsValidOptionAlias);
+                       TryUnbundle(arg.RemovePrefix(), out replacement);
+
+                Token TokenForOptionAlias(char c) =>
+                    argumentDelimiters.Contains(c) 
+                    ? null
+                    : knownTokens.SingleOrDefault(t => 
+                        t.Type == TokenType.Option && 
+                        t.Value.RemovePrefix() == c.ToString());
```
Keboo: I don't believe this check is valid since it is possible for tokens to differ by only their prefix. This does raise the question of what should be done when this case occurs.

src/System.CommandLine/StringExtensions.cs
```diff
@@ -173,13 +174,81 @@ bool CanBeUnbundled(string arg)
...
+                                return false;
+                            }
+                        }
 
-                bool CharacterIsValidOptionAlias(char c) =>
-                    knownTokens.Where(t => t.Type == TokenType.Option)
-                               .Select(t => t.Value.RemovePrefix())
-                               .Contains(c.ToString());
+                        var option = currentCommand?.Children.GetByAlias(token.Value) as IOption;
+                        builder.Add("-" + arg[i]);
```
Keboo: Because the prefixes are configurable, would it be better to use `configuration.Prefixes.First()` instead of `"-"`?

src/System.CommandLine.Tests/ParserTests.cs
```diff
@@ -402,6 +402,69 @@ public void Optional_option_arguments_are_unbundled()
...
+                optionA,
+                optionB,
+                optionC
+            };
+
+            var result = command.Parse(arguments);
+            result.HasOption(optionA).Should().BeTrue();
+            result.HasOption(optionB).Should().BeTrue();
+
+            if (bvalue != null)
```
Keboo: Can we break this up into multiple unit test so the tests don't need to be conditional?

src/System.CommandLine/StringExtensions.cs
Alxandr: This is how the old functionality worked though? Alternative prefixes should maybe disable unbundeling?

src/System.CommandLine.Tests/ParserTests.cs
Alxandr: Will do. Just thought this was more DRY.

src/System.CommandLine/StringExtensions.cs
Keboo: @jonsequitur any feedback on what is best here? This looks like an area where POSIX behavior conflicts with the desire to make invocation east for command line app user.

src/System.CommandLine.Tests/ParserTests.cs
Alxandr: I've unbundled the unbundling test 😛.

src/System.CommandLine/StringExtensions.cs
Alxandr: How about `token.Value`, given that I have a known token already?

Alxandr:
A test failed on the last run, but I can't see which one in the ADO portal (seems the test run didn't upload the test results), and it only happened on Windows which I don't currently have access to. Can anyone figure out what test failed and why?

src/System.CommandLine/StringExtensions.cs
jonsequitur: I think it's reasonable to only support unbundling on `"-"` and not on other prefixes.

src/System.CommandLine/StringExtensions.cs
jonsequitur: Supporting unbundling only on `"-"` simplifies this as well I think.

src/System.CommandLine/StringExtensions.cs
Alxandr: So this can be considered resolved?

src/System.CommandLine/StringExtensions.cs
Alxandr: Would `token.Value` still make more sense?

src/System.CommandLine/StringExtensions.cs
jonsequitur: I think so, yes.

jonsequitur:
@Alxandr I'm good to merge this. Was there anything more you wanted to add?

Alxandr:
@jonsequitur No, I'm good. I was waiting for resolutions on the prefix issues. I just missed the comment you posted 5 days ago.

jonsequitur:
Thanks, @Alxandr. Great PR!

PR #660, opened 2019-10-21
Do not set LangVersion in downstream projects

### LINKED ISSUE #594: Forcing C# version to 7.1 messes up a project on .NET Core 3/C#8
The targets file in the NuGet package forces the version to 7.1 if it isn't set. This is at odds with "new" .NET SDK projects where the language version is set as a byproduct of the target framework. It should be made aware of these new kind of projects.

sharwell:
Fixes #594

📝 Please do not rebase or squash this pull request during the merge.


sharwell:
Awesome! As a quick update, I reviewed the history of this line into #93, and was unable to find an indication of why it was added. It seemed incidental, meaning there would not be any downside of simply removing it.

jonsequitur:
This was added because the generated entry point is async, which requires 7.1 or higher.

sharwell:
Ah, we should be good then. If the user isn't using 7.1, we have a code fix for the compiler error which will offer to upgrade their project.

jonsequitur:
The potential issue here is that the compile error will be inside a generated file: https://github.com/dotnet/command-line-api/blob/52463e66eb1cf36845f5980bc3d640d8b0274ba7/src/System.CommandLine.DragonFruit/targets/System.CommandLine.DragonFruit.targets#L37

sharwell:
If there is a specific code fix case that is not working, please post a bug to dotnet/roslyn so we can work out a solution from that side. 👍

PR #664, opened 2019-10-23
Add GetInvocationContext extension methods

couven92:
When using the hosting extensions, you normally need a way to access the `ParseResult` or the `InvocationContext` to actually be able to evaluate the parsed command-line arguments.

Sure, `ParseResult` is registered as a DI-service, so any DI class constructor can accept a `ParseResult` and get it injected, but that approach is not necessary the most typical way to handle this in a Hosted application architecture (like ASP.NET).

You could argue that command-line arguments fit better when transferred to an `IOptions` instance, like most other configurable parts in the Generic Host do (e.g. `ConsoleLiftetimeOptions`, `LoggingOptions`, etc.).

So why not just call `services.ConfigureServices(services => services.Configure<CommandLineOptions>(opts => { }))` in your `ConfigureHost` delegate? Well you very quickly discover that accessing the `ParseResult` instance when configuring the DI container is quirky.

This PR propses to add extension methods to `IHostBuilder` and `HostBuilderContext` to easily obtain the `InvocationContext` (and from there the `ParseResult`) for scenarios within `ConfigureHost`.

Originally, to configure my `MyCommandLineOptions` options, you'd currently have to do this:
``` C#
.UseHost(host =>
{
  InvocationContext invocation = null;
  if (host.Properties.TryGetValue(typeof(InvocationContext), out invocationObj))
    invocation = invocationObj as InvocationContext;
  var parseResult = invocation?.ParseResult;
  {
    services.Configure<MyCommandLineOptions>(opts =>
    {
      // Do something with parseResult here...
    });
  });
})
```

*Alternatively you could also access the `Properties` from the `HostBuilderContext` argument that you can get if you call the other overload of `ConfigureServices`.*

With this PR, this code gets more readable and you can then do this:

``` C#
.UseHost(host =>
{
  host.ConfigureServices((context, services) =>
  {
    var parseResult = context.GetInvocationContext()?.ParseResult;
    services.Configure<MyCommandLineOptions>(opts =>
    {
      // Do something with parseResult here...
    });
  });
})
```

### DIFF
--- a/src/System.CommandLine.Hosting.Tests/HostingTests.cs
+++ b/src/System.CommandLine.Hosting.Tests/HostingTests.cs
@@ -8,7 +8,7 @@ using FluentAssertions;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
-
+using Microsoft.Extensions.Options;
 using Xunit;
 
 namespace System.CommandLine.Hosting.Tests
@@ -190,6 +190,40 @@ namespace System.CommandLine.Hosting.Tests
             parser.InvokeAsync(commandLine).GetAwaiter().GetResult();
 
             testConfigValue.Should().BeEquivalentTo(testValue);
+        }
+
+        [Fact]
+        public static void UseHost_binds_parsed_arguments_to_options()
+        {
+            const int myValue = 4224;
+            string commandLine = $"-{nameof(MyOptions.MyArgument)} {myValue}";
+
+            var rootCmd = new RootCommand();
+            rootCmd.AddOption(
+                new Option($"-{nameof(MyOptions.MyArgument)}")
+                { Argument = new Argument<int>() }
+                );
+            rootCmd.Handler = CommandHandler.Create((IHost host) =>
+            {
+                var options = host.Services.GetRequiredService<IOptions<MyOptions>>();
+                Assert.Equal(myValue, options.Value.MyArgument);
+            });
+
+            Assert.Equal(0, new CommandLineBuilder(rootCmd)
+                .UseHost(host =>
+                {
+                    host.ConfigureServices(services =>
+                    {
+                        services.AddOptions<MyOptions>().BindCommandLine();
+                    });
+                })
+                .Build()
+                .Invoke(commandLine));
+        }
+
+        private class MyOptions
+        {
+            public int MyArgument { get; set; }
         }
     }
 }
--- a/src/System.CommandLine.Hosting.Tests/HostingTests.cs
+++ b/src/System.CommandLine.Hosting.Tests/HostingTests.cs
@@ -197,7 +197,7 @@ namespace System.CommandLine.Hosting.Tests
         {
             const int myValue = 4224;
             string commandLine = $"-{nameof(MyOptions.MyArgument)} {myValue}";
-            IOptions<MyOptions> options = null;
+            MyOptions options = null;
 
             var rootCmd = new RootCommand();
             rootCmd.AddOption(
@@ -206,7 +206,9 @@ namespace System.CommandLine.Hosting.Tests
                 );
             rootCmd.Handler = CommandHandler.Create((IHost host) =>
             {
-                options = host.Services.GetRequiredService<IOptions<MyOptions>>();
+                options = host.Services
+                    .GetRequiredService<IOptions<MyOptions>>()
+                    .Value;
             });
 
             int result = new CommandLineBuilder(rootCmd)
@@ -222,8 +224,7 @@ namespace System.CommandLine.Hosting.Tests
 
             Assert.Equal(0, result);
             Assert.NotNull(options);
-            Assert.NotNull(options.Value);
-            Assert.Equal(myValue, options.Value.MyArgument);
+            Assert.Equal(myValue, options.MyArgument);
         }
 
         private class MyOptions
--- a/src/System.CommandLine.Hosting.Tests/HostingTests.cs
+++ b/src/System.CommandLine.Hosting.Tests/HostingTests.cs
@@ -197,6 +197,7 @@ namespace System.CommandLine.Hosting.Tests
         {
             const int myValue = 4224;
             string commandLine = $"-{nameof(MyOptions.MyArgument)} {myValue}";
+            IOptions<MyOptions> options = null;
 
             var rootCmd = new RootCommand();
             rootCmd.AddOption(
@@ -205,11 +206,10 @@ namespace System.CommandLine.Hosting.Tests
                 );
             rootCmd.Handler = CommandHandler.Create((IHost host) =>
             {
-                var options = host.Services.GetRequiredService<IOptions<MyOptions>>();
-                Assert.Equal(myValue, options.Value.MyArgument);
+                options = host.Services.GetRequiredService<IOptions<MyOptions>>();
             });
 
-            Assert.Equal(0, new CommandLineBuilder(rootCmd)
+            int result = new CommandLineBuilder(rootCmd)
                 .UseHost(host =>
                 {
                     host.ConfigureServices(services =>
@@ -218,7 +218,12 @@ namespace System.CommandLine.Hosting.Tests
                     });
                 })
                 .Build()
-                .Invoke(commandLine));
+                .Invoke(commandLine);
+
+            Assert.Equal(0, result);
+            Assert.NotNull(options);
+            Assert.NotNull(options.Value);
+            Assert.Equal(myValue, options.Value.MyArgument);
         }
 
         private class MyOptions
--- a/src/System.CommandLine.Hosting/HostingExtensions.cs
+++ b/src/System.CommandLine.Hosting/HostingExtensions.cs
@@ -67,19 +67,24 @@ namespace System.CommandLine.Hosting
             });
         }
 
-        public static InvocationContext GetInvocationContext(this IHostBuilder hostBuilder) =>
-            GetInvocationContext(hostBuilder?.Properties);
-
-        public static InvocationContext GetInvocationContext(this HostBuilderContext context) =>
-            GetInvocationContext(context?.Properties);
-
-        private static InvocationContext GetInvocationContext(IDictionary<object, object> properties)
+        public static InvocationContext GetInvocationContext(this IHostBuilder hostBuilder)
         {
-            object ctxObj = null;
-            if (properties?.TryGetValue(typeof(InvocationContext), out ctxObj) ?? false)
-                return ctxObj as InvocationContext;
-            return null;
+            if (hostBuilder is null)
+                throw new ArgumentNullException(nameof(hostBuilder));
+            return GetInvocationContext(hostBuilder.Properties);
         }
+
+        public static InvocationContext GetInvocationContext(this HostBuilderContext context)
+        {
+            if (context is null)
+                throw new ArgumentNullException(nameof(context));
+            return GetInvocationContext(context.Properties);
+        }
+
+        private static InvocationContext GetInvocationContext(IDictionary<object, object> properties) => 
+            properties.TryGetValue(typeof(InvocationContext), out object ctxObj)
+                ? ctxObj as InvocationContext
+                : null;
 
         public static void ConfigureFromCommandLine<TOptions>(
             this IServiceCollection services, HostBuilderContext context)
--- a/src/System.CommandLine.Hosting/HostingExtensions.cs
+++ b/src/System.CommandLine.Hosting/HostingExtensions.cs
@@ -81,7 +81,7 @@ namespace System.CommandLine.Hosting
             return null;
         }
 
-        public static void ConfigureFromInvocation<TOptions>(
+        public static void ConfigureFromCommandLine<TOptions>(
             this IServiceCollection services, HostBuilderContext context)
             where TOptions : class
         {
--- a/src/System.CommandLine.Hosting/HostingExtensions.cs
+++ b/src/System.CommandLine.Hosting/HostingExtensions.cs
@@ -1,4 +1,5 @@
-﻿using System.CommandLine.Builder;
+﻿using System.Collections.Generic;
+using System.CommandLine.Builder;
 using System.CommandLine.Invocation;
 using System.Linq;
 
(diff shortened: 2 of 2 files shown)


jonsequitur:
At this layer it seems simpler to use the model binder to directly materialize your options class rather than leaving people to pull the values one at a time off of the `ParseResult`. The wire-up could be entirely transparent and implemented via the DI configuration, meaning this should work:

```csharp
var myOptions = services.GetService<MyCommandLineOptions>();
```

and `MyCommandLineOptions` should be injectable everywhere that other services are.

couven92:
@jonsequitur No this during the set-up of the DI container. You cannot materialize DI services while configuring them. The only other way to configure Options is to register a singleton service of type `IPostConfigureOptions<MyCommandLineOptions>` and give it a class implementing that interface. In that class constructor you can get any DI-service you'd like injected the usual way.

jonsequitur:
Understood. My comment was more about the use of `ParseResult`. It should not be needed. You should be able to bind an instance of `MyCommandLineOptions` without ever directly accessing a `ParseResult`.

couven92:
@jonsequitur you mean as arguments to the handler, right? Yeah that works anyways.

But if you actually want a typed options instance making use of the `Microsoft.Extension.Options` infrastructure, you'll need to somehow bind the parsed arguments to an Options instance. You'd typically do that using the `service.Configure<TOptions>(Action<TOptions> configreOptions)` method.

I view the `ParseResult` -> `IOptions<T>` as basically an operation similar to `IConfiguration` -> `IOptions<T>`. In the case of `IConfiguration`, you need to do the exact same thing: You need to actually write the code mapping an `IConfiguration` to an options instance yourself (either using `.Configure<TOptions>` or `IPostConfigureOptions<TOptions>`).

But very often mapping `IConfiguration` to an options instance is very easy: You simply use the `IConfiguration.Bind(object)` extension method provided by `Microsoft.Extensions.Configuration.Binder`.

I agree it would be nice to have a Binder that directly maps a `ParseResult` to an object. If we had that, you'd still need to call that method though. So there is definitely need for the easy access to the `InvocationContext` and the `ParseResult`.

couven92:
> it would be nice to have a Binder that directly maps a ParseResult to an object.

You can use `System.CommandLine.Binding.ModelBinder.UpdateInstance<T>(T, BindingContext)` for that, can't you?

Well... Just one more reason to be able to easily access the `InvocationContext`: You need the `BindingContext` from the invocation to use the binder.

jonsequitur:
> I agree it would be nice to have a Binder that directly maps a ParseResult to an object. If we had that, you'd still need to call that method though. So there is definitely need for the easy access to the InvocationContext and the ParseResult.

This exists today and unlike `IConfiguration.Bind(string key, object instance)` it does object instantiation and recursive binding via constructor calls. Using this approach will give a behavior that's more consistent with the way command handlers are bound.

jonsequitur:
To make this more concrete, I think people should be able to do something like this:

```csharp
.UseHost(host =>
{
  host.ConfigureServices((context, services) =>
  {
    services.AddFromCommandLine<MyCommandLineOptions>();
  });
})
```

The implementation of `AddFromCommandLine` would resolve the `BindingContext`, instantiate the `ModelBinder<MyCommandLineOptions>`, and call that to get an instance of `MyCommandLineOptions`. Avoiding using the `ParseResult` directly is important here because binding can pull in values from other sources than the `ParseResult`.

couven92:
@jonsequitur I agree, using the model binder together with the binding context is way better.

I suggest you'd do this:

``` C#
.UseHost(host =>
{
  host.ConfigureServices((context, services) =>
  {
    var bindingContext = context.GetInvocationContext()?.BindingContext;
    services.Configure<MyCommandLineOptions>(opts =>
    {
      var modelBinder = new ModelBinder<MyCommandLineOptions>();
      modelBinder.UpdateInstance(opts, bindingContext);
    });
  });
})
```

Because that would make on par with `IConfiguration`. Point here is that you might want to do additional stuff in the configure delegate than what is shown here, for example, you could want to also bind from the host confiuration in addition to the command-line arguments.

I can add a ConfigureFromInvocation extension with the following signature for the simple use-case whose implementation is equal to the code above:
``` C#
public static void ConfigureFromInvocation<TOptions>(this IServiceCollection services, HostBuilderContext context);
``` 

I think the `Configure` part in the name is important to reflect the fact, that we are implementing a specialization case for the already existing `Configure<TOptions>()` method.

couven92:
@jonsequitur so that makes the easy-use case even simpler:

``` C#
.UseHost(host =>
{
  host.ConfigureServices((context, services) =>
  {
    services.ConfigureFromInvocation<MyCommandLineOptions>(context);
  });
})
```

couven92:
@jonsequitur it would! I'll change the name! :)

couven92:
@jonsequitur okay, then I'd say this is ready for review :)

loic-sharma:
How should I use `AddFromCommandLine` if different commands have different options? Should I register the options for all commands?

couven92:
@loic-sharma Hmm... depends, I guess. Have common stuff in one options type and have the command specific options in another?

In that case you'd probably configure your options conditionally. Remember you can still call the convenicence `context.GetInvocationContext()` extension on the `HostBuilderContext` argument of the `ConfigureServices` delegate. E.g.:

``` C#
.UseHost(host =>
{
  host.ConfigureServices((context, services) =>
  {
    var invocation = context.GetInvocationContext();
    switch (invocation.ParseResult.CommandResult.Name)
    {
      case "cmd1":
        services.ConfigureFromInvocation<MyCommandLineOptions1>(context);
        break;
      case "cmd2":
        services.ConfigureFromInvocation<MyCommandLineOptions1>(context);
        break;
    }
  });
})
```

I personally dislike doing the switch on the name of the command (you get magic strings), but the idea is the same, right?

I'd try to keep the common stuff together in one option type and only have specific stuff in more specialized classes.

src/System.CommandLine.Hosting/HostingExtensions.cs
```diff
@@ -64,5 +66,31 @@ public static class HostingExtensions
...
+
+        private static InvocationContext GetInvocationContext(IDictionary<object, object> properties)
+        {
+            object ctxObj = null;
+            if (properties?.TryGetValue(typeof(InvocationContext), out ctxObj) ?? false)
+                return ctxObj as InvocationContext;
+            return null;
+        }
+
+        public static void ConfigureFromCommandLine<TOptions>(
```
loic-sharma: What do y'all think of `ConfigureCommand` as a name? `ConfigureFromCommandLine` is a little long

src/System.CommandLine.Hosting/HostingExtensions.cs
```diff
@@ -64,5 +66,31 @@ public static class HostingExtensions
                     services.Configure(configureOptions);
             });
         }
+
+        public static InvocationContext GetInvocationContext(this IHostBuilder hostBuilder) =>
```
loic-sharma: If `hostBuilder` is `null`, this should throw an `ArgumentNullException`. Same comment for `GetInvocationContext`.

loic-sharma: For example, from aspnet/Extensions: https://github.com/aspnet/Extensions/blob/07021213ce1947a910570263eb543f6fe01e7e09/src/Options/Options/src/ConfigureOptions.cs#L32

src/System.CommandLine.Hosting/HostingExtensions.cs
jonsequitur: The thing being configured is the `IServiceCollection`, not a command,  no?

couven92: correct, or to be more precise, it's configuring the instance of `TOptions` passed as generic type argument.

jonsequitur: To be utterly and totally pedantic:

```
ConfigureServiceProviderToUseCommandLineParseResultToCreateInstancesOf<T>
```

Now remove the words that aren't adding critical information and you have your method name.

couven92: I still think `ConfigureFromCommandLine` is best.

Technically `ConfigureFromBindingContext` would also be fairly accurate, but that would also be a little overloaded (similar to `ConfigureFromInvocation`).

couven92: And the method does not actually "configure the service provider to use command line parse result to create an instance of T". The actual method signature I am basing this on is `Configure<TOptions>() where TOptions : class, new()`.

So the implementation gets an already instantiated `TOptions` and is expected to set settable properties on that type. It does so using `ModelBinder<TOptions>.UpdateInstance(opts, bindingContext)`.

That leads the DI container to know how to resolve `IOption<TOption>` not actually directly the instance of `TOption`!

To register `TOption` you'd actually call `AddSingleton`, `AddScoped` or `AddTransient`, and all those overloads already have the possibility to resolve other registered DI services, no that we could already do before. `Configure<TOptions>` however does not a have a way of resolving DI services, that's why we needed to get the `BindingContext` from somewhere else. (*`BindingContext` is already and has always been registered as a DI singleton when `UseHost` is invoked.*)

src/System.CommandLine.Hosting/HostingExtensions.cs
couven92: https://github.com/dotnet/command-line-api/pull/664/commits/2449ac12c33a3f2a10450dd8e6be4e89d2af19eb

src/System.CommandLine.Hosting/HostingExtensions.cs
jonsequitur: > So the implementation gets an already instantiated TOptions and is expected to set settable properties on that type. It does so using ModelBinder<TOptions>.UpdateInstance(opts, bindingContext).

Out of curiosity, why favor `IOptions<T>` over simply `T` and therefore require setters over constructor parameters?

loic-sharma: > The thing being configured is the `IServiceCollection`, not a command, no?

Yes, but this lines up the rest of the ecosystem. For example, `services.Configure<T>(...)` adds a `IConfigureOptions<T>` to the collection. I don't think we should establish a new `AFromB` pattern.

> Out of curiosity, why favor IOptions<T> over simply T and therefore require setters over constructor parameters?

Also, you can setup a hierarchy of configurations. I can set some defaults from my app, let an `appsettings.json` file override the defaults, and finally let my CLI arguments do some final overrides. Silly example:

```csharp
services.Configure<CliOptions>(options =>
{
  options.A = "Default value";
});

services.Configure<CliOptions>(options =>
{
  options.A = "Override value";
});
```

This model is very powerful, especially when you have many configuration sources (defaults from app, appsettings.json for local development, Key Vault, Azure App Service settings, Azure App Configuration)

loic-sharma: Another example: Azure SDK for .NET uses `services.ConfigureDefaults` (see [this](https://github.com/Azure/azure-sdk-for-net/blob/e0bdf52750a60af46c62973423674118d67bb105/sdk/core/Microsoft.Extensions.Azure/samples/Startup.cs#L46)). The paradigm seems to be `ConfigureThing`, so I think `ConfigureCommand` follows that pattern best. Please let me know if you'd like more example, I can find more.

couven92: Still, `ConfigureFromCommandLine` is still the best and most accurate name.

couven92: okay what about this:

``` C#
.UseHost(host =>
{
  {
    services.AddOptions<MyCommandLineOptions>()
      .UseCommandLine();
  });
})
```

The change is:

* Rename method to `UseCommandLine`
* Extend `OptionsBuilder<TOptions>` instead of `IServiceCollection`

couven92: I actually like that one soo much, I am going to add it anyways 😂

couven92:
Renamed to `BindCommandLine` and it extends `OptionsBuilder<TOptions>` instead.

The new version matches much more closely with the `IConfiguration` binding. (`Microsoft.Extensions.Configuration.Binder` add extension method `Bind<TOptions>(this OptionsBuilder<TOptions>, IConfiguration)`)

https://github.com/couven92/dotnet-command-line-api/blob/1f9317bb9bc76ed55892a12516cf864c98542809/src/System.CommandLine.Hosting.Tests/HostingTests.cs#L213-L219

And for good measure I also added a test case for options binding from command line.

src/System.CommandLine.Hosting.Tests/HostingTests.cs
```diff
@@ -191,5 +191,39 @@ void Execute(IHost host)
...
+
+            var rootCmd = new RootCommand();
+            rootCmd.AddOption(
+                new Option($"-{nameof(MyOptions.MyArgument)}")
+                { Argument = new Argument<int>() }
+                );
+            rootCmd.Handler = CommandHandler.Create((IHost host) =>
+            {
+                var options = host.Services.GetRequiredService<IOptions<MyOptions>>();
+                Assert.Equal(myValue, options.Value.MyArgument);
```
jonsequitur: Generally it's a bit easier to read if you just capture the value here and assert at the end of the test. This also avoids the problem of the handler, and therefore the assertion, not being called.

couven92: @jonsequitur https://github.com/dotnet/command-line-api/pull/664/commits/aca8727feede0e8ec29ae2e2dff84f76522e46d8

couven92:
Hmm... so now my tests fail... I am investigating what went wrong...

jonsequitur:
Are the `GetInvocationContext` methods actually desired now?

couven92:
@jonsequitur Hmm... we don't need them now, I guess... But they don't really hurt. But the code necessary to get hold of the InvocationContext should you need it is a little obscure... We might leave them for easier access.

But I can remove them, just say the word 😃

jonsequitur:
Let's leave them out until we have a clear need. Even simple convenience methods have a maintenance cost over time.

couven92:
@jonsequitur I understand, I have removed the extensions that are no longer needed. All CI checks have passed :)

PR #666, opened 2019-10-23
Accept IValueDescriptor in ModelBinder

couven92:
Currently the `BindMemberFrom...` methods on `ModelBinder` and `ModelBinder<T>` currently only accept `Option` or `IOption`. Thus, binding an `Argument` will only work if the target member matches the argument name. `BindMemberFrom...` can be used to configure the binder to bind any option to any member regardless of name matching.

There is no real reason why you should not be able to use `BindMemberFrom...` with an `Argument` or `IArgument`.

* Renames internal class `OptionValueSource` to `SpecificSymbolValueSource` (matching the filename the type is located in)  
   Changes its `TryGetValue` implementation to use `CurrentSymbolValueSource` with value descriptor provided at construction.
* **Renames publically visible method `BindMemberFromOption` to `BindMemberFromValue`**
* Changes argument in `BindMemberFrom...` from `Option` and `IOption` to `IValueDescriptor`

As this will introduce a breaking change, we can also choose to not rename the public method `BindMemberFromOption`, add the new method, and forward the call from the old one to the new more general one.

### DIFF
--- a/src/System.CommandLine.Tests/Binding/ModelBinderTests.cs
+++ b/src/System.CommandLine.Tests/Binding/ModelBinderTests.cs
@@ -324,7 +324,7 @@ namespace System.CommandLine.Tests.Binding
 
             var binder = new ModelBinder<ClassWithMultiLetterSetters>();
 
-            binder.BindMemberFromOption(
+            binder.BindMemberFromValue(
                 c => c.IntOption,
                 option);
 
--- a/src/System.CommandLine.Tests/Binding/ModelBinderTests.cs
+++ b/src/System.CommandLine.Tests/Binding/ModelBinderTests.cs
@@ -377,5 +377,65 @@ namespace System.CommandLine.Tests.Binding
 
             instance.IntOption.Should().Be(42);
         }
+
+        [Fact]
+        public void PropertyInfo_can_be_bound_to_argument()
+        {
+            var command = new Command("the-command");
+            var argument = new Argument<int> { Arity = ArgumentArity.ExactlyOne };
+            command.AddArgument(argument);
+
+            var type = typeof(ClassWithMultiLetterSetters);
+            var binder = new ModelBinder(type);
+            var propertyInfo = type.GetProperty(nameof(ClassWithMultiLetterSetters.IntOption));
+
+            binder.BindMemberFromValue(propertyInfo, argument);
+
+            var bindingContext = new BindingContext(command.Parse("the-command 42"));
+
+            var instance = (ClassWithMultiLetterSetters)binder.CreateInstance(bindingContext);
+
+            instance.IntOption.Should().Be(42);
+        }
+
+        [Fact]
+        public void PropertyExpression_can_be_bound_to_option()
+        {
+            var command = new Command("the-command");
+            var option = new Option("--fred") { Argument = new Argument<int>() };
+            command.AddOption(option);
+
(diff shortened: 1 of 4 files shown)


src/System.CommandLine.Tests/Binding/ModelBinderTests.cs
```diff
@@ -319,7 +319,7 @@ public void Values_from_options_on_parent_commands_can_be_bound_regardless_of_na
 
             var binder = new ModelBinder<ClassWithMultiLetterSetters>();
 
-            binder.BindMemberFromOption(
+            binder.BindMemberFromValue(
```
jonsequitur: It's probably worthwhile to also add a test for binding from an `Argument`.

couven92:
Hmm... it seems the Beast finally cought up with me... Appearently the ModelBinder does not really like binding against Arguments. I am investigating how to make it behave

couven92:
@jonsequitur Can you look at the build? I have no idea why it fails. It claims that I do not have access to `ArgumentResult2`, which is an internal type within the same assembly and namespace. And it does not help that it build fine on my machine...

src/System.CommandLine/Binding/SpecificSymbolValueSource.cs
```diff
@@ -1,34 +1,47 @@
...
-            IValueDescriptor valueDescriptor, 
-            BindingContext bindingContext, 
+        public bool TryGetValue(IValueDescriptor valueDescriptor,
+            BindingContext bindingContext,
             out object boundValue)
         {
-            var result = bindingContext.ParseResult.FindResultFor(Option);
-
-            switch (result)
+            valueDescriptor = ValueDescriptor;
```
Keboo: This always overwrites the passed in parameter with the property. Did you mean `valueDescriptor ??= ValueDescriptor;`?

src/System.CommandLine/ParseResult.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System.Collections.Generic;
+using System.CommandLine.Binding;
```
Keboo: Is this change needed?

src/System.CommandLine/Binding/SpecificSymbolValueSource.cs
```diff
@@ -1,34 +1,47 @@
...
 namespace System.CommandLine.Binding
 {
-    internal class OptionValueSource : IValueSource
+    internal class SpecificSymbolValueSource : IValueSource
     {
-        public OptionValueSource(IOption option)
+        public SpecificSymbolValueSource(IValueDescriptor valueDescriptor)
         {
-            Option = option;
+            ValueDescriptor = valueDescriptor;
```
Keboo:
```suggestion
            ValueDescriptor = valueDescriptor ?? throw new ArgumentNullException(nameof(valueDescriptor));
```

src/System.CommandLine/ParseResult.cs
couven92: nope

src/System.CommandLine/Binding/SpecificSymbolValueSource.cs
couven92: No, But I see that the code is misleading. I updated the implementation and during that also simplified it 😃

PR #672, opened 2019-11-02
Add option count

### LINKED ISSUE #669: Multiple flag options
How do you create something that can parse `-vvv` or `--verbose --verbose --verbose`? The way to make an option take multiple values is by having it accept an array, but in this case, there is no values provided.

Alxandr:
This PR adds a `Count` property to the `OptionResult` class, which is incremented whenever that option is visited during parsing. This allows for count-like options (which works like flags, except you can repeat them).

I was mainly just going to experiment a bit to look at how #669 could be resolved, but then I figured this is a potentially pretty easy solution to the issue, and given that there is so little code change needed I figured showing it as a PR would be the simplest.

I think this might make for a good starting point to deal with count-like options, enabling the core functionality, and allowing for validation and similar to be added later.

Resolves #669

### DIFF
--- a/src/System.CommandLine.Tests/OptionTests.cs
+++ b/src/System.CommandLine.Tests/OptionTests.cs
@@ -253,6 +253,17 @@ namespace System.CommandLine.Tests
             option.Argument.Name.Should().Be("arg");
         }
 
+        [Fact]
+        public void Option_Result_Count_equals_to_occurences_of_option()
+        {
+            var option = new Option(new[] { "--verbose", "-v" });
+
+            var result = option.Parse("-vv --verbose");
+            var optionResult = result.FindResultFor(option);
+
+            Assert.Equal(3, optionResult.Count);
+        }
+
         protected override Symbol CreateSymbol(string name) => new Option(name);
     }
 }
--- a/src/System.CommandLine/OptionResult.cs
+++ b/src/System.CommandLine/OptionResult.cs
@@ -20,6 +20,8 @@ namespace System.CommandLine
 
         public IOption Option { get; }
 
+        public int Count { get; internal set; } = 1;
+
         public bool IsImplicit => Token is ImplicitToken;
 
         private protected override int RemainingArgumentCapacity
--- a/src/System.CommandLine/Parsing/ParseResultVisitor.cs
+++ b/src/System.CommandLine/Parsing/ParseResultVisitor.cs
@@ -100,9 +100,10 @@ namespace System.CommandLine.Parsing
 
         protected override void VisitOptionNode(OptionNode optionNode)
         {
-            if (_innermostCommandResult.Children.ResultFor(optionNode.Option) == null)
+            var optionResult = (OptionResult)_innermostCommandResult.Children.ResultFor(optionNode.Option);
+            if (optionResult == null)
             {
-                var optionResult = new OptionResult(
+                optionResult = new OptionResult(
                     optionNode.Option,
                     optionNode.Token,
                     _innermostCommandResult);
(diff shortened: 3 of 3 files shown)


jonsequitur:
FYI:

```csharp
var option = new Option(new[] { "--verbose", "-v" });

var result = option.Parse("-vv --verbose");

result.Tokens.Count.Should().Be(3);
```

This works right now.

Alxandr:
It does? Doesn't only the first token get added? https://github.com/dotnet/command-line-api/blob/33300b9523e0555d9e421fa048c70a89ba5398d6/src/System.CommandLine/Parsing/ParseResultVisitor.cs#L108

From my understanding, only with multiple arguments tokens gets added. I even checked references for the `AddToken` method.

Regardless, I don't think you should have to know about tokens to use an API like this, but it does mean the implementation can be something like `=> option.Tokens.Where(IsSelfSymbolCheck)`.

Alxandr:
Oh, wait. I missed the fact that you checked tokens in the root result, and not on the option. The following test fails with `optionResult.Tokens.Count` being 0.

```csharp
        [Fact]
        public void Option_Result_Count_equals_to_occurences_of_option()
        {
            var option = new Option(new[] { "--verbose", "-v" });

            var result = option.Parse("-vv --verbose");
            var optionResult = result.FindResultFor(option);

            Assert.Equal(3, optionResult.Count);
            Assert.Equal(3, optionResult.Tokens.Count);
        }
```

Regardless of whether or not it makes sense for that to be the case, I don't think I should be required to look at the tokens in the root command, because that's mostly just giving up on the framework at that point. It only really works if you just have 1 option. And also, reading the tokens is really low level, and I would not expect to be required to do that.

jonsequitur:
Yeah, it doesn't make sense to make people check the root command for the tokens. I was mainly pointing out that there's an implementation that should be usable to get the token count, but it doesn't seem to be working for this. The design for the tokens API might need some adjustment.

Alxandr:
Do you think that this PR is a viable way forward though?

jonsequitur:
I think the idea of binding to some model that gets passed to the handler is more consistent than adding a property to `OptionResult`. Most of the time I would expect people not to have to interact directly with the `ParseResult`.

Alxandr:
I agree. But I also think being able to get the count from the result might make sense even when you bind to other things. I was planning to add a bindable model that gets it's value from this count property after this pr, though I've not yet tested to see if it's viable.

Alxandr:
I'm not sure how to move forward here. Should we use tokens for this instead of explicitly tracking count? Should I add a bindable struct? Should we allow getting the count even if you bind to something else than said struct?

jonsequitur:
What would be ergonomic at the handler level? We've tended toward keeping `ParseResult` and its parts out of sight in favor of letting people bind to their own abstractions, so something like:

```csharp
command.Handler = CommandHandler.Create(VerboseCount count =>
{
    // make it verbose
});
```

Alxandr:
The "end goal", or "best handler level" would be something like `CommandHandler.Create(Verbosity verbosity => { ... })`, but that requires 2 options (`--quiet` and `--verbose`, one adding 1, and one subtracting 1, disallowed to exist together) being merged to one value which I have no idea if is supported or not. Still, I think it make sense to work on this feature bottom-up, and get something working, and then build the more higher-level concepts on top of it.

Alxandr:
Given the lack of progress, I'm choosing to close this PR.

CEbbinghaus:
Its sad to see that this functionality is still not implemented. I have found the model binding in general to be unreliable and not particularity usable. As such I am really hoping to be able too add a flag option with a int count that returns me how many times that flag was set. In the end I don't think we should have -v and -q as one of the same but rather as two seperate counts and the resulting verbosity becomes `result = verboseAmount + (quietAmount * -1)`.

jonsequitur:
I've just implemented this in #2164. Please take a look.

Issue #560, opened 2019-06-15
add support for required options


jonsequitur:
Add support for making an option, or one option out of a group of options, required. While this can currently be done using a custom validator, it's awkward and not obvious.

Related: #381.

tmds:
UX example using dotnet-counters from https://github.com/dotnet/diagnostics:

```
$ dotnet counters --help
Usage:
  dotnet-counters [options] [command]

Options:
  --version    Display version information

Commands:
  monitor <counter_list>    Start monitoring a .NET application
  list                      Display a list of counter names and descriptions, grouped by provider.
  list-processes            Display a list of dotnet processes that can be monitored.

$ dotnet counters monitor
ProcessId is required.
```

code:
```
        private static Command MonitorCommand() =>
            new Command(
                "monitor", 
                "Start monitoring a .NET application", 
                new Option[] { ProcessIdOption(), RefreshIntervalOption() },
                argument: CounterList(),
                handler: CommandHandler.Create<CancellationToken, List<string>, IConsole, int, int>(new CounterMonitor().Monitor));

        private static Option ProcessIdOption() =>
            new Option(
                new[] { "-p", "--process-id" }, 
                "The ID of the process that will be monitored.",
                new Argument<int> { Name = "pid" });
```

darinclarkdev:
Until the api is decided this is what works for me.
```
cmdBuilder = new CommandLineBuilder(rootCommand)
				.UseRequiredOptions()
				.UseDefaults();

public class RequiredOption : Option {
	public RequiredOption(string alias, string description) : base(alias, description) { }
	public RequiredOption(string[] aliases, string description) : base(aliases, description) { }
}

public static class CommandLineBuilderExtensions {
    public static CommandLineBuilder UseRequiredOptions(this CommandLineBuilder value) {
        value.UseMiddleware(async (context, next) => {
	    bool errorfound = false;
            var options = context.ParseResult.CommandResult;
            foreach (Option item in options.Command.Children) {
                if (item is RequiredOption && !options.Children.Any(s => s.Name == item.Name)) {
                    context.Console.Out.WriteLine($"{item.Name} is a [REQUIRED] option\n");
                    errorfound = true;
                }
            }
            if (!errorfound)
                await next(context);
            else {
                var help = new HelpResult();
		help.Apply(context);
            }
        });
	return value;
    }
}
```

guy-lud:
Doesn't `ArgumentArity` helps you ?

[...]

PR #673, opened 2019-11-02
Start work on required options

Alxandr:
This PR attempts to add support for required options, by making options with arguments that have a minimum arity of 1 to be required. This is based on discussions in #560. Note that there are currently failing tests (as mentioned in https://github.com/dotnet/command-line-api/issues/560#issuecomment-540692266). This PR is published as a draft to show the work as it's ongoing and allow for feedback. It might also be better to continue the discussion of the broken tests here, rather than in #560, given that they are more implementation-specific than required options in general.

### DIFF
--- a/src/System.CommandLine.Tests/Invocation/CommandHandlerTests.cs
+++ b/src/System.CommandLine.Tests/Invocation/CommandHandlerTests.cs
@@ -327,7 +327,7 @@ namespace System.CommandLine.Tests.Invocation
             };
             command.Handler = CommandHandler.Create<IConsole>(console => { console.Out.Write("Hello!"); });
 
-            await command.InvokeAsync("command", _console);
+            await command.InvokeAsync("command -x 123", _console);
 
             _console.Out.ToString().Should().Be("Hello!");
         }
--- a/src/System.CommandLine.Tests/ParsingValidationTests.cs
+++ b/src/System.CommandLine.Tests/ParsingValidationTests.cs
@@ -83,6 +83,24 @@ namespace System.CommandLine.Tests
         }
 
         [Fact]
+        public void When_a_required_option_is_not_supplied_then_an_error_is_returned()
+        {
+            var parser = new Parser(new Option("-x")
+            {
+                Argument = new Argument
+                {
+                    Arity = ArgumentArity.ExactlyOne
+                }
+            });
+
+            var result = parser.Parse("");
+
+            result.Errors
+                  .Should()
+                  .Contain(e => e.Message == "Required argument missing for option: -x");
+        }
+
+        [Fact]
         public void When_no_option_accepts_arguments_but_one_is_supplied_then_an_error_is_returned()
         {
             var parser = new Parser(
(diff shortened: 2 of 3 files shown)


Alxandr:
There are currently 2 failing tests. 

The first one is a argument of type `int?`, which has an arity of 1 (both minimum *and* maximum), thus failing the test (because an option with an argument with min arity > 0 is not provided). The test can be seen bellow:

```csharp
        [Fact]
        public async Task Nullable_parameters_are_bound_to_null_when_option_is_not_specified()
        {
            var wasCalled = false;
            int? boundAge = default;

            var command = new Command("command")
            {
                new Option("--age")
                {
                    Argument = new Argument<int?>()
                }
            };
            command.Handler = CommandHandler.Create<int?>(age =>
            {
                wasCalled = true;
                boundAge = age;
            });

            await command.InvokeAsync("command", _console);

            wasCalled.Should().BeTrue();
            boundAge.Should().BeNull();
        }
```

In my opinion, this seems like a bug in `Argument` (probably with compensating logic elsewhere), where an `Argument` of type `int?` should have a minimum arity of 0, not 1.

The second one is a test with the same option in multiple places of a command hierarchy:

```csharp
            [Theory]
            [InlineData("outer --the-option xyz inner")]
            [InlineData("outer inner --the-option xyz")]
            public void When_an_option_is_shared_between_an_outer_and_inner_command_then_specifying_in_one_does_not_result_in_error_on_other(string commandLine)
            {
                var option = new Option("--the-option")
                {
                    Argument = new Argument<string>()
                };

                var command = new Command("outer")
                {
                    new Command("inner")
                    {
                        option
                    },
                    option
                };

                var parseResult = command.Parse(commandLine);

                parseResult.Errors.Should().BeEmpty();
            }
```

This test has lead me to question how some of this is supposed to work. For instance, if I take the command in the test above, and use the following parse the `outer --the-option xyz inner --the-option abc`, what's the expected outcome? How does this relate to required options? Which command handlers are expected to run, and what will the value of `--the-option` be there?

jonsequitur:
This isn't the approach I would recommend for making an option required, as it will significantly change the meaning of argument arity. A minimum argument arity of 1 indicates that when the option is specified, it must have at least one value. This is a syntax concept and is orthogonal to the bound type (though inferred from it by convention most of the time).

I'm about to contradict what I wrote [here](https://github.com/dotnet/command-line-api/issues/560#issuecomment-544670318). Apologies for the mistake. I was not thinking it through, which is more evidence that this isn't intuitive. 

Let's take the example of `int?`:

```csharp
var option = new Option("-x") { Argument = new Argument<int?>() };

option.Argument.Arity.MinimumNumberOfValues.Should().Be(1);
option.Argument.Arity.MaximumNumberOfValues.Should().Be(1);
```

The reason `Argument.Arity.MinimumNumberOfValues` is not `0` is that the following command line is nonsensical:

```console
> myapp -x
```

The difference between `Argument<int>` and `Argument<int?>` is not the arity, but the bound type.

So if the following is submitted, it's valid in either case:

```console
> myapp 
```

The handler parameter will be bound to null in the case of `Argument<int?>` and will be bound to `0` in the case of `Argument<int>`.

So I think that making an option required needs to be orthogonal to the arity of the option's argument.

PR #678, opened 2019-11-11
Enforce ModelBinder explicit member binding

couven92:
By default the ModelBinder will bind to all public instance properties of its configured type. There is no way to instruct the binder to only bind against the configured bindings (using the `BindMemberFromValue` methods).

This PR adds the `EnforceExplicitBinding` method that prevents model binding against all unconfigured properties.

This is done by adding a `null` value source to the internal `NamedValueSources` dictionary of the `ModelBinder` for each property that has no specified binding. When the `ModelBinder` gets a value source that is `null` (which otherwise cannot happen), binding to the member is skipped.

I updated the logic in the `BindMemberFromValue` methods to allow overriding a `null` value source in `NamedValueSources`, allowing the call to `EnforceExplicitBinding` to be independent from previous or succeeding calls to `BindMemberFromValue`.

### DIFF
--- a/src/System.CommandLine/Binding/ModelBinder.cs
+++ b/src/System.CommandLine/Binding/ModelBinder.cs
@@ -208,10 +208,13 @@ namespace System.CommandLine.Binding
                     if (includeMissingValues)
                     {
                         if (valueDescriptor is ParameterDescriptor parameterDescriptor &&
-                            parameterDescriptor.Parent is ConstructorDescriptor constructorDescriptor &&
-                            ShouldPassNullToConstructor(constructorDescriptor.Parent, constructorDescriptor))
+                            parameterDescriptor.Parent is ConstructorDescriptor constructorDescriptor)
                         {
-                            boundValue = BoundValue.DefaultForType(valueDescriptor);
+                            if (parameterDescriptor.HasDefaultValue)
+                                boundValue = BoundValue.DefaultForValueDescriptor(parameterDescriptor);
+                            else if (parameterDescriptor.AllowsNull && 
+                                ShouldPassNullToConstructor(constructorDescriptor.Parent, constructorDescriptor))
+                                boundValue = BoundValue.DefaultForType(valueDescriptor);
                         }
                     }
                 }
--- a/src/System.CommandLine/Binding/ModelBinder.cs
+++ b/src/System.CommandLine/Binding/ModelBinder.cs
@@ -33,9 +33,29 @@ namespace System.CommandLine.Binding
 
         public void BindMemberFromValue(PropertyInfo property, IValueDescriptor valueDescriptor)
         {
-            NamedValueSources.Add(
-                (property.PropertyType, property.Name),
-                new SpecificSymbolValueSource(valueDescriptor));
+            var key = (property.PropertyType, property.Name);
+            if (NamedValueSources.TryGetValue(key, out var existingValueSource) &&
+                existingValueSource is null)
+            {
+                // Override existing null value source
+                NamedValueSources[key] = new SpecificSymbolValueSource(valueDescriptor);
+            }
+            else
+            {
+                NamedValueSources.Add(
+                    key,
+                    new SpecificSymbolValueSource(valueDescriptor)); 
+            }
+        }
+
+        public void EnforceExplicitBinding()
+        {
+            foreach (var property in _modelDescriptor.PropertyDescriptors)
+            {
+                var key = (property.Type, property.ValueName);
+                if (!NamedValueSources.ContainsKey(key))
+                    NamedValueSources[key] = null;
+            }
         }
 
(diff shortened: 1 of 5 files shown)


jonsequitur:
@couven92 Please rebase on the latest master. #680 fixes the Ubuntu CI error.

Thanks, @wli3!

src/System.CommandLine/Binding/ModelBinder.cs
```diff
@@ -33,9 +33,29 @@ internal ModelBinder(IValueDescriptor valueDescriptor)
...
+            var key = (property.PropertyType, property.Name);
+            if (NamedValueSources.TryGetValue(key, out var existingValueSource) &&
+                existingValueSource is null)
+            {
+                // Override existing null value source
+                NamedValueSources[key] = new SpecificSymbolValueSource(valueDescriptor);
+            }
+            else
+            {
+                NamedValueSources.Add(
```
jonsequitur: It was probably a bug that this called `Add` rather than overwriting, which this change makes clear. We should avoid having the behavior differ based on whether there was a strict binding specified, so let's just always overwrite.

src/System.CommandLine/Binding/ModelBinder.cs
```diff
@@ -33,9 +33,29 @@ internal ModelBinder(IValueDescriptor valueDescriptor)
 
         public void BindMemberFromValue(PropertyInfo property, IValueDescriptor valueDescriptor)
         {
-            NamedValueSources.Add(
-                (property.PropertyType, property.Name),
-                new SpecificSymbolValueSource(valueDescriptor));
+            var key = (property.PropertyType, property.Name);
+            if (NamedValueSources.TryGetValue(key, out var existingValueSource) &&
+                existingValueSource is null)
```
jonsequitur: Rather than assign a semantic meaning to null and then always have to check for null, a type representing a strict binding might be preferable. Although another approach that doesn't require populating the dictionary might make things easier to understand.

src/System.CommandLine/Binding/ModelBinder.cs
```diff
@@ -33,9 +33,29 @@ internal ModelBinder(IValueDescriptor valueDescriptor)
...
+            {
+                NamedValueSources.Add(
+                    key,
+                    new SpecificSymbolValueSource(valueDescriptor)); 
+            }
+        }
+
+        public void EnforceExplicitBinding()
+        {
+            foreach (var property in _modelDescriptor.PropertyDescriptors)
```
jonsequitur: This would also need to consider `ModelDescriptor.ConstructorDescriptors`.

jonsequitur:
Let's discuss an approach that makes the policy explicit, rather than basing it on assigning meaning to `null` and having to keep the dictionary populated with markers. 

Also, please add some tests.

couven92:
@jonsequitur While adressing your review I actually realized that I had misunderstood how exactly the parts in the `ModelBinder` fit together. The value sources added to `NamedValueSources` were actually explicit bindings, and I had simply placed the return `null` at the wrong place.

The actual culprit is `CurrentSymbolResultValueSource`: It is a source that will attempt to bind a value from the parse result if it finds an option or argument that matches the target descriptor name.

So my updated code will still cause `GetValueSource` to return `null` if `EnforceExplicitBinding` (now a boolean property) is `true`. In `GetValues` a `null` value source is interpreted as: "If there is no value source, I cannot bind the value" which seems reasonable to me.

So now for the additions:  
I splitted `NamedValueSources` into two dictionaries: one for member bindings and one for constructor argument bindings. That also requires a new `BindConstructorArgumentFromValue` method to allow specifying how to bind against a ParameterInfo.

The `Bind<sth>FromValue` methods now all confirm whether the specified `ParameterInfo` or `PropertyInfo` make sense, and instead of storing the type and name in the dictionary we now store the actual descriptor that `CreateInstance` and `UpdateInstance` will use (i.e. a descriptor coming from `_modeldescriptor`).

couven92:
I also changed `CreateInstance` to try out all Constructors instead of just the first found by reflection.

I try to bind the longest constructor (the one with most arguments) first and attempt binding first with all parameters before filling out default parameters values.

couven92:
@jonsequitur I rebased off master, but the tests still fail?

And yes, I'll or course write up some new test for the stuff I added/changed here 😃

jonsequitur:
This sounds like a nice improvement to the design.

It also sounds like perhaps `CurrentSymbolResultValueSource` could more accurately be called something like `CurrentSymbolResultMatchingByNameValueSource`, if there's a more concise way to express that.

couven92:
How about `ParseResultMatchingValueSource`?

________________________________

couven92:
@jonsequitur tests added, but CI is trolling me again... No we don't get any logs at all... 🤔

jonsequitur:
Thanks, @couven92!

PR #680, opened 2019-11-11
Change to forward slash

wli3:


wli3:
filed https://github.com/NuGet/Home/issues/8806

Issue #674, opened 2019-11-02
Add support for Option<T>


jonsequitur:
The current API for specifying an argument is a bit more verbose than needed:

```csharp
new Option("-x")
{ 
    Argument = new Argument<int>()
}
```

The following could be a way to simplify it:

```csharp
new Option<int>("-x")
```

jonsequitur:
#687 adds simple support for `Option<T>` but there's probably more to be done here, since a number of methods for customizing suggestions and validation operate on `Argument`. Should equivalent extension methods be added for `Option<T>`, so you can do things like this:

```csharp
var option = new Option<T>("-x").WithSuggestions("one", "two", "three");
```

PR #687, opened 2019-11-23
Option<T> and Option.Required

jonsequitur:
This adds support for `Option<T>` (#674) and `Option.Required` (#560).

### DIFF
--- a/src/System.CommandLine/Parsing/ParseResultVisitor.cs
+++ b/src/System.CommandLine/Parsing/ParseResultVisitor.cs
@@ -153,88 +153,78 @@ namespace System.CommandLine.Parsing
         {
             ValidateCommandHandler(_innermostCommandResult);
 
-            foreach (var commandResult in _innermostCommandResult.RecurseWhileNotNull(c => c.ParentCommandResult))
+            foreach (var symbol in _innermostCommandResult.Command.Children)
             {
-                foreach (var symbol in commandResult.Command.Children)
+                PopulateDefaultValues(_innermostCommandResult, symbol);
+            }
+
+            ValidateCommandResult(_innermostCommandResult);
+
+            foreach (var result in _innermostCommandResult.Children)
+            {
+                switch (result)
                 {
-                    PopulateDefaultValues(commandResult, symbol);
-                }
+                    case ArgumentResult argumentResult:
 
-                ValidateCommand(commandResult);
+                        ValidateArgumentResult(argumentResult);
 
-                foreach (var result in commandResult.Children)
-                {
-                    switch (result)
-                    {
-                        case ArgumentResult argumentResult:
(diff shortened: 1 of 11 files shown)


src/System.CommandLine/Parsing/ParseResultVisitor.cs
```diff
@@ -153,88 +153,78 @@ protected override void Stop(SyntaxNode node)
...
-            foreach (var a in commandResult
-                              .Command
-                              .Arguments)
+            foreach (var option in commandResult
+                                   .Command
+                                   .Options)
             {
-                if (a is Argument argument)
+                if (option is Option o &&
+                    o.Required &&
```
Keboo: Should `Required` be added to `IOption` to avoid the need for the `Option` cast here? Thouhg looking at the implementation, it appears that the only children that will be returned in the `Command.Options` will be `Options`.

src/System.CommandLine/Parsing/ParseResultVisitor.cs
```diff
@@ -153,88 +153,78 @@ protected override void Stop(SyntaxNode node)
...
+                        new ParseError($"Option '{o.RawAliases.First()}' is required.",
+                                       commandResult));
+                }
+            }
+
+            foreach (var symbol in commandResult
+                                   .Command
+                                   .Arguments)
+            {
+                if (symbol is Argument argument)
```
Keboo: This `if` check does not appear to add much. It prevents a `null` argument from getting through. But it is not possible to get a `null` back from `Command.Arguments` due to the implementation using `.OfType`. I would recommend just changing the foreach to be:
```C#
            foreach (IArgument argument in commandResult
                                   .Command
                                   .Arguments)
            {
```

src/System.CommandLine/Parsing/ParseResultVisitor.cs
```diff
@@ -253,25 +243,54 @@ private void ValidateCommandHandler(CommandResult commandResult)
...
+                    var message = validate(optionResult);
+
+                    if (!string.IsNullOrWhiteSpace(message))
+                    {
+                        _errors.Add(new ParseError(message, optionResult));
+                    }
+                }
+            }
+
+            foreach (var a in optionResult.Children
```
Keboo:
```C#
foreach (var argumentResult in optionResult.Children
```

src/System.CommandLine/Parsing/ParseResultVisitor.cs
jonsequitur: This is still a little awkward but maybe, yes. `IOption` is, in practice, a sort of read-only `Option`, but we didn't go to the additional step of making it a truly immutable copy of the original mutable `Option`.

Issue #686, opened 2019-11-22
[rendering] Scrolling/Growing Region


josalem:
It would be helpful to have an API for printing to a region inside of a collection of Views as if you were just printing to the console.  The result of writing a line to this region would be similar to that of calling `Console.Out.WriteLine` in a typical application.

An example of the rendering behavior I would expect from such a construct would be similar to that of `apt` on Ubuntu.  When invoked to install a package, the console starts printing output with a status bar at the bottom while the log grows upwards, e.g.,
```
oldest log message         ^
older log message          |
newer log message          |- direction of logs
newest log message
[||||||||||||                        25% ] <-- stationary
```

I would expect the API for using this region to be a drop in replacement for printing to the console using the `Console.{Out|Error}.WriteLine` APIs.  If the view isn't inside an explicitly sized region, it should continue to grow in the appropriate direction, pushing the the bottom of the console buffer out to accommodate.  If the view _is_ inside a region with a set height/width, I would expect the region to stay the same size and the text to "fall off the edge" of the region in the appropriate direction.  To achieve something like the `apt` example above, I could imagine an API like:

```csharp
var stackLayout = new StackLayout(/* configured without an explicit region */);
var scrollingView = new ScrollingView(new Region(x1,y1,x2,y2), ScrollDirection.Up);
var statusBar = StatusBarView.FromObservable<float>(progressObservable); // doesn't exist but would also be convenient

stackLayout.Add(scrollingView);
stackLayout.Add(statusBar);

//...

foreach (var dependency in packageToInstall.Dependencies)
{
    // ...
    scrollingView.WriteLine($"Installed {dependency.Name} to {installDir}");
}
```

I know the APIs above don't 100% map to the current APIs, but hopefully my example explains the desired API and expected results.  Please let me know if there is anything I can clarify.  I could imagine this API being useful inside `TableView`s for putting log output to actions taken in another pane of the table.

Currently, to achieve a semblance of this behavior, I have been creating `ContentView`s from `Observable<string>`s and then updating the `string` with new lines of text.

adamsitnik:
Hello!

We have decided to deprecate the System.CommandLine.Rendering package: https://github.com/dotnet/command-line-api/issues/2576#Deprecating-experimental-projects

We recommend using https://www.nuget.org/packages/Spectre.Console which became the golden standard in the .NET Ecosystem

PR #698, opened 2019-12-03
Fix NRE in ScreenView.Render when there is no terminal

qmfrederik:
When using redirected output (e.g. `myapp > out` or `myapp | less`), `_console.GetTerminal()` would return `null` and `_console.GetTerminal().HideCursor()` results in a NRE, with a stack trace like this:

```
Unhandled exception: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at System.CommandLine.Rendering.Views.ScreenView.Render(Region region)
   at scrach.Program.List(InvocationContext context) in /home/fcarlier/scrach/Program.cs:line 61
   --- End of inner exception stack trace ---
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor, Boolean wrapExceptions)
   at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at System.Delegate.DynamicInvokeImpl(Object[] args)
   at System.Delegate.DynamicInvoke(Object[] args)
   at System.CommandLine.Invocation.ModelBindingCommandHandler.InvokeAsync(InvocationContext context)
   at System.CommandLine.Invocation.InvocationPipeline.<>c__DisplayClass4_0.<<BuildInvocationChain>b__0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.CommandLine.Invocation.InvocationExtensions.<>c.<<UseParseErrorReporting>b__22_0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.CommandLine.Invocation.InvocationExtensions.<>c__DisplayClass8_0.<<UseTypoCorrections>b__0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.CommandLine.Invocation.InvocationExtensions.<>c.<<UseSuggestDirective>b__7_0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.CommandLine.Invocation.InvocationExtensions.<>c.<<UseParseDirective>b__6_0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.CommandLine.Invocation.InvocationExtensions.<>c.<<UseHelp>b__20_0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.CommandLine.Builder.CommandLineBuilderExtensions.<>c__DisplayClass4_0.<<UseVersionOption>b__0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.CommandLine.Invocation.InvocationExtensions.<>c.<<RegisterWithDotnetSuggest>b__23_0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.CommandLine.Invocation.InvocationExtensions.<>c__DisplayClass5_0.<<UseExceptionHandler>b__0>d.MoveNext()
```

### DIFF
--- a/src/System.CommandLine.Rendering/Views/ScreenView.cs
+++ b/src/System.CommandLine.Rendering/Views/ScreenView.cs
@@ -68,7 +68,7 @@ namespace System.CommandLine.Rendering.Views
         // may not want this?
         public void Render(Region region)
         {
-            _console.GetTerminal().HideCursor();
+            _console.GetTerminal()?.HideCursor();
 
             Child?.Render(Renderer, region);
         }
@@ -82,7 +82,7 @@ namespace System.CommandLine.Rendering.Views
 
         public void Dispose()
         {
-            _console.GetTerminal().ShowCursor();
+            _console.GetTerminal()?.ShowCursor();
 
             if (_child is View child)
             {


jonsequitur:
The intention is that the `ScreenView` should not be usable when there is no terminal. `NullReferenceException` isn't the right exception here but `InvalidOperationException` might be, rather than silently doing nothing.

qmfrederik:
@jonsequitur In my case, I'm trying to render a simple view to stdout (basically the result of an operation).

I'm using a `TableView` and calling `.Render(region)`, with a custom region.
With this PR applied, I'm able to do that even if stdout is redirected (without the PR I get the NRE).

The NRE comes from the `ScreenView` trying to toggle the cursor; the normal rendering logic works just fine.

Is there a better way to do what I'm trying to achieve?

jonsequitur:
The rendering APIs are very preliminary and there are a few usability issues we're hoping to work through. The `ScreenView` for example is intended to be able to update what's displayed in a terminal (whether via `System.Console` on older Windows consoles or via ANSI sequences on xterm or Windows Terminal). For redirected output this should typically be an invalid operation, since there's no terminal. 

What you're trying to do looks like it should not need a `ScreenView` at all. But there's a missing concept that needs to be added (#686).

That's the long explanation for why I think calling `Render` in your example would be a programmer error that should be surfaced as an exception. It might be cleaner and clearer to create a new extension method for this directly off of `IConsole`.

qmfrederik:
@jonsequitur I find it odd for a program which only renders text to the console to crash when no terminal is present. I may not have enough context for what plans you have with `ScreenView` & friends, though.

If you do intend `ScreenView` to have a strong dependency on `Terminal`, perhaps consider making that dependency explicit. Currently the `ScreenView` constructor takes a `ConsoleRenderer` and a `IConsole` in its constructor. That seems to be a bit misleading, then, as you clearly expect a `Terminal`.

A workaround which seems to work for me, is to create a `VirtualTerminal`, passing the current console and setting `virtualTerminalMode` to `null`.

qmfrederik:
@jonsequitur Based on your feedback, silently ignoring the absence of a terminal is not the intended behavior, so I'm going to go ahead and close this PR for now.

PR #717, opened 2020-01-05
move some types into more specific namespaces

jonsequitur:


src/System.CommandLine.Benchmarks/CommandLine/Perf_Parser_DefaultCtor.cs
```diff
@@ -1,12 +1,13 @@
...
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.CommandLine.Parsing;
 using BenchmarkDotNet.Attributes;
 
 namespace System.CommandLine.Benchmarks.CommandLine
 {
     /// <summary>
-    /// Measures the performance of <see cref="System.CommandLine.Parser"/> default constructor.
+    /// Measures the performance of <see cref="Parsing.Parser"/> default constructor.
```
Keboo: Shouldn't this just be
```suggestion
    /// Measures the performance of <see cref="Parser"/> default constructor.
```

PR #718, opened 2020-01-05
Xml comments for RootCommand

KathleenDollard:


src/System.CommandLine/RootCommand.cs
```diff
@@ -7,12 +7,22 @@
 
 namespace System.CommandLine
 {
+    /// <summary>
+    /// A command representing entry point of an application and used to Invoke the correct handlers.
```
Keboo: Not sure the Invoke potion of this is really accurate. This class is not actually responsible for doing the invocation, rather holding the data for parsing/invocation.

```suggestion
    /// A command representing entry point of an application.
```

jonsequitur:
```suggestion
    /// A command representing an application entry point.
```

src/System.CommandLine/RootCommand.cs
```diff
@@ -7,12 +7,22 @@
...
+    /// </summary>
     public class RootCommand : Command
     {
+        /// <summary>
+        /// Create a new instance of RootCommand
+        /// </summary>
+        /// <param name="description">The description of the command that could be used in a help display.</param>
```
jonsequitur: Simpler?

```suggestion
        /// <param name="description">The description of the command shown in help.</param>
```

src/System.CommandLine/RootCommand.cs
```diff
@@ -42,8 +52,14 @@ public override string Name
             Assembly.GetEntryAssembly() ??
             Assembly.GetExecutingAssembly();
 
+        /// <summary>
+        /// The name of the executable currently running, or if missing the first string argument.
```
jonsequitur: > or if missing the first string argument.

I'm not sure I understand this bit.

KathleenDollard: Looking at the code, that's what I thought it was doing.

KathleenDollard:
Abandoning in favor of cleaner PR

PR #724, opened 2020-01-07
remove obsolete Command.Argument and Command.GetValueOrDefault

jonsequitur:


src/System.CommandLine/Argument{T}.cs
```diff
@@ -18,9 +18,9 @@ public Argument() : base(null)
             ArgumentType = typeof(T);
         }
 
-        public Argument(string name, T defaultValue) : this(name)
+        public Argument(string name, Func<T> defaultValue) : this(name)
         {
-            SetDefaultValue(defaultValue);
+            SetDefaultValue(() => defaultValue());
```
Keboo: Is there a reason we are wrapping the Func in another Func?
```suggestion
            SetDefaultValue(defaultValue);
```

jonsequitur: `SetDefaultvalue` has an overload that takes `object`. 

A cast could also work here.

Keboo: ah I missed that. I would suggest the cast here (and in the few other usages in this file) to avoid extra allocations.

Issue #723, opened 2020-01-07
Support for taking control of an argument's behavior totally


rynowak:
## Summary

I'm trying to do something that I'm not sure if there's already a recipe for. 

I'm trying to make an option that works like `dotnet run -p <path to project>`. Basically:

- Finds a project file in the current directory if option is unspecified
- Find a project file in the specified directory if the argument specified is a directory
- Uses the argument specified if the argument is a file
- Validation that we found a single project file and it exists

I'm using this option in multiple places, so it's pretty important to me that these errors are reported by the parameter binding system (not by code that I write **after** getting the option values bound).

## What I Tried

My code is here: https://github.com/rynowak/opulence/blob/master/src/dotnet-opulence/StandardOptions.cs#L50

What I tried is creating an argument bound to `FileInfo` - and then taking control of the conversion process. This doesn't meet all my requirements because it doesn't default to the current directory - conversion delegates don't run unless a value is passed in at the command line.

So I added a default value, however you don't have the chance to report validation errors when setting the default value. The validation error can occur when defaulting if there's 0, or more than one project file in the current directory.

So I added a validator. 

At this point it seems to cover all of my bases, but it's pretty complex and non-obvious. My default value has to no-op when it hits a validation error so that my validator can report it, etc.

## What I'd like

It be nice if there were a way to specify a `Func<(something), FileInfo>` for my option that can:

- Run regardless of whether a the option is specified
- Be responsible for converting the unparsed value to a `FileInfo`
- Add my own validation errors 

All of these capabilities already exist but they're part of three different extensibility points.

jonsequitur:
Thanks for taking the time to write this up! You're spot on.

I think it makes a lot of sense to consolidate these into a single API. I've wanted to replace the custom validation API because the intent isn't clear. (Returning a `string` to indicate a validation failure while returning `null` to indicate success was a pretty poor design decision, which I'm ok with saying because I made it.) And this has been one of the pieces I've most wanted to fix before leaving preview.

I'll make an attempt at a better design and post an update soon.

rynowak:
I'm happy to give feedback as well when you're ready for it. I see that you've already started looking at this 👍

jonsequitur:
@rynowak Take a look at #736 and let me know what you think. Most of the commits are refactoring but the last commit adds a number of tests that outline the ways the new API can be used: https://github.com/dotnet/command-line-api/pull/736/files#diff-2d70ca1ca71633e0fdb081a2f67da024R57

[...]

Issue #704, opened 2019-12-09
TryConvertArgument and defaultValue


dashesy:
Can we change 

```C#
        public Argument(TryConvertArgument<T> convert);
```

To 

```C#
        public Argument(TryConvertArgument<T> convert, string defaultValue);
```

And pass `defaultValue` to the converter if nothing provided? 
To be similar to 

```C#
        public Argument(string name, T defaultValue);
```

That has a single-line way of specifying default values?

jonsequitur:
Related: #581

dashesy:
Yes, same issue. I follow the other thread.

Issue #581, opened 2019-07-02
Argument with custom converter and defaultValue?


mmierzwa11:
Is it possible to define custom converter for argument and defaultValue?

Keboo:
@mmierzwa11 at present the Argument class has constructors for specifying delegates for either [default value](https://github.com/dotnet/command-line-api/blob/b207fa37ae7255ecdbdcef31226dc7cc51e31599/src/System.CommandLine/Argument%7BT%7D.cs#L23), or [custom converter](https://github.com/dotnet/command-line-api/blob/b207fa37ae7255ecdbdcef31226dc7cc51e31599/src/System.CommandLine/Argument%7BT%7D.cs#L28). If you wanted both you would need to handle the "default" as part of your TryConvertArgument delegate.

mmierzwa11:
I thought so at first, but from what I see CustomConverter is only invoked if option is specified, and default value delegate is always invoked - I want both behaviors for my option.

jonsequitur:
@mmierzwa11 Are you wanting to feed the output of the `defaultValue` delegate into a custom converter? Can you describe a little more generally what you're looking to do?

mmierzwa11:
I don't quite remember what was the purpose exactly, but I think I wanted custom converter for byte array (written in hex notation), with default value if not specified.

qmfrederik:
I have two use cases which I think require me to

qmfrederik:
I have two use cases which I think would benefit from having a custom converter which is able to specify a default value:

- One case is an option where user specifies a path to a license file. The type for this option is `License`. Ideally:
  - If a user specifies a path, we resolve the path and create a `License` object for it
  - If a user does not specify a path, we look in a couple of "well known" locations (e.g. working dir, app data,...) for a license file
- A second option where the user specifies the ID of a device connected to her/his computer. The type for this option is `Device`. Ideally:
  - If the user does not specify a value, check whether one (and only one) device is connected and return that value, or error out
  - If the user does specify a value, convert that value (or error out)

As @mmierzwa11 indicates, if you specify a custom converter, that converter is not invoked when the option is not specified, so I don't think these scenarios are currently possible.

dashesy:
I have inline `Option`  that pass to `Command`. 
Currently I have

```C#
                new Option(
                    new[] { "--output", "-o" },
                    "File path to output the results. Pass null to suppress outputs. (Default=stdout)")
                {
[...]

PR #730, opened 2020-01-08
fix #728

### LINKED ISSUE #728: Argument<T> constructor shouldn't accept a null default value func
This was a simple gotcha that took a bit too long to sort out:

```
var option = new Option("--log", "file to write a log to") { Argument = new Argument<string?>(defaultValue: null).LegalFilePathsOnly() }
```

defaultValue is a parameter not of type `string?` but `Func<string?>`, so passing null meant the default argument didn't actually exist, but the library would internally null ref when trying to use that func. Compare to this:

```
var option = new Option("--log", "file to write a log to") { Argument = new Argument<string?>(defaultValue: () => null).LegalFilePathsOnly() }
```

which is correct!

It seems a null check should be added to here:

https://github.com/dotnet/command-line-api/blob/5c16ca9f61fe45275614fa52f561c833d0943178/src/System.CommandLine/Argument%7BT%7D.cs#L23

to prevent defaultValue from being null. Also, if this constructor was null annotated with C# 8.0 nullable annotations this could have been caught statically.

jonsequitur:
This addresses #728, in part, by preventing `null` from being set as the default argument via the `Argument.SetDefaultValue(object)` overload. Additional design questions [here](https://github.com/dotnet/command-line-api/issues/728#issuecomment-571883738).

### DIFF
--- a/src/System.CommandLine/Argument.cs
+++ b/src/System.CommandLine/Argument.cs
@@ -11,7 +11,7 @@ namespace System.CommandLine
 {
     public class Argument : Symbol, IArgument
     {
-        private Func<object> _defaultValue;
+        private Func<object> _getDefaultValue;
         private readonly List<string> _suggestions = new List<string>();
         private readonly List<ISuggestionSource> _suggestionSources = new List<ISuggestionSource>();
         private IArgumentArity _arity;
@@ -114,24 +114,27 @@ namespace System.CommandLine
 
         public void AddValidator(ValidateSymbol<ArgumentResult> validator) => Validators.Add(validator);
 
-        public object GetDefaultValue() => _defaultValue?.Invoke();
+        public object GetDefaultValue()
+        {
+            if (_getDefaultValue is null)
+            {
+                throw new InvalidOperationException($"Argument \"{Name}\" does not have a default value");
+            }
+
+            return _getDefaultValue.Invoke();
+        }
 
         public void SetDefaultValue(object value)
         {
-            if (value is null)
-            {
-                throw new ArgumentNullException(nameof(value));
-            }
-
-            SetDefaultValue(() => value);
+            SetDefaultValueFactory(() => value);
         }
 
-        public void SetDefaultValue(Func<object> value)
+        public void SetDefaultValueFactory(Func<object> getValue)
         {
-            _defaultValue = value ?? throw new ArgumentNullException(nameof(value));
+            _getDefaultValue = getValue ?? throw new ArgumentNullException(nameof(getValue));
         }
 
-        public bool HasDefaultValue => _defaultValue != null;
+        public bool HasDefaultValue => _getDefaultValue != null;
 
         internal static Argument None => new Argument { Arity = ArgumentArity.Zero };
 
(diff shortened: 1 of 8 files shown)


src/System.CommandLine/Argument.cs
```diff
@@ -116,9 +116,20 @@ bool DefaultConvert(SymbolResult symbol, out object value)
 
         public object GetDefaultValue() => _defaultValue?.Invoke();
 
-        public void SetDefaultValue(object value) => SetDefaultValue(() => value);
+        public void SetDefaultValue(object value)
+        {
+            if (value == null)
```
Keboo: Should we be adopting C# 8-isms with `is null`?
```suggestion
            if (value is null)
```

src/System.CommandLine/Argument.cs
```diff
@@ -116,9 +116,20 @@ bool DefaultConvert(SymbolResult symbol, out object value)
 
         public object GetDefaultValue() => _defaultValue?.Invoke();
 
-        public void SetDefaultValue(object value) => SetDefaultValue(() => value);
+        public void SetDefaultValue(object value)
+        {
+            if (value is null)
```
jasonmalinowski: So I don't get this block here, since this implies calling Argument.SetDefaultValue((object)null) would throw, but Argument.SetDefaultValue(() => null) _would_ be fine, even though I'd expect those to meant the same thing.

jonsequitur: Agreed. I think the API needs to change so that intention is more clear between:

* there is a default value, and it is null, and
* there is no default value.

The latter is not currently representable in the API.

jonsequitur: In the latest commit, I've differentiated `SetDefaultValue` from `SetDefaultValueFactory` and added tests to clarify the behaviors. Do you think this helps with the ambiguity?

These APIs still feel fairly provisional though, and a few issues (#581, #704, #723) have pointed toward the need for a more holistic approach to default values, custom conversion, and validation. I won't try to address this here but I suspect it might result in  `SetDefaultValue` and `SetDefaultValueFactory` becoming obsolete.

src/System.CommandLine/Argument{T}.cs
```diff
@@ -18,17 +18,17 @@ public Argument() : base(null)
...
+        public Argument(string name, Func<T> getDefaultValue) : this(name)
         {
-            SetDefaultValue(() => defaultValue());
+            SetDefaultValueFactory(() => getDefaultValue());
         }
 
         public Argument(Func<T> defaultValue) : this()
         {
-            SetDefaultValue(() => defaultValue());
+            SetDefaultValueFactory(() => defaultValue());
```
jasonmalinowski: @jonsequitur: it seems the original bug I hit isn't being fixed here: if I pass null in for defaultValue, this will still call SetDefaultValueFactory and ultimately null ref once the property tries being accessed.

jonsequitur: Thanks! I realized I made this change on another branch where I'm refactoring toward fixing #723. These APIs may change significantly by the time that's done but the refactorings are worth merging earlier which will actually fix this issue.

Issue #166, opened 2018-08-01
add support for outputting content in different formats


jonsequitur:
A useful capability would be a model whereby developers can set the output based on an object rather than explicitly writing text. We could provide a number of ways of representing that output so that the end user can expect consistent behaviors and choose output formats that suit their needs.

Here's an example of how this might work:

```csharp
 var commandLine = new CommandLineBuilder()
                              .AddOutputFormat(OutputFormats.Json)
                              .AddOutputFormat(OutputFormats.Csv)
                              .AddOutputFormat(OutputFormats.TabDelimited)
                              .AddOutputFormat(OutputFormats.Xml);
```

The above would also add an option that would be visible in help:

```
>my.exe -h

Options:
  --format [json|csv|tab-delimited|xml]        Specifies the output format. 
```

amervitz:
Has there been any progress on this, or are there any alternative methods for obtaining the return value of a command so that it can be manually formatted? I have multiple commands I want to be able to execute without having `Console.WriteLine` statements in them. Unfortunately the `InvokeAsync` method on commands only returns an integer. I have tried to replicate the [InvocationPipeline](https://github.com/dotnet/command-line-api/blob/master/src/System.CommandLine/Invocation/InvocationPipeline.cs) methods in an effort to capture the task result, but get stuck on https://github.com/dotnet/command-line-api/blob/f26d9805b891b7eed77522e6bc42f8c2cd93622c/src/System.CommandLine/Invocation/InvocationPipeline.cs#L44 because the Middleware property isn't public.

Thanks,

Alan

jonsequitur:
We do have work in progress on making the middleware more accessible and easier to customize, including giving you more control over the order of execution. That should be ready in the next few days. Did you try `CommandLineBuilder.AddMiddleware`?

As for returns types from command handlers, it's unlikely we would add the ability to return additional types directly, and might actually remove the `int` and `Task<int>` variants to reduce the number of overloads. The "return value" should be set on the `InvocationContext.InvocationResult`, as is done for example here:

https://github.com/dotnet/command-line-api/blob/f26d9805b891b7eed77522e6bc42f8c2cd93622c/src/System.CommandLine/Invocation/InvocationExtensions.cs#L175

The next question, of how to format the output, likely falls into the `System.CommandLine.Rendering` area. The view concepts there are a sketch of the direction we're looking in, so that you can express a layout but we'll handle rendering correctly based on the terminal type, output redirection, etc.

[...]

PR #734, opened 2020-01-10
Add ResultObject to InvocationContext to store return value of command execution

amervitz:
Please see the changes I've made so far for #166 to access the return value of a command. I'm not sure if this is the desired approach for accomplishing this but hopefully this helps explain further what I'm trying to accomplish. A lot of the tests fail now though, so there must be something problematic about what I've done. I would appreciate any recommendations for alternate desired implementation approaches to take.

I added a `InvocationContext.ResultObject` to store the return value of a command. I did not try to set the value on the `InvocationContext.InvocationResult` because it seems that property is reserved for storing parsing results and I wasn't sure if that made sense to use, and it has the `IInvocationResult` interface which I couldn't just set to the return value.

### DIFF
--- a/src/System.CommandLine/Invocation/InvocationContext.cs
+++ b/src/System.CommandLine/Invocation/InvocationContext.cs
@@ -4,6 +4,7 @@
 using System.CommandLine.Binding;
 using System.CommandLine.Parsing;
 using System.Threading;
+using System.Threading.Tasks;
 
 namespace System.CommandLine.Invocation
 {
@@ -35,7 +36,7 @@ namespace System.CommandLine.Invocation
 
         public int ResultCode { get; set; }
 
-        public object ResultObject { get; set; }
+        internal object InvokeResult { get; set; }
 
         public IInvocationResult InvocationResult { get; set; }
 
@@ -66,6 +67,27 @@ namespace System.CommandLine.Invocation
             }
 
             return _cts.Token;
+        }
+
+        /// <summary>
+        /// Set <see cref="InvokeResult"/> to the result of the command that was invoked.
+        /// </summary>
+        /// <param name="value">The result of the command invocation. When a task, it's assumed to already have been awaited.</param>
+        internal void SetInvokeResult(object value)
+        {
+            if (value is Task task)
+            {
+                var result = task.GetType().GetProperty("Result").GetValue(value);
+
+                if (result != null)
+                {
+                    InvokeResult = result;
+                }
+            }
+            else if (value != null)
+            {
+                InvokeResult = value;
+            }
         }
 
         public void Dispose()
--- a/src/System.CommandLine/Invocation/InvocationContext.cs
+++ b/src/System.CommandLine/Invocation/InvocationContext.cs
@@ -35,6 +35,8 @@ namespace System.CommandLine.Invocation
 
         public int ResultCode { get; set; }
 
+        public object ResultObject { get; set; }
+
         public IInvocationResult InvocationResult { get; set; }
 
         internal event Action<CancellationTokenSource> CancellationHandlingAdded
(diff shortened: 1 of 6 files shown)


jonsequitur:
I would recommend using the `InvocationContext.InvocationResult` rather than adding a property. It's intended to be a general purpose property to store the result of the invocation. We currently have these implementations of `IInvocationResult`:

`HelpResult`
`ParseDirectiveResult`
`ParseErrorResult`
`SuggestDirectiveResult`

Have a look at `InvocationPipeline.cs` for an example of how it's used.

amervitz:
I've made changes to use middleware to obtain the return value of commands via a new `IInvocationResult` implementation called `ReturnValueResult`. An example of using it is in the unit test created.

There were a couple complications.

1. From what I can tell, all middleware execute before the command is invoked, so I store the `ReturnValueResult` value in a temporary variable within the `UseMiddleware` code and access it again after the command executes. A potential improvement would be if middleware could execute after the command is invoked, so that the code that runs within `UseMiddleware` could obtain the completed `IInvocationResult` .
2. The invocation result object is still stored on the `InvocationContext` so that the new `ReturnValueResult` can access it. I could not find another place to store it.
3. I could not figure out how to return a value from a delegate based handler and was only able to do it from a `MethodInfo` based handler. It seems all the `CommandHandler.Create` overloads for `Func` have been designed to return integers only, but I'm not 100% sure.
4. All the unit tests are passing now. I had to update `CommandHandler.GetResultCodeAsync` to allow for command return types beyond just `void` and `int`, and `null`, it would throw an exception if any other type was returned.

I am going to hold on making further unit tests until getting some feedback and guidance on these complications.

Thanks,

Alan

jonsequitur:
This is really helpful. A few thoughts:

* We can make this work independent of the middleware. One of the capabilities of middleware is to replace a previous pipeline segment's result, so we don't actually know the final, committed result until after the pipeline runs, at which point we call `IInvocationResult..Apply` here: https://github.com/dotnet/command-line-api/blob/f5ee0abf24d03fb0827abec4c4f1d096f65a823f/src/System.CommandLine/Invocation/InvocationPipeline.cs#L69. 

* Return values, including the delegate and `Func<int>` variants, could be reworked to deal with any type. So instead of handlers returning `Task` or `Task<int>` we would have `Task` or `Task<T>` (where `Task` means, as it does now, that we should set the result code to 0 as long as no exception is thrown.)

* Since the developer using the library will know the type at compile time, `ReturnValueResult` can be `ReturnValueResult<T>`.

src/System.CommandLine/Invocation/InvocationContext.cs
```diff
@@ -66,6 +69,27 @@ public CancellationToken GetCancellationToken()
...
 
+        /// <summary>
+        /// Set <see cref="InvokeResult"/> to the result of the command that was invoked.
+        /// </summary>
+        /// <param name="value">The result of the command invocation. When a task, it's assumed to already have been awaited.</param>
+        internal void SetInvokeResult(object value)
+        {
+            if (value is Task task)
+            {
+                var result = task.GetType().GetProperty("Result").GetValue(value);
```
jonsequitur: Just a warning, calling `Task<T>.Result` should generally be avoided. If the task is not yet completed this can cause a deadlock under certain circumstances. https://docs.microsoft.com/en-us/archive/msdn-magazine/2013/march/async-await-best-practices-in-asynchronous-programming

Anyway, I think if we move this evaluation into the `InvocationPipeline` you won't need to do anything like this.

amervitz:
Would you like me to try make some of these changes?

src/System.CommandLine/Invocation/InvocationContext.cs
amervitz: Would you like me to try make some of these changes?

amervitz:
I could work on updating the delegates to return `Task` and `Task<T>`, but I could use more guidance on what to do for replacing the previous pipeline segment's result.

jonsequitur:
You won't need to replace the result set during a previous pipeline segment, if any. It's just important to be aware that this can happen, and that the place to actually apply the result is after the middleware has all been run, in `InvocationPipeline.GetResultCode`.

PR #736, opened 2020-01-12
Consolidate argument default values, custom conversion, and validation

jonsequitur:
This will address #723.


jonsequitur:
I'm going to merge this while leaving #723 open for further discussion. This leaves both the `TryConvertArgument<T>` and `ParseArgument<T>` variants in the code while we work out whether one is preferable to the other.

Issue #649, opened 2019-09-29
enable public API for specifying middleware order 


jonsequitur:
Currently, the public interface for specifying the order in which middleware is executed confines added middlewares to a specific section of the pipeline, e.g. users can't add middleware that executes before the exception handler, the intention being that the exception handler can catch exceptions thrown anywhere else in the middleware pipeline.

Allowing public access to this makes it possible to break existing behaviors, but might be useful in advanced scenarios.

PR #740, opened 2020-01-15
make control of middleware order public

jonsequitur:
This addresses #649.

### DIFF
--- a/src/System.CommandLine/Invocation/MiddlewareOrder.cs
+++ b/src/System.CommandLine/Invocation/MiddlewareOrder.cs
@@ -1,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace System.CommandLine.Invocation
+{
+    public enum MiddlewareOrder : byte
+    {
+        ProcessStart = byte.MinValue,
+        ExceptionHandler = ProcessStart + 10,
+        Configuration = ExceptionHandler + 10,
+        Default = Configuration + 50,
+        ErrorReporting = Default + 10,
+    }
+}
--- a/src/System.CommandLine/Invocation/MiddlewareOrder.cs
+++ b/src/System.CommandLine/Invocation/MiddlewareOrder.cs
@@ -3,12 +3,26 @@
 
 namespace System.CommandLine.Invocation
 {
-    public enum MiddlewareOrder : byte
+    public enum MiddlewareOrder
     {
-        ProcessStart = byte.MinValue,
-        ExceptionHandler = ProcessStart + 10,
-        Configuration = ExceptionHandler + 10,
-        Default = Configuration + 50,
-        ErrorReporting = Default + 10,
+        ExceptionHandler = -2000,
+        Configuration = -1000,
+        Default = default,
+        ErrorReporting = 1000,
+    }
+
+    internal enum MiddlewareOrderInternal
+    {
+        Startup = -4000,
+        ExceptionHandler = -3000,
+        ConfigureConsole = -2500,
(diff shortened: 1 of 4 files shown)


src/System.CommandLine/Invocation/MiddlewareOrder.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace System.CommandLine.Invocation
+{
+    public enum MiddlewareOrder : byte
+    {
+        ProcessStart = byte.MinValue,
```
Keboo: This effectively prevents anyone from injecting anything before us. Is there a reason for this restriction?

I would also suggest that MiddlewareOrder.Default should equal default(MiddlewareOrder).

jonsequitur: The default should be after most of the middleware we've defined, not at the beginning. The existing middleware that uses the `ProcessStart` and `ExceptionHandler` positions will be incorrect if other middleware executes before them. Similarly, if `ErrorReporting` happens before custom middleware, then the custom middleware won't be reached.

We're opening this up for advanced scenarios so with that in mind it's probably reasonable for `ProcessStart` to be at `MinValue + 10`.

Also, all of these could use better names.

src/System.CommandLine/Invocation/MiddlewareOrder.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace System.CommandLine.Invocation
+{
+    public enum MiddlewareOrder : byte
```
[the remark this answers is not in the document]
jonsequitur: Only that smaller, and unsigned, numbers seemed easier to reason about than larger numbers, though we might consider spreading them out and fixing every internal implementation at a distinct number to remove the possibility of behavioral differences based on the order in which the middleware methods are called.

src/System.CommandLine/Invocation/MiddlewareOrder.cs
Keboo: I agree that ProcesStart should be greater than the minimum possible value to allow for people to inject middleware before the default stuff (this is also making the assumption that they include the default middleware at all). I think it is important to not assume that our middleware will always be the first one that needs to be executed.

On the note of individual values, I do like the idea of spreading out the predefined values a bit. I would probably suggest giving an enum member to each of the built-int middleware that we have define. I think then rather than trying to reason able which middleware ends up at each value, you simply can look at the enum name and value and reason through it.

src/System.CommandLine/Invocation/MiddlewareOrder.cs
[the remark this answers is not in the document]
Keboo: My thoughts on byte. I agree that small unsigned numbers are easy to reason about. However, having a member on `MiddlewareOrder` called "Default" that does not line up to the default value of the enum seems wrong. Perhaps a better name could be "CustomProcessing"?

src/System.CommandLine/Invocation/MiddlewareOrder.cs
jonsequitur: Because the built-in middleware will certainly change over time, this would make breaking changes to this enum common. And for most cases the specific order isn't going to matter and should not be a contract.

jonsequitur: Because the built-in middleware will certainly change over time, this would make breaking changes to this enum common. And for most cases the specific order isn't going to matter and should not be a contract.

PR #743, opened 2020-01-20
Xml comments: helping discovery of Invoke

KathleenDollard:


src/System.CommandLine/Command.cs
```diff
@@ -73,7 +73,7 @@ public Command(string name, string description = null) : base(new[] { name }, de
         public bool TreatUnmatchedTokensAsErrors { get; set; } = true;
 
         /// <summary>
-        /// The action to perform when the command is executed. 
+        /// The action to perform when the command is invoked. 
         /// </summary>
```
jonsequitur: Adding some remark comments here with links to the various `Invoke` and `InvokeAsync` methods / extension methods on `Parser` and `ICommand` would help as well.

KathleenDollard:
Out of date

PR #748, opened 2020-01-23
FileSystemInfo binding

jonsequitur:
This enables a convention by which the model binder will bind a path string to a `FileSystemInfo` parameter as either `DirectoryInfo` or `FileInfo` depending on which it is.

For nonexistent paths that don't end in a directory separator, it chooses `FileInfo`.

Thoughts on whether this is a reasonable convention?

### DIFF
--- a/src/System.CommandLine.Tests/Binding/ModelBindingCommandHandlerTests.cs
+++ b/src/System.CommandLine.Tests/Binding/ModelBindingCommandHandlerTests.cs
@@ -297,6 +297,7 @@ namespace System.CommandLine.Tests.Binding
         [InlineData(typeof(FileSystemInfo), true, nameof(ExistingFile))]
         [InlineData(typeof(FileSystemInfo), true, nameof(ExistingDirectory))]
         [InlineData(typeof(FileSystemInfo), true, nameof(NonexistentPathWithTrailingSlash))]
+        [InlineData(typeof(FileSystemInfo), true, nameof(NonexistentPathWithTrailingAltSlash))]
         [InlineData(typeof(FileSystemInfo), true, nameof(NonexistentPathWithoutTrailingSlash))]
 
         [InlineData(typeof(string[]), false)]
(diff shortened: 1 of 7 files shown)


src/System.CommandLine.Tests/Binding/ModelBindingCommandHandlerTests.cs
```diff
@@ -443,14 +479,82 @@ public void Apply(InvocationContext context)
...
 
             BindingTestCase.Create<FileInfo[]>(
-                $"{Path.Combine(Directory.GetCurrentDirectory(), "file1.txt")} {Path.Combine(Directory.GetCurrentDirectory(), "file2.txt")}",
+                $"{Path.Combine(ExistingDirectory(), "file1.txt")} {Path.Combine(ExistingDirectory(), "file2.txt")}",
                 o => o.Select(f => f.FullName)
                       .Should()
-                      .BeEquivalentTo(new[] { Path.Combine(Directory.GetCurrentDirectory(), "file1.txt"), Path.Combine(Directory.GetCurrentDirectory(), "file2.txt") })),
+                      .BeEquivalentTo(new[]
+                      {
+                          Path.Combine(ExistingDirectory(), "file1.txt"),
```
Keboo: There seems to be an (arguably accurate) assumption that all calls to `ExistingDirectory()` return the same value. I am wondering if it would be better to cache the value in a local and use that to make it more obvious when we are expecting values to match.

src/System.CommandLine/Binding/ArgumentConverter.cs
```diff
@@ -13,6 +14,24 @@ namespace System.CommandLine.Binding
...
+        private static readonly Dictionary<Type, Func<string, object>> _converters = new Dictionary<Type, Func<string, object>>
+        {
+            [typeof(FileSystemInfo)] = value =>
+            {
+                if (Directory.Exists(value))
+                {
+                    return new DirectoryInfo(value);
+                }
+
+                if (value.EndsWith(Path.DirectorySeparatorChar.ToString()))
```
Keboo: What about also checking for `Path.AltDirectorySeparatorChar`?

Issue #587, opened 2019-07-17
add support for global options


guy-lud:
Hi,
I wonder if adding a global options by adding it to the `RootCommand` so it does not matter where in the command line i'll put them The value will be available?

jonsequitur:
This can be done now, but it would have to be specified once for each subcommand, e.g. you would have to walk the command tree and add the option to each command. Take a look at #548 for an example. 

Maybe an API something like this would simplify things when you want the option to be global?

```csharp
var root = new RootCommand();
root.AddOptionToAllSubcommands(option);
```

phi1ipp:
Would be nice to have it, I also ran into the same necessity.

Alxandr:
Doing something like `AddOptionToAllSubcommands` requires that the sub-commands are available before the options are added though. Wouldn't it make more sense to allow options to be inherited?

jonsequitur:
The intention could be captured at the time `AddOptionToAllSubcommands` is called and then applied when the `CommandLineConfiguration` is actually created.

> Wouldn't it make more sense to allow options to be inherited?

Could you say more about this?

Alxandr:
What I mean is that an option could have a flag (`Inherited` for instance), and any sub-command of a command will inherit all it's inherited options. Example (pseudo - don't have the actual APIs available atm):

```csharp
var root = new RootCommand();
var repo = new Command("repo");
var updateRepo = new Command("update");

var read = new Command("read");

var source = new Option("--source") { Inherited = true };
source.Argument = new Argument<string>("URI");
repo.Add(source);

// valid: prog.exe repo --source http://test.com
// valid: prog.exe repo update --source http://test.com
// invalid: prog.exe --source http://test.com
// invalid: prog.exe read --source http://test.com
```

jonsequitur:
Thanks for the clarification.

Alxandr:
If this would be interesting, I can take a shot at implementing it. I think it would be pretty straight forward.

jonsequitur:
So far, none of the properties of a `Command`, `Option`, or `Argument` (symbols) directly affect the behavior of the others. The relationship between these objects is determined by their hierarchy, and reuse is supported like this:

```csharp
var optionX = new Option("-x");

var root = new RootCommand
{
    optionX,
    new Command("subcommand1")
    {
        optionX // optionX is valid here
    },
    new Command("subcommand2") // optionX is not valid here
}
```

[...]

Issue #351, opened 2018-12-23
Consider supporting POSIX, PowerShell, and Windows prefixes for all options by default


jonsequitur:
Many users have different expectations of command line syntax depending on which ecosystem they're used to. 

It may help users if all prefixes "just work". For example, given the following code:

```csharp
var myApp = new RootCommand("myapp");
myApp.AddOption(new Option("option"));
```

then the following command lines would be equivalent:

```console
> myapp --option
> myapp -option
> myapp /option
```

The developer should retain the ability to opt out of this behavior, to specify exactly which prefixes they support, and to mix and match within one application. This is currently supported and should be retained.

These issues have a related motivation: #43, #133.

jonsequitur:
@KathleenDollard @MarkMichaelis

KathleenDollard:
Yes, yes, yes. 

At least on first and third. The middle presents a problem in Posix, so I'm less confident on that being the default

Is your example `--option` or 

`--oz --panem --toon-town --interactive --narnia`

Perhaps we could make this an error and have a way to communicate "Danger, danger" if the letters of an option could be a combination of other boolean options one letter aliases.

jonsequitur:
During parser building we already throw if there are conflicting symbols.

zivkan:
POSIX programs usually allow single character arguments to be provided in "batch" using a single switch. For example, the following are all equivalent:

```
> ls -lah
> ls -l -a -h
> ls -l --all --human-readable
```

If powershell style single `-` switches are used for multi-character aguments, does that mean that System.CommandLine by design will not support argument batching?

I haven't used either Linux or Powershell very much in my life, but I have used Linux a lot more than Powershell, so I personally prefer batched arguments over single switch multi-character arguments. But that's just my personal preference.

jonsequitur:
@zivkan `System.CommandLine` supports the POSIX batching convention: https://github.com/dotnet/command-line-api/wiki/Syntax-Concepts-and-Parser#bundling.

In the case where `abc` is a valid token then that token would be chosen rather than unbundling into `-a -b -c`.  We already perform a check before unbundling, e.g. each single character must be valid tokens for your program.

PR #771, opened 2020-01-27
add global options, make -h a proper option

### LINKED ISSUE #506: UseHelp does not include help description
How can I supply a help description for the argument `--help` when using `System.CommandLine.Invocation.InvocationExtensions.UseHelp`? 

As fas as I saw it's not possible to give a description on `UseHelp` and there is also not automatically a help description added.

Several dotnet command, e.g. `dotnet build` have the help argument documented, thus I would like to Keep that also using `command-line-api`.

Is here an extension needed or didn't I found the required API? Thanks.

### LINKED ISSUE #691: CommandExtensions.Parse returns ParseResult with an Error for the --help argument
Not sure if this is by design or not, but CommandExtensions.Parse will treat --help (and it's variations) as an invalid argument even though that's part of the built-in, default help feature.

The code below is copied from [Your first app with System.CommandLine](https://github.com/dotnet/command-line-api/wiki/Your-first-app-with-System.CommandLine) and returns the following error message (`Unrecognized command or argument '-h'.`).  And it will also return the default help text.  

```C#
static int Main(string[] args)
{
    // Setup test scenario: Pass a help argument.
    args = new[] { "-h" };

    // Create a root command with some options
    {
        new Option(
            "--int-option",
            "An option whose argument is parsed as an int")
        {
            Argument = new Argument<int>(defaultValue: () => 42)
        },
        new Option(
            "--bool-option",
            "An option whose argument is parsed as a bool")
        {
            Argument = new Argument<bool>()
        },
        new Option(
            "--file-option",
            "An option whose argument is parsed as a FileInfo")
        {
            Argument = new Argument<FileInfo>()
        }
    };

    rootCommand.Description = "My sample app";

    rootCommand.Handler = CommandHandler.Create<int, bool, FileInfo, bool>((intOption, boolOption, fileOption, help) =>
    {
        Console.WriteLine($"The value for --int-option is: {intOption}");
        Console.WriteLine($"The value for --bool-option is: {boolOption}");
        Console.WriteLine($"The value for --file-option is: {fileOption?.FullName ?? "null"}");
    });

    // START - New code to check for parsing errors.
    var result = rootCommand.Parse(args);
    if (result.Errors.Count > 0)
    {
        Console.WriteLine($"{ result.CommandResult.ErrorMessage}");  // This returns: null

        foreach (var e in result.Errors)
            Console.WriteLine($"{e.Message}.");  // This returns: Unrecognized command or argument '-h'.
    }
    // END - New code to check for parsing errors.

    // Parse the incoming args and invoke the handler
    return rootCommand.InvokeAsync(args).Result;
}
```

I am using the following:
0.3.0-alpha.19573.2
.NET Core 3.0
Any CPU

### LINKED ISSUE #766: Issues with Help When TypoCorrection is Enabled
`UseTypoCorrections` causes problems with the `--help` option in the latest version when other global arguments are present.

For example with:
```cs
commandLineBuilder.UseVersionOption().UseHelp().UseTypoCorrections().UseParseErrorReporting().CancelOnProcessTermination();
commandLineBuilder.UseMyMiddleware();
```

Where `UseMyMiddleWare` registers:
```cs
commandLineBuilder.AddOption(new Option(new[] {"--file", "-f"}, "Change where ImageCaster should find the configuration file")
{
    Argument = new Argument<FileInfo>("file", () => new FileInfo("imagecaster.yml")).ExistingOnly()
});
```

When executing `--help`, I see `'--help' was not matched. Did you mean '--file'?`.  
When executing `-h`, I see `'-h' was not matched. Did you mean '-l', or '-t', or '-L', or '-c', or '-f'?`.

When `UseTypoCorrections` is set lower, such as 2, it addresses the issue with `--help` and `--file` colliding. (Presumably because they have a distance of 3 which is the default.)

However, I need to remove it entirely to get `-h` working.

It seems the typo correction doesn't recognize `help` or it's aliases so reports problems that aren't actually there.

jonsequitur:
This reimplements help as a proper `Option`, which fixes #691. In order for the help option to be a single instance, looked up using `FindResultFor` after parsing, I also added support for global options (#587). 

This should also fix #766, the root cause of which is #691.

This will also fix #506.

### DIFF
--- a/src/System.CommandLine/CommandLineConfiguration.cs
+++ b/src/System.CommandLine/CommandLineConfiguration.cs
@@ -20,7 +20,6 @@ namespace System.CommandLine
         public CommandLineConfiguration(
             IReadOnlyCollection<Symbol> symbols,
             IReadOnlyCollection<char> argumentDelimiters = null,
-            IReadOnlyCollection<string> prefixes = null,
             bool enablePosixBundling = true,
             bool enableDirectives = true,
             ValidationMessages validationMessages = null,
@@ -81,29 +80,27 @@ namespace System.CommandLine
 
             _symbols.Add(RootCommand);
 
+            AddGlobalOptionsToChildren(rootCommand);
+
             EnablePosixBundling = enablePosixBundling;
             EnableDirectives = enableDirectives;
             ValidationMessages = validationMessages ?? ValidationMessages.Instance;
             ResponseFileHandling = responseFileHandling;
             _middlewarePipeline = middlewarePipeline;
             _helpBuilderFactory = helpBuilderFactory;
-            Prefixes = prefixes;
+        }
 
-            if (prefixes?.Count > 0)
+        private void AddGlobalOptionsToChildren(Command parentCommand)
+        {
+            foreach (var globalOption in parentCommand.GlobalOptions)
             {
-                foreach (var symbol in symbols)
+                foreach (var child in parentCommand.Children.FlattenBreadthFirst(c => c.Children))
                 {
-                    if (symbol is Option option)
+                    if (child is Command childCommand)
                     {
-                        foreach (var alias in option.RawAliases.ToList())
+                        if (!childCommand.Children.IsAnyAliasInUse(globalOption, out var inUse))
                         {
-                            if (!prefixes.All(prefix => alias.StartsWith(prefix)))
-                            {
-                                foreach (var prefix in prefixes)
-                                {
-                                    option.AddAlias(prefix + alias);
-                                }
-                            }
+                            childCommand.AddOption(globalOption);
                         }
                     }
                 }
--- a/src/System.CommandLine.DragonFruit.Tests/CommandLineTests.cs
+++ b/src/System.CommandLine.DragonFruit.Tests/CommandLineTests.cs
@@ -2,7 +2,6 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System.CommandLine.Rendering;
-using System.Reflection;
 using System.Threading.Tasks;
 using FluentAssertions;
 using Xunit;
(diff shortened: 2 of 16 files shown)


src/System.CommandLine/CommandLineConfiguration.cs
```diff
@@ -81,29 +80,27 @@ public class CommandLineConfiguration
...
-            if (prefixes?.Count > 0)
+        private void AddGlobalOptionsToChildren(Command parentCommand)
+        {
+            foreach (var globalOption in parentCommand.GlobalOptions)
             {
-                foreach (var symbol in symbols)
+                foreach (var child in parentCommand.Children.FlattenBreadthFirst(c => c.Children))
                 {
-                    if (symbol is Option option)
+                    if (child is Command childCommand)
```
Keboo: Consider `.OfType<Command>()` on the foreach to reduce nesting.

jonsequitur:
A consequence of this change is that it also adds help output to `--help` which is a fairly common behavior and has been requested (#506), but it might be a little *too* helpful. 

```console
Usage:
  myapp [options]

Options:
  --version                Show version information
  -h, /h, -?, /?, --help   Show help and usage information
  -f                       Do some other stuff
```

Keboo:
Any thoughts on changing the output on help to not display items that only differ by prefix? I would assume that would then make the output look something like:
```
Usage:
  myapp [options]

Options:
  --version                Show version information
  -h, -?, --help          Show help and usage information
  -f                            Do some other stuff
```

jonsequitur:
I like that. It does leave aside the question of Windows-style option prefixes (#351).

KathleenDollard:
1) Looking at the issues list, a trifecta is always nice :)
2) I think this is a good change. I do think it highlights that we have more work to do on the help output (including prefixes and how --help appears from this conversation). I don't think that should hold up this PR. 

Let's do it!

jonsequitur:
> Any thoughts on changing the output on help to not display items that only differ by prefix? I would assume that would then make the output look something like:
> 
> ```
> Usage:
>   myapp [options]
> 
> Options:
>   --version               Show version information
>   -h, -?, --help          Show help and usage information
>   -f                      Do some other stuff
> ```

@Keboo Looks what you've made me do...

https://github.com/dotnet/command-line-api/pull/771/commits/c9d2900f96ee909f6c17484a05054dab22c941c4#diff-ed0644cdb2d64176756014bb5d1a7115R335-R342

Keboo:
@jonsequitur nice I like it

PR #779, opened 2020-02-10
fix: default values for option arguments was not validated

beppemarazzi:
When a default is specified for an option's argument, it is not validated.
In my use-case i have this setup:
```
var ret = new Command(...)
{
  new Option("-gv", "...")
  {
    Required = true,                    
    Argument = new Argument<FileInfo>(() => new FileInfo("defaultfile.dat"))
      .ExistingOnly()
  },
};
```

and at run time i don't receive any error when `defaultfile.dat` does not exists...

My PR will fix this issue.

### DIFF
--- a/src/System.CommandLine.Tests/OptionTests.cs
+++ b/src/System.CommandLine.Tests/OptionTests.cs
@@ -3,6 +3,7 @@
 
 using System.CommandLine.Invocation;
 using System.CommandLine.Parsing;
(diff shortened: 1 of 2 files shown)


jonsequitur:
Thanks, @beppemarazzi!

beppemarazzi:
@jonsequitur: Thank you for this great job!!!!!

PR #781, opened 2020-02-13
don't mutate command on subsequent calls to Invoke extension

jonsequitur:
This fixes a bug where calling the `Command.Invoke` / `Command.InvokeAsync` extension methods more than once on the same `Command` instance would throw due to global options being added twice to the implictly-created parser. Now, that parser is cached and reused.

### DIFF
--- a/src/System.CommandLine/CommandExtensions.cs
+++ b/src/System.CommandLine/CommandExtensions.cs
@@ -42,11 +42,12 @@ namespace System.CommandLine
 
         private static InvocationPipeline GetInvocationPipeline(Command command, string[] args)
         {
-            var parser = new CommandLineBuilder(command)
-                         .UseDefaults()
-                         .Build();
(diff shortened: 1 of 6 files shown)


src/System.CommandLine/CommandExtensions.cs
```diff
@@ -42,11 +42,12 @@ public static class CommandExtensions
 
         private static InvocationPipeline GetInvocationPipeline(Command command, string[] args)
         {
-            var parser = new CommandLineBuilder(command)
-                         .UseDefaults()
-                         .Build();
+            var parser = command.ImplicitParser ??=
```
Keboo: You shouldn't need to explicitly set `ImplicitParser` here. Calling `.Build` will set it on the command passed into the builder.
```suggestion
            var parser = command.ImplicitParser ??
```

PR #799, opened 2020-02-25
Bind options by name

### LINKED ISSUE #798: Option is not bound by name
It seems that option value only binds to a command handler parameter which is named after one of the option aliases. Options also have the `Name` property, which seems to be unused. It would be nice if the library will also bind options by name the same way as arguments are bound, so I could use more sensible name for my command handler parameters.

An example of the code which I would like to work, but it isn't:

```cs
class Program
{
    public static async Task<int> Main(string[] args)
    {
        {
            new Option("-H", "Authentication header")
            {
                Argument = new Argument<HttpHeaderInfo[]>("header"),
                Name = "headers",
            },
            new Argument<DirectoryInfo>("target-dir")
            {
                Description = "Path to the directory where downloaded files will be placed",
                Arity = ArgumentArity.ExactlyOne,
            },
        };
        rootCommand.Handler = CommandHandler.Create((DirectoryInfo targetDir, HttpHeaderInfo[] headers) =>
        {
            // `targetDir` is successfully bound by name
            // `headers` is not bound, it has to be named `h` which is inconvenient
        });
        return await rootCommand.InvokeAsync(args);
    }
}
```

Example invocation:
`program.exe "c:\TargetDir" -H "accept: application/json" -H "cookie: key=value"`

As a workaround I could add `headers` as an alias for the option, but this would pollute help output.

DarkDaskin:
This PR fixes #798 by matching parameter name with option name in addition to option aliases.

### DIFF
--- a/src/System.CommandLine/Binding/Binder.cs
+++ b/src/System.CommandLine/Binding/Binder.cs
@@ -15,7 +15,7 @@ namespace System.CommandLine.Binding
                           StringComparison.OrdinalIgnoreCase);
 
         internal static bool IsMatch(this string parameterName, ISymbol symbol) =>
-            symbol.Aliases.Any(parameterName.IsMatch);
+            parameterName.IsMatch(symbol.Name) || symbol.Aliases.Any(parameterName.IsMatch);
 
         internal static bool IsNullable(this Type t)
         {


jonsequitur:
Thanks @DarkDaskin!

jonsequitur:
@DarkDaskin I forgot to ask for a test but I've got you covered: #800

PR #810, opened 2020-03-04
Fixes for 723

jonsequitur:
This fixes a couple of bugs related to #723:

* Errors from custom parse delegates were not propagating when the delegate was called to provide a default value
* When multiple arguments with custom parse delegates were present, only the first would propagate custom errors.

There are two behavior changes introduced here as well:

- [x] Implicit tokens are no longer added to the results. Tokens should only reflect what was actually parsed.
- [ ] #811 ~Default value factories will be re-evaluated rather than cached, which is important because default values emitted by custom parse delegates are expected to be allowed to vary based on other aspects of the command line input.~

I'll check these off as tests are added.


src/System.CommandLine/Parsing/ArgumentResult.cs
```diff
@@ -55,9 +60,25 @@ internal ParseError CustomError(Argument argument)
...
+                            argument,
+                            defaultValue);
+                    }
+                    else
+                    {
+                        return ArgumentConversionResult.Failure(
+                            argument,
+                            argumentResult.ErrorMessage);
+                    }
+                }
```
rynowak: <3

rynowak: Thanks for the quick turnaround

PR #844, opened 2020-04-10
Proposed changes to implement explicit binding of results to instance…

markolbert:
… properties (as opposed to doing so implicitly by matching names).

I've added comments to the various changes I made.


markolbert:
I updated the code to fix, I think, the build-and-test problems. But while I've pushed the changes over to my forked repository it doesn't seem to have restarted the code checking process. How do I do that? Or do I need to cancel this PR and submit a new one?

markolbert:
I came up with a better way of implementing UseObjectBinding().

PR #848, opened 2020-04-12
enable easier custom model binding

jonsequitur:
There are a few changes here:

* Added `BindingContext.AddModelBinder` to enable specifying a model binder other than the one that would be generated by convention.
* `IValueDescriptor.Type` is renamed to `ValueType`
* Removed the expression-based `HandlerDescriptor.FromExpression` methods.


src/System.CommandLine/Binding/BindingContext.cs
```diff
@@ -68,31 +70,31 @@ public void AddService<T>(Func<IServiceProvider, T> factory)
                 throw new ArgumentNullException(nameof(factory));
             }
 
-            ServiceProvider.AddService(typeof(T), s => factory(s));
+            ServiceProvider.AddService(typeof(T), s => factory(s)!);
```
Keboo, line 73: Why are we asserting not null from the factory? Wouldn't throwing be better?

src/System.CommandLine/Invocation/ServiceProvider.cs
```diff
@@ -7,29 +7,27 @@
...
                             [typeof(CancellationToken)] = _ => CancellationToken.None,
-                            [typeof(IHelpBuilder)] = _ => _bindingContext.ParseResult.Parser.Configuration.HelpBuilderFactory(_bindingContext),
-                            [typeof(BindingContext)] = _ => _bindingContext
+                            [typeof(IHelpBuilder)] = _ => bindingContext.ParseResult.Parser.Configuration.HelpBuilderFactory(bindingContext),
+                            [typeof(BindingContext)] = _ => bindingContext
                         };
         }
 
-        public void AddService<T>(Func<IServiceProvider, T> factory) => _services[typeof(T)] = p => factory(p);
+        public void AddService<T>(Func<IServiceProvider, T> factory) => _services[typeof(T)] = p => factory(p)!;
```
Keboo, line 30: I don't think we want to assert not null from the factory.

src/System.CommandLine/Invocation/ServiceProvider.cs
```diff
@@ -42,7 +40,7 @@ public object GetService(Type serviceType)
                 return factory(this);
             }
 
-            return null;
+            return null!;
```
Keboo, line 43: Any reason we are doing null! rather than making the return nullable?

src/System.CommandLine/Binding/BindingContext.cs
jonsequitur: Actually it's allowed to be null. Fixing.

PR #851, opened 2020-04-13
move test utilities into a Utilities namespace, add ConsoleAssertions

jonsequitur:


src/System.CommandLine.Tests/Utility/WindowsOnlyFactAttribute.cs
```diff
@@ -4,15 +4,15 @@
 using Microsoft.DotNet.PlatformAbstractions;
 using Xunit;
 
-namespace System.CommandLine.Tests
+namespace System.CommandLine.Tests.Utility
```
Keboo, line 7: I typically try to avid the "junk drawer" namespaces like "Utility" because it becomes a grab bag of stuff. Could we move this one into something like "System.CommandLine.Tests.Xunit"

src/System.CommandLine.Tests/Utility/ReleaseBuildOnlyFactAttribute.cs
```diff
@@ -3,7 +3,7 @@
 
 using Xunit;
 
-namespace System.CommandLine.Tests
+namespace System.CommandLine.Tests.Utility
```
Keboo, line 6: What about "System.CommandLine.Tests.Xunit"?

src/System.CommandLine.Tests/Utility/NonWindowsOnlyFactAttribute.cs
```diff
@@ -4,15 +4,15 @@
 using Microsoft.DotNet.PlatformAbstractions;
 using Xunit;
 
-namespace System.CommandLine.Tests
+namespace System.CommandLine.Tests.Utility
```
Keboo, line 7: What about "System.CommandLine.Tests.Xunit"?

src/System.CommandLine.Tests/Utility/LinuxOnlyTheory.cs
```diff
@@ -2,18 +2,17 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System.Runtime.InteropServices;
-using Microsoft.DotNet.PlatformAbstractions;
 using Xunit;
 
-namespace System.CommandLine.Tests
+namespace System.CommandLine.Tests.Utility
```
Keboo, line 7: What about "System.CommandLine.Tests.Xunit"?

src/System.CommandLine.Tests/Utility/RemoteExecutor.cs
```diff
@@ -10,7 +11,7 @@
 using System.Text;
 using System.Threading.Tasks;
 
-namespace System.CommandLine.Tests
+namespace System.CommandLine.Tests.Utility
```
Keboo, line 14: What about "System.CommandLine.Tests.Remote"?

src/System.CommandLine.Tests/Utility/RemoteExecution.cs
```diff
@@ -4,14 +4,13 @@
 using System.Diagnostics;
 using System.IO;
 using Xunit;
-using Xunit.Sdk;
 
-namespace System.CommandLine.Tests
+namespace System.CommandLine.Tests.Utility
```
Keboo, line 8: What about "System.CommandLine.Tests.Remote"?

src/System.CommandLine.Tests/Utility/WindowsOnlyFactAttribute.cs
jonsequitur: Agreed, though I tend to be less concerned about this with test projects. In this case, the child namespaces in the tests all corresponded to child namespaces in System.CommandLine:

image

So my thought was to make this easier to navigate in terms of the tests. Everything that's not tests, I swept into the junk drawer.

But I'm not strongly opinionated either way.

jonsequitur: (Sometimes I also put folders like this in parentheses, but that just garners funny looks.)

Keboo: Not strongly opinionated either, if we want to stick with the junk drawer i am fine with it

PR #853, opened 2020-04-14
Fix for multiple parameters that have the same type and default values

Keboo:
Previously this was caching the first model binder that was resolved based on the parameter's type. This would cause the second parameter to get resolved with the first parameter's default value.

This change stop caching the generated model binders.

### DIFF
--- a/src/System.CommandLine.DragonFruit.Tests/ConfigureFromMethodTests.cs
+++ b/src/System.CommandLine.DragonFruit.Tests/ConfigureFromMethodTests.cs
@@ -235,6 +235,17 @@ namespace System.CommandLine.DragonFruit.Tests
             _receivedValues.Should().BeEquivalentTo(true);
         }
 
+        [Fact]
+        public async Task Method_with_multiple_parameters_with_default_values_are_resolved_correctly()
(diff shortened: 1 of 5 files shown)


src/System.CommandLine.DragonFruit.Tests/ConfigureFromMethodTests.cs
```diff
@@ -235,6 +235,17 @@ public async Task Method_parameters_on_the_invoked_member_method_are_bound_to_ma
             _receivedValues.Should().BeEquivalentTo(true);
         }
 
+        [Fact]
+        public async Task Method_with_multiple_parameters_with_default_values_are_resolved_correctly()
```
jonsequitur, line 239: It looks like the bug is in `System.CommandLine` but the test is in `System.CommandLine.DragonFruit`. I'm assuming that means the same bug can express itself in non-DragonFruit code as well, so maybe add a test for this in `System.CommandLine.Tests`?

Keboo: Done

PR #856, opened 2020-04-15
C# 8 Nullable reference types

Keboo:
Starting work on handling C# 8 nullable reference types.

The current work only handles it in the core library. 

Outstanding TODO items:

- [x] Conditionally include nullable reference type attributes to support `net462`
- [x] Confirm upgrade to netstandard2.1 is acceptable. This is not explicitly needed since we will already be including attributes.
- [x] Handle possible `null` setter values on `Argument.ArgumentType`
- [x] `ParseResultVisitor.Result` currently throws `InvalidOperationException`. What should the error messages be for these?
- [x] `ErrorMessage` on `*Result` classes is declared as nullable, because some localization implementation could in theory return a null. Currently explicitly ignoring with null forgiveness operator.

### DIFF
--- a/src/System.CommandLine/Argument.cs
+++ b/src/System.CommandLine/Argument.cs
@@ -25,7 +25,7 @@ namespace System.CommandLine
         {
             if (!string.IsNullOrWhiteSpace(name))
             {
-                Name = name;
+                Name = name!;
             }
         }
 
--- a/src/System.CommandLine/Argument.cs
+++ b/src/System.CommandLine/Argument.cs
@@ -11,17 +11,17 @@ namespace System.CommandLine
 {
     public class Argument : Symbol, IArgument
     {
-        private Func<ArgumentResult, object> _defaultValueFactory;
+        private Func<ArgumentResult, object>? _defaultValueFactory;
         private readonly List<string> _suggestions = new List<string>();
         private readonly List<ISuggestionSource> _suggestionSources = new List<ISuggestionSource>();
-        private IArgumentArity _arity;
-        private TryConvertArgument _convertArguments;
+        private IArgumentArity? _arity;
+        private TryConvertArgument? _convertArguments;
 
         public Argument()
         {
         }
 
-        public Argument(string name) 
+        public Argument(string? name) 
         {
             if (!string.IsNullOrWhiteSpace(name))
             {
@@ -29,7 +29,7 @@ namespace System.CommandLine
             }
         }
 
-        internal HashSet<string> AllowedValues { get; private set; }
+        internal HashSet<string>? AllowedValues { get; private set; }
 
         public IArgumentArity Arity
         {
@@ -52,7 +52,7 @@ namespace System.CommandLine
             set => _arity = value;
         }
 
-        internal TryConvertArgument ConvertArguments
+        internal TryConvertArgument? ConvertArguments
         {
             get
             {
@@ -108,7 +108,7 @@ namespace System.CommandLine
             set => _convertArguments = value;
         }
 
-        public Type ArgumentType { get; set; }
+        public Type? ArgumentType { get; set; }
 
         internal List<ValidateSymbol<ArgumentResult>> Validators { get; } = new List<ValidateSymbol<ArgumentResult>>();
 
@@ -193,7 +193,7 @@ namespace System.CommandLine
             AllowedValues.UnionWith(values);
         }
 
-        public override IEnumerable<string> GetSuggestions(string textToMatch = null)
+        public override IEnumerable<string> GetSuggestions(string? textToMatch = null)
         {
             var fixedSuggestions = _suggestions;
 
--- a/src/System.CommandLine/Argument.cs
+++ b/src/System.CommandLine/Argument.cs
@@ -35,7 +35,7 @@ namespace System.CommandLine
         {
             get
             {
-                if (_arity == null)
+                if (_arity is null)
                 {
                     if (ArgumentType != null)
                     {
@@ -137,7 +137,7 @@ namespace System.CommandLine
 
         public void SetDefaultValueFactory(Func<object?> getDefaultValue)
         {
-            if (getDefaultValue == null)
+            if (getDefaultValue is null)
             {
                 throw new ArgumentNullException(nameof(getDefaultValue));
             }
@@ -156,7 +156,7 @@ namespace System.CommandLine
 
         public void AddSuggestions(IReadOnlyCollection<string> suggestions)
         {
-            if (suggestions == null)
+            if (suggestions is null)
             {
                 throw new ArgumentNullException(nameof(suggestions));
             }
@@ -166,7 +166,7 @@ namespace System.CommandLine
 
         public void AddSuggestionSource(ISuggestionSource suggest)
         {
-            if (suggest == null)
+            if (suggest is null)
             {
                 throw new ArgumentNullException(nameof(suggest));
             }
@@ -176,7 +176,7 @@ namespace System.CommandLine
 
         public void AddSuggestionSource(Suggest suggest)
         {
-            if (suggest == null)
+            if (suggest is null)
             {
                 throw new ArgumentNullException(nameof(suggest));
             }
@@ -186,7 +186,7 @@ namespace System.CommandLine
 
         internal void AddAllowedValues(IEnumerable<string> values)
         {
-            if (AllowedValues == null)
+            if (AllowedValues is null)
             {
                 AllowedValues = new HashSet<string>();
             }
--- a/src/System.CommandLine/Argument.cs
+++ b/src/System.CommandLine/Argument.cs
@@ -16,6 +16,7 @@ namespace System.CommandLine
         private readonly List<ISuggestionSource> _suggestionSources = new List<ISuggestionSource>();
         private IArgumentArity? _arity;
         private TryConvertArgument? _convertArguments;
+        private Type _argumentType = typeof(void);
 
         public Argument()
         {
@@ -37,7 +38,7 @@ namespace System.CommandLine
             {
                 if (_arity is null)
                 {
-                    if (ArgumentType != null)
+                    if (ArgumentType != typeof(void))
                     {
                         return ArgumentArity.Default(ArgumentType, this, Parents.FirstOrDefault());
                     }
@@ -57,7 +58,7 @@ namespace System.CommandLine
             get
             {
                 if (_convertArguments == null &&
-                    ArgumentType != null)
+                    ArgumentType != typeof(void))
                 {
                     if (ArgumentType.CanBeBoundFromScalarValue())
                     {
@@ -108,8 +109,11 @@ namespace System.CommandLine
             set => _convertArguments = value;
         }
 
-        //TODO: What if null gets set?
-        public Type ArgumentType { get; set; } = typeof(void);
+        public Type ArgumentType
+        {
+            get => _argumentType;
+            set => _argumentType = value ?? throw new ArgumentNullException(nameof(value));
+        }
 
         internal List<ValidateSymbol<ArgumentResult>> Validators { get; } = new List<ValidateSymbol<ArgumentResult>>();
 
--- a/src/System.CommandLine/Argument.cs
+++ b/src/System.CommandLine/Argument.cs
@@ -11,7 +11,7 @@ namespace System.CommandLine
 {
     public class Argument : Symbol, IArgument
     {
-        private Func<ArgumentResult, object>? _defaultValueFactory;
+        private Func<ArgumentResult, object?>? _defaultValueFactory;
         private readonly List<string> _suggestions = new List<string>();
         private readonly List<ISuggestionSource> _suggestionSources = new List<ISuggestionSource>();
         private IArgumentArity? _arity;
@@ -64,7 +64,7 @@ namespace System.CommandLine
                         if (Arity.MaximumNumberOfValues == 1 &&
                             ArgumentType == typeof(bool))
                         {
-                            _convertArguments = (ArgumentResult symbol, out object value) =>
+                            _convertArguments = (ArgumentResult symbol, out object? value) =>
                             {
                                 value = ArgumentConverter.ConvertObject(
                                     this,
@@ -114,12 +114,12 @@ namespace System.CommandLine
 
         public void AddValidator(ValidateSymbol<ArgumentResult> validator) => Validators.Add(validator);
 
-        public object GetDefaultValue()
+        public object? GetDefaultValue()
         {
             return GetDefaultValue(new ArgumentResult(this, null));
         }
 
-        internal object GetDefaultValue(ArgumentResult argumentResult)
+        internal object? GetDefaultValue(ArgumentResult argumentResult)
         {
             if (_defaultValueFactory is null)
             {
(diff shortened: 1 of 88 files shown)


src/System.CommandLine/Symbol.cs
```diff
@@ -95,7 +95,7 @@ private protected void AddArgumentInner(Argument argument)
 
         public SymbolSet Children { get; } = new SymbolSet();
 
-        public void AddAlias(string alias)
+        public void AddAlias(string? alias)
```
jcouv, line 98: >? [](start = 35, length = 1)

Since this method will throw on a `null` input, I'd mark as not nullable. This will help callers not pass a `null` value in.

jcouv:
>                                      .FirstOrDefault();

I'm surprised there was no warning here. `rootCommand` is declare not nullable, yet we could put a null value into it here and then we null test it below..

---
Refers to: src/System.CommandLine/CommandLineConfiguration.cs:66 in 74b3676. [](commit_id = 74b3676fd5f71169b25bb62ca32acb59c16d23e9, deletion_comment = False)

jcouv:
>             if (command == null)

parameter should probably be marked as nullable.
When annotating a project, I like to hunt not only for `null` values being created (the compiler is good at tracing those), but also null tests.

---
Refers to: src/System.CommandLine/Help/HelpBuilder.cs:567 in 74b3676. [](commit_id = 74b3676fd5f71169b25bb62ca32acb59c16d23e9, deletion_comment = False)

src/System.CommandLine/Parsing/ParseResultVisitor.cs
```diff
@@ -81,7 +81,7 @@ protected override void VisitCommandNode(CommandNode commandNode)
             var commandResult = _innermostCommandResult;
 
             var argumentResult =
-                commandResult.Children
+                commandResult!.Children
```
jcouv, line 84: consider assertions for program invariants, as opposed to checks in the current method body which the compiler didn't or couldn't understand.

jcouv:
>             if (args == null)

make `args` nullable?

---
Refers to: src/System.CommandLine/Parsing/StringExtensions.cs:312 in 74b3676. [](commit_id = 74b3676fd5f71169b25bb62ca32acb59c16d23e9, deletion_comment = False)

src/System.CommandLine/IDirectiveCollection.cs
```diff
@@ -2,13 +2,14 @@
...
+using System.Diagnostics.CodeAnalysis;
 
 namespace System.CommandLine
 {
     public interface IDirectiveCollection : IEnumerable<KeyValuePair<string, IEnumerable<string>>>
     {
         bool Contains(string name);
 
-        bool TryGetValues(string name, out IEnumerable<string> values);
+        bool TryGetValues(string name,  [NotNullWhen(true)] out IEnumerable<string>? values);
```
jcouv, line 13: nit: inconsistent formatting of attribute here and elsewhere. Not sure what is the guideline for this repo. For Roslyn, we're moving to having a space after `]`.

src/System.CommandLine/Binding/ServiceProviderValueSource.cs
```diff
@@ -7,10 +7,10 @@ internal class ServiceProviderValueSource : IValueSource
     {
         public bool TryGetValue(
             IValueDescriptor valueDescriptor,
-            BindingContext bindingContext,
-            out object boundValue)
+            BindingContext? bindingContext,
+            out object? boundValue)
         {
-            boundValue = bindingContext.ServiceProvider.GetService(valueDescriptor.ValueType);
+            boundValue = bindingContext?.ServiceProvider.GetService(valueDescriptor.ValueType);
```
jcouv, line 13: I don't know about this repo much. Should logic changes come with a corresponding test?

src/System.CommandLine/Parsing/ArgumentResultExtensions.cs
```diff
@@ -2,14 +2,16 @@
...
 namespace System.CommandLine.Parsing
 {
     public static class ArgumentResultExtensions
     {
-        public static object GetValueOrDefault(this ArgumentResult argumentResult) => 
-            argumentResult.GetValueOrDefault<object>();
+        public static object? GetValueOrDefault(this ArgumentResult argumentResult) => 
+            argumentResult.GetValueOrDefault<object?>();
 
+        [return:MaybeNull]
```
jcouv, line 14: nit: formatting `[return: MaybeNull]`

src/System.CommandLine/Binding/ModelBinder.cs
```diff
@@ -250,7 +251,7 @@ public void UpdateInstance<T>(T instance, BindingContext bindingContext)
                 return valueSource;
             }
 
-            if (bindingContext.TryGetValueSource(valueDescriptor, out valueSource))
+            if (bindingContext.TryGetValueSource(valueDescriptor, out valueSource!))
```
jcouv, line 254: Consider changing the `var` above to be an explicit type with nullable annotation instead of using a suppression here.
In 16.6, `var` will automatically be nullable and this problem should go away.

I'm actually surprised there is no warning on `TryGetValue` above... If the `var` is inferred to be not nullable, then there should be a warning for one of the two return cases.

src/System.CommandLine/Parsing/ParseResultVisitor.cs
Keboo: @jcouv Just to make sure I understand. You are suggesting that it is better to leverage Debug.Assert here rather than the null forgiveness operator? Similar to your answer [here](https://stackoverflow.com/a/59701577/3755169)

jcouv: Yes, that's what I'm suggesting.

Keboo: Perhaps I am doing something wrong here, but when I place `Debug.Assert(commandResult != null)` directly above this line and remove the null forgiveness operator, I still get the warning. Is this because the project is targeting netstandard2.0 (meaning the annotations that were added on Debug.Assert are not there)?

src/System.CommandLine/Argument.cs
```diff
@@ -25,7 +26,7 @@ public Argument(string? name)
         {
             if (!string.IsNullOrWhiteSpace(name))
             {
-                Name = name;
+                Name = name!;
```
jcouv, line 29: nit: Consider using an annotated helper method instead

src/System.CommandLine/ArgumentArity.cs
```diff
@@ -47,7 +47,7 @@ public ArgumentArity(int minimumNumberOfValues, int maximumNumberOfValues)
 
             if (tokenCount < minimumNumberOfValues)
             {
-                if (symbolResult.UseDefaultValueFor(argument))
+                if (symbolResult!.UseDefaultValueFor(argument))
```
jcouv, line 50: Didn’t follow. If symbolResult is null, then argumentResult is null, then tokenCount is zero. Wouldn’t we get here with null symbolResult?

src/System.CommandLine/Argument{T}.cs
```diff
@@ -7,19 +7,19 @@ namespace System.CommandLine
 {
     public class Argument<T> : Argument
     {
-        public Argument(string name) : this()
+        public Argument() : base(null)
```
jcouv, line 10: Should this be `base()`?

src/System.CommandLine/Argument{T}.cs
```diff
@@ -68,7 +68,7 @@ public Argument(Func<T> getDefaultValue) : this()
                 }
                 else
                 {
-                    value = default(T);
+                    value = default(T)!;
```
jcouv, line 71: This depends on the specific version of the compiler used. This suppression should no longer be necessary (as of Roslyn 3.6 I suspect, could double-check)

src/System.CommandLine/Binding/ArgumentConverter.cs
```diff
@@ -262,7 +263,7 @@ public static bool CanBeBoundFromScalarValue(this Type type)
             }
             else
             {
-                info = (null, null);
+                info = null;
```
jcouv, line 266: nit: not sure if it matters, but have you considered returning two `out` parameters, to avoid boxing tuple?

src/System.CommandLine/Binding/ArgumentConverter.cs
```diff
@@ -310,21 +311,22 @@ public static bool CanBeBoundFromScalarValue(this Type type)
...
+            result.GetValueOrDefault<object?>();
 
+        [return: MaybeNull]
         internal static T GetValueOrDefault<T>(this ArgumentConversionResult result)
         {
             switch (result)
             {
                 case SuccessfulArgumentConversionResult successful:
-                    return (T)successful.Value;
+                    return (T)successful.Value!;
```
jcouv, line 323: This should also no longer be needed in newer version of Roslyn (3.6, I think). May be worth a follow-up issue?

src/System.CommandLine/Binding/SpecificSymbolValueSource.cs
```diff
@@ -16,23 +16,23 @@ public SpecificSymbolValueSource(IValueDescriptor valueDescriptor)
         public IValueDescriptor ValueDescriptor { get; }
 
         public bool TryGetValue(IValueDescriptor valueDescriptor,
-            BindingContext bindingContext,
-            out object boundValue)
+            BindingContext? bindingContext,
```
jcouv, line 19: Pausing my review here. Will resume on Monday on a proper computer (tablet is challenging) ;-)

src/System.CommandLine/Parsing/ParseResultVisitor.cs
jonsequitur: @jcouv I typically avoid `Debug.Assert` because it will cause the entire test run to fail in CI with no actionable feedback. Thoughts?

src/System.CommandLine/CommandLineConfiguration.cs
```diff
@@ -13,21 +13,21 @@ namespace System.CommandLine
 {
     public class CommandLineConfiguration
     {
-        private IReadOnlyCollection<InvocationMiddleware> _middlewarePipeline;
-        private Func<BindingContext, IHelpBuilder> _helpBuilderFactory;
+        private IReadOnlyCollection<InvocationMiddleware>? _middlewarePipeline;
```
jcouv, line 16: nit: Consider moving the defaulting logic into the constructor instead, so the field never has to hold null. Same for `_helpBuilderFactory`

src/System.CommandLine/EnumerableExtensions.cs
```diff
@@ -33,10 +33,12 @@ internal static class EnumerableExtensions
         }
 
         internal static IEnumerable<T> RecurseWhileNotNull<T>(
-            this T source,
-            Func<T, T> next)
+            this T? source,
+            Func<T, T?> next)
             where T : class
         {
+            if (source is null) yield break;
```
jcouv, line 40: nit: method could probably be simplified:
```
while (source != null)
{
  yield return source;
  source = next(source);
}
```

src/System.CommandLine/Binding/BindingContext.cs
```diff
@@ -75,15 +76,15 @@ public void AddService<T>(Func<IServiceProvider, T> factory)
...
+            [MaybeNullWhen(false)] out IValueSource valueSource)
         {
             if (ServiceProvider.AvailableServiceTypes.Contains(valueDescriptor.ValueType))
             {
                 valueSource = new ServiceProviderValueSource();
                 return true;
             }
 
-            valueSource = default;
+            valueSource = default!;
```
jcouv, line 87: nit: Newer Roslyn should also help avoid this suppression

src/System.CommandLine/Binding/ExpressionExtensions.cs
```diff
@@ -7,30 +7,26 @@ internal static class ExpressionExtensions
...
 
             if (expression.Body is MemberExpression memberExpression)
             {
-                return TypeAndName();
+                return TypeAndName(memberExpression);
             }
 
             // when the return type of the expression is a value type, it contains a call to Convert, resulting in boxing, so we get a UnaryExpression instead
-            if (expression.Body is UnaryExpression unaryExpression)
+            if (expression.Body is UnaryExpression unaryExpression &&
```
jcouv, line 21: nit: could go one step further with recursive patterns: 
`if (expression.Body is UnaryExpression { Operand: MemberExpression operandMemberExpression } _)`

src/System.CommandLine/Invocation/ModelBindingCommandHandler.cs
```diff
@@ -11,39 +11,39 @@ namespace System.CommandLine.Invocation
...
     internal class ModelBindingCommandHandler : ICommandHandler
     {
-        private readonly Delegate _handlerDelegate;
-        private readonly object _invocationTarget;
-        private readonly ModelBinder _invocationTargetBinder;
-        private readonly MethodInfo _handlerMethodInfo;
+        private readonly Delegate? _handlerDelegate;
+        private readonly object? _invocationTarget;
+        private readonly ModelBinder? _invocationTargetBinder;
+        private readonly MethodInfo? _handlerMethodInfo;
```
jcouv, line 17: could this field be marked non-nullable? (same question for `_parameterDescriptors`)

src/System.CommandLine/Invocation/ModelBindingCommandHandler.cs
```diff
@@ -63,9 +63,9 @@ public async Task<int> InvokeAsync(InvocationContext context)
                                    _invocationTargetBinder?.CreateInstance(bindingContext);
 
             object result;
-            if (_handlerDelegate == null)
+            if (_handlerDelegate is null)
             {
-                result = _handlerMethodInfo.Invoke(
+                result = _handlerMethodInfo!.Invoke(
```
jcouv, line 68: how do we know this is not null?

src/System.CommandLine/Parsing/Parser.cs
```diff
@@ -42,7 +42,7 @@ public Parser() : this(new RootCommand())
                 operation.Errors,
                 rawInput);
 
-            visitor.Visit(operation.RootCommandNode);
+            visitor.Visit(operation.RootCommandNode!);
```
jcouv, line 45: nit: consider assertion instead

src/System.CommandLine/Parsing/ParseResultVisitor.cs
jcouv: @jonsequitur Ouch. That would make assertions painful indeed :( 
We use a lot of assertions in Roslyn and I never noticed it failing entire test suits. Is this behavior new with Core vs. desktop? Is it the same with a NullRefException?

The nullability feature relies a significantly on `Debug.Assert` to give the compiler hints, so it might be worthwhile to make it behave in tests (ie. fail a single test rather than the entire suite).

jcouv: I've been thinking about this problem. Maybe an option is to add your own assertion method with the `[DoesNotReturnIf(false)]` annotation, but don't have it actually assert anything until the CI issue could be resolved (exceptions not being reported on individual test that failed). 
That one helper method could be marked with follow-up issue and eventually be made into a bonafide assertion.

jonsequitur: On .NET Core it crashes the process and leaves an empty `.trx` file (if you're using `--logger trx`). If someone happens to be relying on `.trx` file output instead of the test runner result code, this can lead to an incorrect pass on the test run. On .NET Framework, it pops a modal dialog and hangs the build.

These behaviors can be customized by removing the `DefaultTraceListener` and, for example, replacing it with a `TraceListener` that throws an exception, but this is kind of a non-obvious ritual.

PR #874, opened 2020-04-22
Modifications to support external middleware

markolbert:
I am proposing these two changes (ignoring the project file tweak) to support an external middleware library I wrote which I believe simplifies binding the results of System.CommandLine processing to a configuration object.

Changing the access rights on HelpOption is necessary because I need to be able to detect when help has been requested and the only way I can see to do that is to search ParseResult for the presence of the HelpOption object defined in the CommandLineBuilder.

Changing the access rights on TokenizeResult() is needed so I can present only those command line arguments that relate to a particular Command to the parsing routine I wrote.

If you'd like to see the library I wrote in action check out https://github.com/markolbert/InvestmentSimulator.


jonsequitur:
@markolbert If you're still interested in pursuing this PR, it would be very helpful to have one or two concrete examples (e.g. as tests) of what scenarios this change enables.

markolbert:
Hi Jon,

You can reject/close/whatever the PR. As you may recall I wrote my own command line parsing library which I find works with the way I tend to design command line apps better than CommandLineApi. If you're curious you can check it out at https://github.com/markolbert/J4JCommandLine.

jonsequitur:
Thanks, I'll close it.

I took a look at J4JCommandLine but I haven't had time to give it a try. Very cool to see more people working in this space!

PR #889, opened 2020-05-06
Extend Symbol with IEquatable<ISymbol>

BobSilent:


jonsequitur:
For design changes as impactul as this, it's better to open an issue first so we can discuss the goals.

I believe this change would break a currently-supported feature, which is that you can have different options with the same name that are actually different options. For example:

```console
> myapp one -x
> myapp two -x
```

This could be defined like this:

```cs
new RootCommand
{
  new Command("one")
  {
    new Option<string>("-x")
  },
  new Command("two")
  {
    new Option<int>("-x")
  }
}
```

Issue #826, opened 2020-03-22
Allow a simple implementation of enabling dependency injection


taori:
These days it would be nice if there was a simple way of utilizing dependency injection in command line applications as well.

Currently i am using the latest version, using middleware to populate the bindingcontext with services, but that does not allow open generic registration (for easy logger registration). Perhaps that is a scenario worth keeping in mind.

jonsequitur:
The way you're doing it is a simple approach that lets you use a dependency injection library of your choice but was intended for the 80% cases. 

For more complex cases, Microsoft.Extensions.DependencyInjection is supported via the [System.CommandLine.Hosting](https://www.nuget.org/packages/System.CommandLine.Hosting/) package. Would this work for you?

taori:
> 
> 
> The way you're doing it is a simple approach that lets you use a dependency injection library of your choice but was intended for the 80% cases.
> 
> For more complex cases, Microsoft.Extensions.DependencyInjection is supported via the [System.CommandLine.Hosting](https://www.nuget.org/packages/System.CommandLine.Hosting/) package. Would this work for you?

That sounds very promising actually. Somehow i only stumbled upon DragonFruit and dotnet-suggest through the wiki of this project. I'll have a look at it and get back to this issue with feedback

taori:
This certainly seems to be what i am looking for. What i plan on doing is having an invocation structure where i implement something like `ITypedCommand<TArg1, TArg2, ...>` so i can utilize constructor injection and map arguments through Attributes. Do you think this is doable right now? Essentially i need something similar to how things work with mvc actions.

taori:
@jonsequitur This is my current state.

I can't seem to get it working unfortunately. At least i would expect it to forward input to the "testCommand" handler, but not even that appears to work. 

Do you think what i am attempting to do is doable with *.Hosting?
```cs
using System;
using System.CommandLine;
using System.CommandLine.Builder;
using System.CommandLine.Hosting;
using System.CommandLine.Invocation;
using System.CommandLine.Parsing;
using System.Reflection;
using System.Threading.Tasks;
using Generator.CLI.Component.Parser;
using Generator.CLI.Dependencies;
using Microsoft.Extensions.DependencyInjection;
using Microsoft.Extensions.Hosting;
using Microsoft.Extensions.Logging;

namespace Generator.CLI
{
	class Program
	{
		static async Task<int> Main(string[] args)
		{
			var builder = new CommandLineBuilder()
[...]

PR #890, opened 2020-05-06
Hosting simplifications when using Dependency Injections

BobSilent:
This PR is not in a mergable state, but to visualize a simplified handling with dependency injections.

it would allow to write the following code
```csharp
    class SampleProgram
    {
        {
            var appBuilder = new HostedCommandLineBuilder()
                             .AddOption(new Option<bool>("--isOption"))
                             .AddOption(new Option<string>("--value"));

            appBuilder.Command.Handler = appBuilder.BuildHandler<SampleCommand>();


            var parser = appBuilder.UseDefaults()
                                   .UseHost(host =>
                                   {
                                       host.ConfigureServices((ctx, services) =>
                                       {
                                           services.AddCommands(appBuilder);
                                           services.AddOptions<SampleOption>()
                                                   .BindCommandLine();
                                       });
                                   })
                                   .Build();

            return await parser.InvokeAsync(args);
        }
    }
```

for completeness
```csharp
    internal class SampleCommand : ICommand
    {
        private readonly ILogger _logger;
        private readonly SampleOption _options;
        private readonly IConsole _console;

        public SampleCommand(ILogger<SampleCommand> logger, IOptions<SampleOption> options, IConsole console)
        {
            _logger = logger;
            _options = options.Value;
            _console = console;
        }

        public async Task RunAsync(CancellationToken cancellationToken = default)
        { /* Execute Command */ }
    }

    internal class SampleOption
    {
        public bool IsOption { get; set; }
        public string Value { get; set; }
    }
```

### DIFF
--- a/src/System.CommandLine.Hosting/HostedCommandLineBuilder.cs
+++ b/src/System.CommandLine.Hosting/HostedCommandLineBuilder.cs
@@ -1,0 +1,62 @@
+﻿using System.Collections.Generic;
+using System.CommandLine.Builder;
+using System.CommandLine.Invocation;
+using System.Threading;
+using System.Threading.Tasks;
+
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+
+
+namespace System.CommandLine.Hosting
+{
+    public interface ICommand
+    {
+        Task RunAsync(CancellationToken cancellationToken = default);
+    }
+
+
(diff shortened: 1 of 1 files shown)


jonsequitur:
Can you explain more about the goal of the PR? Does this relate to #826?

I'd like to discuss the goal and then first explore whether it can be done without introducing new API surface area, so it will "just work".

Issue #12, opened 2018-05-02
Handling suggestion sort order?


MarkMichaelis:
If the provider calls `Define.Arguments().FromAmong("high","medium").FromAmong("low")` should we sort the suggestion list alphabetically or in the order that configured?

kimsey0:
Currently, the default implementation of `GetSuggestions` in `Symbol` [orders child symbol aliases alphabetically](https://github.com/dotnet/command-line-api/blob/afd010ba8cb3cbd714c734465d1a5de1ee133e2d/src/System.CommandLine/Symbol.cs#L149). Could it make sense to change that to order first by position at which the text to match was matched, then alphabetically? I.e., changing:

``` cs
.OrderBy(symbol => symbol, StringComparer.OrdinalIgnoreCase)
.Containing(textToMatch);
```

to

``` cs
.Containing(textToMatch)
.OrderBy(symbol => symbol.IndexOfCaseInsensitive(textToMatch))
.ThenBy(symbol => symbol, StringComparer.OrdinalIgnoreCase);
```

For a concrete tool I'm working on, this would change the suggestions order when typing `my-tool me<tab>` from
> fix-docker-desktop-ti**me**
> [meta](https://github.com/mateodelnorte/meta)-pull

to the, to me, more natural:

> **me**ta-pull
> fix-docker-desktop-time

PR #891, opened 2020-05-07
Sort suggestions by match index, then alphabetically

kimsey0:
Implements suggestion from https://github.com/dotnet/command-line-api/issues/12#issuecomment-625229936

### DIFF
--- a/src/System.CommandLine.Tests/SuggestionTests.cs
+++ b/src/System.CommandLine.Tests/SuggestionTests.cs
@@ -5,6 +5,7 @@ using System.Collections.Generic;
 using System.CommandLine.Builder;
 using System.CommandLine.Invocation;
 using System.CommandLine.Parsing;
+using System.CommandLine.Tests.Utility;
 using System.IO;
 using FluentAssertions;
 using Xunit;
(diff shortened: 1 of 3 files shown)


jonsequitur:
This is a nice solution to #12. Thanks, @kimsey0!

Could you add a unit test that demonstrates the behavior in `SuggestionTests`? (The tests there currently avoid asserting on the order of the suggestions (e.g. by using `BeEquivalentTo`), since this wasn't locked down. We have a `BeEquivalentSequenceTo` extension method which is a convenience for strict ordering.)

kimsey0:
Sure! Would you like the existing tests to be changed to `BeEquivalentSequenceTo` too, or just some new tests?

jonsequitur:
It's good to just add a test. The existing tests are intentionally unconcerned with suggestion order.

kimsey0:
I've added two tests of suggestions for subcommands, one with text to match and one without. Is this sufficient? I was a bit unsure about the naming, since `SuggestionTests` seems to contain both `Noun_Sentence` and `Sentence` style test names.

jonsequitur:
This looks great. Thanks, @kimsey0!

kimsey0:
Thank you!

PR #895, opened 2020-05-09
Use the ordinal ignore case string comparer for ordering option aliases

### LINKED ISSUE #894: Option alias ordering is case sensitive
When running the `Option_aliases_are_shown_before_long_names_regardless_of_alphabetical_order` test from `HelpBuilderTests` on a machine with `da-DK` locale, it fails with the message:

```
Expected string "Usage:
  testhost [options]

Options:
  -a, -z, --zzz, --aaa    

" to contain "-a, -z, --aaa, --zzz".
```

This seems to be due to fact that the [ordering by alias](https://github.com/dotnet/command-line-api/blob/ca5057da1d0a272ef8f31ec651d7aeff99b48232/src/System.CommandLine/Help/HelpBuilder.cs#L339) in `HelpBuilder` doesn't specify a culture setting, defaulting to the current culture, and that `aa` is interpreted as a ligature for `å` in Danish.

kimsey0:
Fixes #894. I've specified the ordinal ignore case string comparer, since it seems to be the norm in this project to order case insensitively. Let me know if it should actually be `StringComparer.Ordinal`.

### DIFF
--- a/src/System.CommandLine/Help/HelpBuilder.cs
+++ b/src/System.CommandLine/Help/HelpBuilder.cs
@@ -335,8 +335,8 @@ namespace System.CommandLine.Help
             var rawAliases = symbol
                              .RawAliases
                              .Select(r => r.SplitPrefix())
-                             .OrderBy(r => r.prefix)
-                             .ThenBy(r => r.alias)
+                             .OrderBy(r => r.prefix, StringComparer.OrdinalIgnoreCase)
+                             .ThenBy(r => r.alias, StringComparer.OrdinalIgnoreCase)
                              .GroupBy(t => t.alias)
                              .Select(t => t.First())
                              .Select(t => $"{t.prefix}{t.alias}");


jonsequitur:
Thanks, @kimsey0!

kimsey0:
How often are new beta releases pushed to nuget.org?

jonsequitur:
The process is manual, depending on the impact and stability of recent changes. Every merged PR however produces signed packages that you can get from this feed: https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet5/nuget/v3/index.json

kimsey0:
OK. I'd like to use this change and #891 in a project with others who I think would prefer not changing to the nightly build feed. I'll just continue looking forward to this releasing in the nuget.org feed. 🙂

jonsequitur:
I'll post an update today then.

kimsey0:
Thanks a lot. I appreciate it.

jonsequitur:
You're welcome!

Issue #619, opened 2019-08-13
Display default values when running help command


Marcel-Lambacher:
I would like to see the defined default values when I request the parameter help command.
If this is not possible by default, I would be really interested to implemented it accordingly.

**Given** is the following strongly typed main definition:
```cs
static void Main(int intOption = 42, bool boolOption = false, FileInfo fileOption)
{
    Console.WriteLine($"The value of intOption is: {intOption}");
    Console.WriteLine($"The value of boolOption is: {boolOption}");
    Console.WriteLine($"The value of fileOption is: {fileOption?.FullName ?? "null"}");
}
```

**When** running the help command:
`> ./myapp -h # or: dotnet run -- -h`

**Then** the default values should be displayed in the arguments documentation:
```
> dotnet run -- -h
Usage:
  myapp [options]

Options:
  --int-option  (Default value: 42)        An option whose argument will bind to an int
  --bool-option (Default value: false)     An option whose argument will bind to a bool
  --file-option                            An option whose argument will bind to a FileInfo
```

Marcel-Lambacher:
Okay, as the labels has been added I'll work on this task, thanks :)

CaptainGinja:
I also would like this feature.  Are there any plans to integrate this functionality into the master branch?

NikiforovAll:
I like this feature and I would like to deliver this feature. Unfortunately, @Marcel-Lambacher hasn't done PR for quite some time.

jonsequitur:
I'm happy to help anyone who would like to put together a PR for this.

PR #898, opened 2020-05-11
Add default values to HelpBuilder output

NikiforovAll:
Fixes: #619

And output deviates from the initially discussed format in favour of a format that resembles  [docopt](http://docopt.org/) style. Please see the added tests for more details.


src/System.CommandLine.Tests/Help/HelpBuilderTests.cs
```diff
@@ -842,6 +842,55 @@ public void Option_argument_descriptor_indicates_enums_values(Type type)
             _console.Out.ToString().Should().Contain($"--opt <Read|ReadWrite|Write>{_columnPadding}{description}");
         }
 
+        [Fact]
+        public void Command_argument_default_value_provided()
```
jonsequitur, line 846: I'd suggest renaming these tests to clarify the scenario and include the expectation, e.g. "Help describes default values for command arguments"

jonsequitur:
Thanks, @NikiforovAll! Aligning to docopt's format makes good sense.

It occurs to me while looking at the code that it would be helpful to have an example of the complete test output. [ApprovalTests](https://www.nuget.org/packages/ApprovalTests/) provides a great way to do this by capturing output into a file that serves as a testable artifact in the repo. Would you mind adding a test using this approach for a single root command help, with a good variety of features in use (give the root command a custom name, options and arguments with and without default values, required and non-required options, and an enum so we can see the values list)?

/cc @KathleenDollard

src/System.CommandLine.Tests/Help/HelpBuilderTests.cs
NikiforovAll: So, "Command_argument_default_value_provided" will be renamed to "Help_desribes_default_value_for_single_command_argument"

NikiforovAll:
Yes, sure thing.

src/System.CommandLine.Tests/Help/HelpBuilderTests.cs
jonsequitur: That looks perfect. 

The naming that we tend toward in this project describes the behavior in plain language, favors clarity over conciseness, and avoids focusing on type names and implementation details. This helps the tests function as a spec and prevents us from having to rename tests unless we're changing a behavior (as opposed to refactoring).

NikiforovAll:
@jonsequitur As I understand, there is no support for [ApprovalTests](https://www.nuget.org/packages/ApprovalTests/) at the moment. Do you want me to create another issue + PR to add support for _ApprovalTests_ to keep things nice and clean?

jonsequitur:
Adding commits to the same PR is fine.

NikiforovAll:
_Azure Pipelines_ build has failed with `UnauthorizedAccessException` because ApprovalTests tries to read _.approved.txt_ and doesn't have access to it.

@jonsequitur Could you please assists with this? 🙂 

---

Tests are green on my machine. ✔

src/System.CommandLine.Tests/ApprovalTests/Help/HelpBuilderTests.Approval.cs
```diff
@@ -0,0 +1,78 @@
...
+                    Description = "the-root-arg-enum-default-description",
+                    ArgumentType = typeof(FileAccess)
+                },
+                new Option(aliases: new string[] {"--the-root-option-no-arg", "-trna"}) {
+                    Description = "the-root-option-no-arg-description",
+                    Required = true
+                },
+                new Option(aliases: new string[] {"--the-root-option-no-default-arg", "-tronda"}) {
+                    Description = "the-root-option-no-default-description",
+                    Argument = new Argument<string>("the-root-option-arg-no-default-arg")
```
jonsequitur, line 38: For these we should prefer `Option<T>` over explicitly setting the `Argument` property.

src/System.CommandLine.Tests/ApprovalTests/Help/HelpBuilderTests.Approval.cs
```diff
@@ -0,0 +1,78 @@
...
+                    Description = "the-root-arg-no-default-description",
+                },
+                new Argument<string>("the-root-arg", () => "the-root-arg-one-value")
+                {
+                    Description = "the-root-arg-description"
+                },
+                new Argument<FileAccess>("the-root-arg-enum-default", () => FileAccess.Read)
+                {
+                    Description = "the-root-arg-enum-default-description",
+                    ArgumentType = typeof(FileAccess)
```
jonsequitur, line 30: Setting the `ArgumentType` property should be unnecessary.

src/System.CommandLine.Tests/ApprovalTests/Help/HelpBuilderTests.Approval.cs
```diff
@@ -0,0 +1,78 @@
...
+namespace System.CommandLine.Tests.Help
+{
+    public partial class HelpBuilderTests
+    {
+        [Fact]
+        public void Help_describes_default_values_for_complex_root_command_scenario()
+        {
+            var command = new RootCommand(description: "Test description")
+            {
+                new Argument<string>("the-root-arg-no-default")
```
jonsequitur, line 19: It might be useful to include an argument with no description set.

src/System.CommandLine.Tests/ApprovalTests/Help/Approvals/HelpBuilderTests.Help_describes_default_values_for_complex_root_command_scenario.approved.txt
```diff
@@ -0,0 +1,18 @@
...
+  <the-root-arg>               the-root-arg-description [default: the-root-arg-one-value]
+  <Read|ReadWrite|Write>       the-root-arg-enum-default-description [default: Read]
+
+Options:
+  -trna, --the-root-option-no-arg (REQUIRED)                                                   the-root-option-no-arg-description
+  -tronda, --the-root-option-no-default-arg <the-root-option-arg-no-default-arg> (REQUIRED)    the-root-option-no-default-description
+  -troda, --the-root-option-default-arg <the-root-option-arg>                                  the-root-option-default-arg-description [default: the-root-option-arg-value]
+  -troea, --the-root-option-enum-arg <Read|ReadWrite|Write>                                    the-root-option-description [default: Read]
+  -trorea, --the-root-option-required-enum-arg <Read|ReadWrite|Write> (REQUIRED)               the-root-option-description [default: Read]
+
```
jonsequitur, line 18: This is super helpful.

src/System.CommandLine.Tests/ApprovalTests/Help/HelpBuilderTests.Approval.cs
NikiforovAll: I've found a suitable way to create an option that takes an argument with description. I will add the case you mentioned above as a separate option example. (trondda)

```csharp
new Option<string>(
    aliases: new string[] {"--the-root-option-no-description-default-arg", "-trondda"}, 
    parseArgument: _ => "the-root-option--no-description-default-arg-value",
    isDefault: true
),
```

NikiforovAll:
I've tried to figure this out, but no luck. It looks like in some cases content of .approved.txt is empty and for others, the file could not be accessed.
<summary>
Windows_NT_Windows Release - System.CommandLine.Tests_netcoreapp3.1_x64.log
<details>
      Assert.Equal() Failure
                 (pos 0)
      Expected: 
      Actual:   the-root-command:\r\n  Test description\r\n\r\nъъъ
                 (pos 0)
</details>
</summary>
<summary>
Linux_Ubuntu 16.04 Debug - System.CommandLine.Tests_netcoreapp3.1_x64.log
<details>
System.UnauthorizedAccessException : Access to the path '/_/src/System.CommandLine.Tests/ApprovalTests/Help/Approvals' is denied.
--- System.IO.IOException : Permission denied
</details>
</summary>

jonsequitur:
I haven't seen this kind of issue before. I won't get a chance to dig into it today but a few experiments I would try:

* Put the approvals files in the same directory as the `.cs` files instead of a subdirectory.
* Add log output for a few file operations (e.g. `File.Exists`, `File.ReadAllText`). The exception might be misleading.
* The `Assent` library is an equivalent library that's based on `ApprovalTests` but has been .NET Core compatible for longer and I've been using it for a couple years without issues, so maybe try that one instead and see if it's more successful?

Apologies for the complications. I can try these investigations myself in a couple of days if you don't have the time.

Thanks.

NikiforovAll:
Turns out _ApprovalTests.Net_ doesn't work in the case of deterministic assemblies. So the workaround is `<DeterministicSourcePaths>false</DeterministicSourcePaths>`. I'm not sure if it is the way to go with this. LMKWYT

--- 
StructuredLogViewer query: `$property DeterministicSourcePaths under($project System.CommandLIne.Tests)`

The thingy that sets MSBuild property.
```xml
<PropertyGroup>
    <DeterministicSourcePaths Condition="'$(DeterministicSourcePaths)' == '' and '$(Deterministic)' == 'true' and '$(ContinuousIntegrationBuild)' == 'true'">true</DeterministicSourcePaths>
</PropertyGroup>
```

jonsequitur:
Thanks, @NikiforovAll! This is a really nice addition.

Issue #482, opened 2019-04-01
Documentation of use with dependency injection 


clmcgrath:
Does system.commandline support dependency injection via MEDI if so 
examples should be documented on the wiki 

if not can this be implemented ? seems like an important feature

jonsequitur:
Agreed. 

The capability to bind arbitrary values by type is roughed in via the model binding infrastructure ([example](https://github.com/dotnet/command-line-api/blob/4fe9e05bee7a9698408d151a92faabd81ef04184/src/System.CommandLine/Binding/BindingContext.cs#L46)) but the API isn't exposed well yet.

We're also investigating integration with GenericHost: #440.

TristanApc:
I have the same request. Considering that the integration with GenericHost is closed, was there any progress on the documentation for DI?

osoykan:
Hi, I have written a [blog post](https://oguzhansoykan.com/posts/2020/2020-04-15-netcore-cli/) about how you use Dependency Injection, which also includes Aspnet.Core. I hope it helps.

PR #912, opened 2020-05-28
Add GenericHost playground sample 

NikiforovAll:
Looking through PRs I found that quite a lot of folks ask for documentation and samples on how to use `System.CommandLine.Hosting`. Also, I think it useful to have a playground for _Hosting_ since it allows people to play with it and share user feedback. Therefore this PR.

Fixes: #482
Relates: #671, #826, #440

### DIFF
--- a/samples/HostingPlayground/GreeterOptions.cs
+++ b/samples/HostingPlayground/GreeterOptions.cs
@@ -1,0 +1,12 @@
+namespace HostingPlayground
+{
+    public class GreeterOptions
+    {
+        public string Name { get; set; }
+
+        public GreeterOptions(string name)
+        {
+            Name = name;
+        }
+    }
+}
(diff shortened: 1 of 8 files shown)


samples/HostingPlayground/HostingPlayground.csproj
```diff
@@ -0,0 +1,23 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <LangVersion>latest</LangVersion>
```
Keboo, line 7: Since `latest` will be the latest installed on the running computer, I recommend setting this to 8 to match the other projects.

samples/HostingPlayground/GreeterOptions.cs
```diff
@@ -0,0 +1,12 @@
+namespace HostingPlayground
+{
+    public class GreeterOptions
+    {
+        public string Name { get; set; }
```
jonsequitur, line 5: If this is intended to be called by the model binder, then you don't need the setter.

PR #919, opened 2020-06-06
Add generic host builder pattern [wip]

### LINKED ISSUE #918: Add generic host builder pattern for System.CommandLine.Hosting
Align `System.CommadLine.Hosting` model with the way ASP.NET Core uses _Generic Host_.

It should be possible to configure CommandLineBuilder like this: 
```csharp
public static IHostBuilder CreateHostBuilder(string[] args) =>
    CommandLineHost.CreateDefaultBuilder(args)
        .ConfigureAppConfiguration((hostingContext, config) =>
        {
        })
        .ConfigureCommandLineDefaults(cmdHost =>
        {
            cmdHost.ConfigureServices(services =>
            {
            })
        })
        .ConfigureCommandLineBuilder(cmdBuilder => // CommandLineBuilder
        {
           cmdBuilder.UseReflectionAppModel(); // e.g. https://github.com/KathleenDollard/command-line-api-starfruit 
        });
    });
```

NikiforovAll:
Hi @jonsequitur,
Here is my take on it. What do you think? Is it the right direction?

Closes #918

### DIFF
--- a/src/System.CommandLine.Hosting.Tests/HostingTests.cs
+++ b/src/System.CommandLine.Hosting.Tests/HostingTests.cs
@@ -3,7 +3,8 @@ using System.CommandLine.Builder;
 using System.CommandLine.Invocation;
 using System.CommandLine.Parsing;
 using System.Linq;
-
+using System.Threading;
+using System.Threading.Tasks;
 using FluentAssertions;
 
 using Microsoft.Extensions.Configuration;
@@ -226,6 +227,37 @@ namespace System.CommandLine.Hosting.Tests
             Assert.Equal(0, result);
             Assert.NotNull(options);
             Assert.Equal(myValue, options.MyArgument);
+        }
+
+
+        [Fact]
+        public static async Task CommandLineHost_default()
+        {
+            //Arrange
+            // var args = new string[] { $"--foo", "42" };
+            // MyOptions options = null;
+            IHost hostToBind = null;
+
+            var rootCmd = new RootCommand();
+            rootCmd.AddOption(new Option($"--foo") { Argument = new Argument<int>() });
+            rootCmd.Handler = CommandHandler.Create<IHost>((host) =>
+            {
+                hostToBind = host;
+            });
+            // Act
+            // var tokenSource = new CancellationTokenSource(TimeSpan.FromSeconds(1));
+            CancellationTokenSource tokenSource = null;
+            await CommandLineHost.CreateDefaultBuilder()
+                .ConfigureCommandLineDefaults((CommandLineBuilder builder) =>
+                {
+                    // TODO: it is not possible to add it like this atm.
+                    builder.AddCommand(rootCmd);
+                })
+                .Build()
+                .RunAsync(tokenSource?.Token ?? default);
+            // Assert
+            // Assert.NotNull(hostToBind);
+            // Assert.Equal(42, options.MyArgument);
         }
 
         private class MyOptions
--- a/src/System.CommandLine.Hosting.Tests/System.CommandLine.Hosting.Tests.csproj
+++ b/src/System.CommandLine.Hosting.Tests/System.CommandLine.Hosting.Tests.csproj
@@ -15,7 +15,7 @@
 
   <ItemGroup>
     <PackageReference Include="FluentAssertions" Version="5.10.3" />
-    <PackageReference Include="Microsoft.Extensions.Configuration.CommandLine" Version="2.2.0" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.CommandLine" Version="3.1.4" />
   </ItemGroup>
 
   <ItemGroup Condition="'$(DisableArcade)' == '1'">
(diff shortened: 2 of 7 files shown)


NikiforovAll:
Currently, I'm looking to the recommended way to add `RootCommand` to already instantiated `CommandLineBuilder`. 🤔 

```csharp
CommandLineHost.CreateDefaultBuilder()
    .ConfigureCommandLineDefaults((CommandLineBuilder builder) =>
    {
        builder.AddCommand(rootCmd);
    })
```

KathleenDollard:
There are some other issues I'd like to look at later, but the problematic code is: 

```
       public static async Task CommandLineHost_creates_host_for_simple_command()
        {
...
            await CommandLineHost.CreateDefaultBuilder()
                .ConfigureCommandLineDefaults((CommandLineBuilder builder) =>
                {
                    // TODO: it is not possible to add it like this atm.
                    builder.AddCommand(rootCmd);
                })
                .Build()
                .RunAsync(tokenSource?.Token ?? default);
```
  
The builder has a command at this point. The philosophy is that the builder always has a command. So, does the following work: 

```
           await CommandLineHost.CreateDefaultBuilder()
                .ConfigureCommandLineDefaults((CommandLineBuilder builder) =>
                {
                    // The following mutates the existing command, rather than add a new one
                    builder.Command.AddOption(myOption);
                })
                .Build()
                .RunAsync(tokenSource?.Token ?? default);
```

This is what I have in mind for StarFruit (with whatever name we land on)

```
           await CommandLineHost.CreateDefaultBuilder()
                .ConfigureCommandLineDefaults((CommandLineBuilder builder) =>
                {
                    // The following mutates the existing command, rather than add a new one
                    builder.Command.ConfigureStarFruit<MyCLIModelRoot>();
                })
                .Build()
                .RunAsync(tokenSource?.Token ?? default);
```

I believe everything _within_ the command is mutable, although replacing the command in the builder is not possible, for reasons that I think remain valid. 

Do you see any issues with this approach?

NikiforovAll:
I think using the mutability of a `Command` is a good thing. Why don't we use this on the builder directly?
`builder.ConfigureStarFruit<MyCLIModelRoot>();`

KathleenDollard:
I am out of time this morning, but this is how far I got on your test: 

* Host is null in the handler for reasons I didn't debug
* The approach you used for testing the options doesn't work

```
      [Fact]
        public static async Task CommandLineHost_creates_host_for_simple_command()
        {
            //Arrange
            // var args = new string[] { $"--foo", "42" };
            MyOptions options = null;
            IHost hostToBind = null;
            var check = 0;

            var option = new Option($"--foo") { Argument = new Argument<int>() };
            var handler = CommandHandler.Create<IHost>((host) =>
           {
               hostToBind = host;
               check = 7;
           });
            // Act
            // var tokenSource = new CancellationTokenSource(TimeSpan.FromSeconds(1));
            CancellationTokenSource tokenSource = null;
            await CommandLineHost.CreateDefaultBuilder()
                .ConfigureCommandLineDefaults((CommandLineBuilder builder) =>
                {
                    builder.Command.AddOption(option);
                    builder.Command.Handler = handler;
                })
                .Build()
                .RunAsync(tokenSource?.Token ?? default);
            // Assert
            //Assert.NotNull(hostToBind);
            //Assert.Equal(42, options.MyArgument);
            Assert.Equal(7, check);
        }
```

NikiforovAll:
@KathleenDollard Sorry for not communicating the status of current PR. Currently, I'm waiting for design and final thoughts in #918 (+ https://github.com/dotnet/command-line-api/issues/918#issuecomment-642070042)

This happens because it is only possible to access `InvocationContext` from `UseMiddleware` method. And the method you see in PR is just a dummy that is not used during actual execution of a handler (ref: `src\System.CommandLine.Hosting\DefaultBuilder\GenericCommandLineHostBuilder.cs:34`)

I stumbled on the cyclic dependency between `CommandLineBuilder` and `IHostBuilder`: 
`CommandLineBuilder` depends on IHost and to build `IHost` we need to provide `ParseResult`, `BindingContext` etc. to IHost . (At least, it is like that at the moment). 

I see two possible solutions:

1. Separate the parsing process and execution process. Expose invocation context as a result of the parsing process and add to DI
```csharp
// step 1
Parser parser = commandLineBuilder.Build();
ParseResult parseResult = parser.Parse(Args);
// access invocation from parseResult
{
    services.TryAddSingleton<InvocationContext>(invocation);
    //...
});
// step 2 inject IHost to CommandLineBuilder right before execution 
// src\System.CommandLine.Hosting\DefaultBuilder\CommandLineExecutorService.cs:36
public async Task StartAsync(CancellationToken cancellationToken)
{
    invocation.BindingContext.AddService(typeof(IHost), _ => host);
    //...
    await parseResult.InvokeAsync();
    appLifetime.StopApplication();
}
```
2. Build `CommandLineBuilder` **one more time** in `HostedService` via `UseMiddleware()`. I think it might introduce a lot of bugs.

PR #942, opened 2020-06-23
Fix output mode will always be PlainText.

ericwj:
Maybe something larger is out of order, but at the very least even `RenderingPlayground` will stick to `PlainText` unless configured with `[output:ansi]`.

With this change, `OutputMode` will remain `Auto` until the first time something is rendered, [here](https://github.com/dotnet/command-line-api/blob/469b7423fa709682c6988513270f14328b9771f9/src/System.CommandLine.Rendering/ConsoleRenderer.cs#L81) at which time the detection will properly determine `Ansi` is OK.

Without this, [UseAnsiTerminalWhenAvailable](https://github.com/dotnet/command-line-api/blob/469b7423fa709682c6988513270f14328b9771f9/src/System.CommandLine.Rendering/CommandLineBuilderExtensions.cs#L13) will have `context.Console` be a `SystemConsole` which is not an `ITerminal` hence will yield `PlainText` from [`DetectOutputMode`](https://github.com/dotnet/command-line-api/blob/469b7423fa709682c6988513270f14328b9771f9/src/System.CommandLine.Rendering/ConsoleExtensions.cs#L33).

### DIFF
--- a/src/System.CommandLine.Rendering/CommandLineBuilderExtensions.cs
+++ b/src/System.CommandLine.Rendering/CommandLineBuilderExtensions.cs
@@ -58,7 +58,7 @@ namespace System.CommandLine.Rendering
                 return mode;
             }
 
-            return context.Console.DetectOutputMode();
+            return Rendering.OutputMode.Auto;
         }
     }
 }


jonsequitur:
Thanks, @ericwj!

Which terminals have you tried this in?

ericwj:
It is part of a solution - this fixes the sample but not my application. My application now gets to detect the mode lazily, but still the `_terminal` is `null` while `IConsole.GetTerminal` extension method neatly returns a `VirtualTerminal`. I think the terminal needs to be created way at the start of anything really such that all initialization can depend on the actual capabilities instead of getting `SystemConsole` by default and saving state to various places before the actual capabilities are determined.

I have tried this in `Visual Studio Debug Console` on Windows 10.0.19041.329 and yeah same happens in Windows Terminal 1.0.1401.0.

jonsequitur:
When you're seeing a null `ITerminal` has output been redirected?

ericwj:
No, without DragonFruit I just had the wrong API. I have to say without source code this would have been fairly impossible to figure out.

Compare the return statements before and after:
```
var a = new Args();
return await a.Root.InvokeAsync(args);

var builder = new CommandLineBuilder(a.Root)
    .UseVersionOption()
    .UseHelp()
    .UseParseDirective()
    .UseDebugDirective()
    .UseSuggestDirective()
    .RegisterWithDotnetSuggest()
    .UseTypoCorrections()
    .UseParseErrorReporting()
    .UseExceptionHandler()
    .CancelOnProcessTermination()
    .UseAnsiTerminalWhenAvailable()
    .Build();
var parser = builder.Parse(args);
return await parser.InvokeAsync();
```
Where `Args` is just where I new up `Option` instances and `RootCommand`.

ericwj:
That said my table still looks like this:
```
R0C0 R1C1
         R1C0 R1C1
                  etc
```

PR #945, opened 2020-06-25
Add a basic straightforward usage example with muliple commands to th…

chhh:
The most basic usage example which I've been looking for.

Not just a "program without parameters", but a program with sevearal commands which is clear how to invoke.

### DIFF
--- a/samples/tutorial/BasicBinding.md
+++ b/samples/tutorial/BasicBinding.md
@@ -1,0 +1,62 @@
+# Quickstart - binding to methods with options objects
+
+Basic scenario - the program has several "commands"/"verbs" each with its
+set of options.
+
+```cs
+class Program {
+  public static void Main(string[] args) {
+    var cmdLine = CreateCommands();
+    cmdLine.Invoke(args);
+  }
+  
+  private static RootCommand CreateCommands() {
+    var root = new RootCommand();
+
+    var list = new Command("list", "list files in dir");
+    list.AddArgument(new Argument<DirectoryInfo>("input", "path to existing file/dir").ExistingOnly());
+    list.Handler = CommandHandler.Create((OptsCmdList opts) => RunCmdList(opts));
+    
+    var info = new Command("info", "show info");
+    info.AddOption(new Option<string>(new[]{"-a", "--aaa"}){Required = true});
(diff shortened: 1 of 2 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/command-line-api?pullRequest=945) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/command-line-api?pullRequest=945) before we can accept your contribution.<br/><br/>:x: chhh [sign now](https://cla.dotnetfoundation.org/dotnet/command-line-api?pullRequest=945)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/command-line-api?pullRequest=945) it.</sub>

jonsequitur:
The tutorial is interactive. You can see the instructions here: https://github.com/dotnet/command-line-api#interactive-tutorials.

One benefit of this is that the code isn't just embedded in a document but is also compiled.

Do you think we can add missing information so that it has more context within the sample code?

We're also working on improving documentation generally (#130).

chhh:
I actually just wanted to add it to wiki, couldn't find how to do it.

I tried to run the tutorial, but I just get "Internal server error" when trying to run any of the samples.

I don't think the "interactiveness" of the tutorial adds any value. If anything it's a roadblock - users have to install something to just see the tutorial not to mention that it might not really work out of the box for everyone. I'm already trying out the library, which means I'm in the IDE and can copy-paste a few examples.

jonsequitur:
We'll be publishing the interactive tutorials back to Markdown so even if you're not using them interactively, the `dotnet-try` tool still serves to ensure that the sample code compiles as things change over time.

jonsequitur:
Regarding the wiki issue, we've planning to move the docs into a `/docs` subfolder off the repo root in order to support pull requests. I think the ideal approach here is to move this in there and structure it so that `dotnet try publish` can be used to inline the compiler-verified source code into the Markdown.

chhh:
Looking forward to the /docs change.

PR #947, opened 2020-06-26
Wrap HelpItems to table. (#629)

### LINKED ISSUE #629: Enum values should be split to multiple lines
I have the following enum:

```csharp
        public enum ExampleType
        {
            PCSort128,
            PCSort256,
            AVX2,
            AVX2VacuumPacked,
            AVX2VacuumPackedx2,
            AVX2DoublePumped,
            AVX2DoublePumpedNaive,
            AVX2DoublePumpedAligned,
            AVX2DoublePumpedWithPCSort,
            AVX2DoublePumpedUnrolled,
            AVX2DoublePumpedUnrolledWithPCSort,
        }
```

When used in a `Main` function like this:
```
        public static unsafe void Main(ExampleType type, int size, int loops, int modulo, int forceValue = int.MinValue, double forceRate = double.NaN)
```

The generated help "docs" could use some love, as the different values are pushing the entire text to the right, where it's probably more sane to make them take up a few more rows.

Here' what it looks like:

```bash
Example:
  Run various quicksort/sort variants

Usage:
  Example [options]

Options:
  --type <AVX2|AVX2DoublePumped|AVX2DoublePumpedAligned|AVX2DoublePumpedNaive|AVX2DoublePumpedUnrolled|AVX2DoublePumpedUnrolledWithPCSort|AVX2DoublePumpedWithPCSort|AVX2VacuumPacked|AVX2VacuumPackedx2|PCSort128|PCSort256>
                                                                                                                                                                                                                                 The
                                                                                                                                                                                                                                 type
                                                                                                                                                                                                                                 of
                                                                                                                                                                                                                                 sort
                                                                                                                                                                                                                                 routine
                                                                                                                                                                                                                                 to
                                                                                                                                                                                                                                 invoke
  --size <size>                                                                                                                     
                                                                                                                                                                                                                                 The
                                                                                                                                                                                                                                 size
                                                                                                                                                                                                                                 of
                                                                                                                                                                                                                                 the
                                                                                                                                                                                                                                 sort
                                                                                                                                                                                                                                 problem
                                                                                                                                                                                                                                 to
                                                                                                                                                                                                                                 invoke
  --loops <loops>                                                                                                                   
                                                                                                                                                                                                                                 #
                                                                                                                                                                                                                                 of
                                                                                                                                                                                                                                 loops
                                                                                                                                                                                                                                 to
                                                                                                                                                                                                                                 re-sort
                                                                                                                                                                                                                                 the
                                                                                                                                                                                                                                 data
  --modulo <modulo>                                                                                                                 
                                                                                                                                                                                                                                 Squash
                                                                                                                                                                                                                                 all
                                                                                                                                                                                                                                 the
                                                                                                                                                                                                                                 values
                                                                                                                                                                                                                                 into
                                                                                                                                                                                                                                 a
                                                                                                                                                                                                                                 certain
                                                                                                                                                                                                                                 range
  --force-value <force-value>                                                                                                       
                                                                                                                                                                                                                                 force
                                                                                                                                                                                                                                 specific
                                                                                                                                                                                                                                 value
  --force-rate <force-rate>                                                                                                         
                                                                                                                                                                                                                                 how
                                                                                                                                                                                                                                 many
                                                                                                                                                                                                                                 elements
                                                                                                                                                                                                                                 should
                                                                                                                                                                                                                                 be
                                                                                                                                                                                                                                 the
                                                                                                                                                                                                                                 forced
                                                                                                                                                                                                                                 value
  --version                                                                                                                         
                                                                                                                                                                                                                                 Display
                                                                                                                                                                                                                                 version
                                                                                                                                                                                                                                 information

```

I think something along this sort of output would be generally more beneficial in cases there are long lists of enum values:

```bash
Example:
  Run various quicksort/sort variants

Usage:
  Example [options]

Options:
  --type <AVX2|AVX2DoublePumped|AVX2DoublePumpedAligned|                The type of sort routine to invoke
          AVX2DoublePumpedNaive|AVX2DoublePumpedUnrolled|
          AVX2DoublePumpedUnrolledWithPCSort|
          AVX2DoublePumpedWithPCSort|
          AVX2VacuumPacked|AVX2VacuumPackedx2|
          PCSort128|PCSort256>
  --size <size>                                                         The size of the sort problem to invoke                                                         
  --modulo <modulo>                                                     Squash all the values into a certain range                                                            
  --force-value <force-value>                                           force specific value                                                            
  --force-rate <force-rate>                                             how many elements should be the forced value                                                            
  --version                                                             Display version information                                                                                                                         

```

apogeeoak:
Wrap `HelpItem`s to table in `HelpBuilder`.

The column widths of the table are allocated to favor minimal rows. Wider columns are allocated more space if it is available. 

Added tests to `HelpBuilderTest` to check that necessary sections properly wrap.

```
Example:
  Command description.

Usage:
  Example [options]

Options:
  --option-alias-for-a-command-th    Description that is long enough
  at-is-long-enough-to-wrap-to-a-    to wrap to a new line.
  new-line
```

Removed erroneous trailing whitespace in HelpBuilderTests.Help_describes_default_values_for_complex_root_command_scenario.approved.txt.

Additional auto-formatting in HelpBuilderTest.cs. Removed trailing whitespace and some argument alignment.

Fixes #629

### DIFF
--- a/src/System.CommandLine/Help/HelpBuilder.cs
+++ b/src/System.CommandLine/Help/HelpBuilder.cs
@@ -215,7 +215,9 @@ namespace System.CommandLine.Help
         protected virtual void AppendHelpItems(IReadOnlyCollection<HelpItem> helpItems)
         {
             if (helpItems is null)
+            {
                 throw new ArgumentNullException(nameof(helpItems));
+            }
 
             var table = CreateTable(helpItems, item => new[]
             {
@@ -228,21 +230,20 @@ namespace System.CommandLine.Help
         }
 
         /// <summary>
-        /// Create a read only table of strings using the projection of a collection.
+        /// Create a table of strings using the projection of a collection.
         /// </summary>
         /// <typeparam name="T">The type of the elements of <paramref name="collection"/>.</typeparam>
         /// <param name="collection">The collection of values to create the table from.</param>
         /// <param name="selector">A transformation function to apply to each element of <paramref name="collection"/>.</param>
         /// <returns>
-        /// A read only table of strings whose elements are the projection of the collection with whitespace formatting removed.
+        /// A table of strings whose elements are the projection of the collection with whitespace formatting removed.
         /// </returns>
-        protected virtual ReadOnlyCollection<ReadOnlyCollection<string>> CreateTable<T>(IEnumerable<T> collection, Func<T, IEnumerable<string>> selector)
+        protected virtual IEnumerable<IReadOnlyList<string>> CreateTable<T>(IEnumerable<T> collection, Func<T, IEnumerable<string>> selector)
         {
             return collection.Select(selector)
                 .Select(row => row
-                    .Select(element => RemoveFormatting(element))
-                    .ToList().AsReadOnly())
-                .ToList().AsReadOnly();
+                    .Select(element => ShortenWhitespace(element))
+                    .ToArray());
         }
 
         /// <summary>
@@ -250,20 +251,25 @@ namespace System.CommandLine.Help
         /// </summary>
         /// <param name="table">The table of values to determine column widths for.</param>
         /// <returns>A collection of column widths.</returns>
-        protected virtual ReadOnlyCollection<int> ColumnWidths(IEnumerable<IList<string>> table)
+        private IReadOnlyList<int> ColumnWidths(IEnumerable<IReadOnlyList<string>> table)
         {
             if (!table.Any())
-                return Array.AsReadOnly(Array.Empty<int>());
+            {
+                return Array.Empty<int>();
+            }
 
             var columns = table.First().Count;
-            Debug.Assert(table.All(e => e.Count == columns), $"Every row in {nameof(table)} must have the same number of columns.");
             var unsetWidth = -1;
             var widths = new int[columns];
             for (int i = 0; i < columns; ++i)
+            {
                 widths[i] = unsetWidth;
+            }
             var maxWidths = new int[columns];
             for (int i = 0; i < columns; ++i)
+            {
                 maxWidths[i] = table.Max(row => row[i].Length);
+            }
 
             var nonEmptyColumns = maxWidths.Count(width => width > 0);
 
@@ -272,7 +278,9 @@ namespace System.CommandLine.Help
             // If available space is not sufficent then do not wrap.
             // If all columns are empty then return array of zeros.
             if (available - nonEmptyColumns < 0 || nonEmptyColumns == 0)
-                return Array.AsReadOnly(maxWidths);
+            {
+                return maxWidths;
+            }
 
             // Loop variables.
             var unset = nonEmptyColumns;
@@ -293,9 +301,13 @@ namespace System.CommandLine.Help
                         // Attempt to fit column to single line.
                         var width = maxWidths[i];
                         if (allocateRemaining)
+                        {
                             width = Math.Min(width, equal);
+                        }
                         if (width <= equal)
+                        {
                             widths[i] = width;
+                        }
                     }
                 }
                 previousUnset = unset;
@@ -303,7 +315,7 @@ namespace System.CommandLine.Help
                 --loopLimit;
             }
 
-            return Array.AsReadOnly(widths);
+            return widths;
         }
 
         /// <summary>
@@ -311,7 +323,7 @@ namespace System.CommandLine.Help
         /// </summary>
         /// <param name="table">The table of values to write.</param>
         /// <param name="columnWidths">The width of each column of the table.</param>
-        protected virtual void AppendTable(IEnumerable<IEnumerable<string>> table, IReadOnlyList<int> columnWidths)
+        private void AppendTable(IEnumerable<IEnumerable<string>> table, IReadOnlyList<int> columnWidths)
         {
             foreach (var row in table)
                 AppendRow(row, columnWidths);
@@ -322,7 +334,7 @@ namespace System.CommandLine.Help
         /// </summary>
         /// <param name="row">The row of elements to write.</param>
         /// <param name="columnWidths">The width of each column of the table.</param>
-        protected virtual void AppendRow(IEnumerable<string> row, IReadOnlyList<int> columnWidths)
+        private void AppendRow(IEnumerable<string> row, IReadOnlyList<int> columnWidths)
         {
             var split = row.Select((element, index) => SplitText(element, columnWidths[index])).ToArray();
             var longest = split.Max(lines => lines.Count);
@@ -361,7 +373,7 @@ namespace System.CommandLine.Help
         /// Collection of lines of at most <paramref name="width"/> characters
         /// generated from the supplied <paramref name="text"/>.
         /// </returns>
-        protected virtual ReadOnlyCollection<string> SplitText(string text, int width)
+        protected virtual IReadOnlyList<string> SplitText(string text, int width)
         {
             if (text is null)
                 throw new ArgumentNullException(nameof(text), $"{nameof(text)} cannot be null.");
@@ -369,13 +381,13 @@ namespace System.CommandLine.Help
                 throw new ArgumentOutOfRangeException(nameof(width), $"{nameof(width)} must be non-negative.");
 
             if (width == 0)
-                return Array.AsReadOnly(Array.Empty<string>());
+                return Array.Empty<string>();
 
             var separator = ' ';
 
             var start = 0;
             var lines = new List<string>();
-            text = RemoveFormatting(text);
+            text = ShortenWhitespace(text);
 
             while (start < text.Length - width)
             {
@@ -398,7 +410,7 @@ namespace System.CommandLine.Help
 
             lines.Add(text.Substring(start, text.Length - start));
 
-            return lines.AsReadOnly();
+            return lines;
         }
 
         /// <summary>
@@ -724,20 +736,13 @@ namespace System.CommandLine.Help
 
         private int GetConsoleWindowWidth(IConsole console)
         {
-            try
-            {
-                if (console is SystemConsole && System.Console.WindowWidth > 0)
-                    return System.Console.WindowWidth;
-                else
-                    return int.MaxValue;
-            }
-            catch
-            {
+            if (console is SystemConsole systemConsole)
+                return systemConsole.GetConsoleWindowWidth();
+            else
                 return int.MaxValue;
-            }
         }
 
-        private string RemoveFormatting(string input)
+        private string ShortenWhitespace(string input)
         {
             return Regex.Replace(input, @"\s+", " ");
         }
--- a/src/System.CommandLine/Help/HelpBuilder.cs
+++ b/src/System.CommandLine/Help/HelpBuilder.cs
@@ -2,8 +2,10 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System.Collections.Generic;
+using System.Collections.ObjectModel;
 using System.CommandLine.IO;
 using System.CommandLine.Parsing;
+using System.Diagnostics;
 using System.Linq;
 using System.Text;
 using System.Text.RegularExpressions;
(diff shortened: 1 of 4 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/command-line-api?pullRequest=947) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/command-line-api?pullRequest=947) before we can accept your contribution.<br/><br/>:x: apogeeoak [sign now](https://cla.dotnetfoundation.org/dotnet/command-line-api?pullRequest=947)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/command-line-api?pullRequest=947) it.</sub>

src/System.CommandLine/Help/HelpBuilder.cs
```diff
@@ -204,103 +206,199 @@ private void AppendDescription(string description)
...
+        protected virtual void AppendHelpItems(IReadOnlyCollection<HelpItem> helpItems)
         {
-            if (helpItem is null)
-            {
-                throw new ArgumentNullException(nameof(helpItem));
-            }
-
-            AppendText(helpItem.Invocation, CurrentIndentation);
+            if (helpItems is null)
+                throw new ArgumentNullException(nameof(helpItems));
```
jonsequitur, line 218: Braces please.

src/System.CommandLine/Help/HelpBuilder.cs
```diff
@@ -204,103 +206,199 @@ private void AppendDescription(string description)
...
+        /// <param name="selector">A transformation function to apply to each element of <paramref name="collection"/>.</param>
+        /// <returns>
+        /// A read only table of strings whose elements are the projection of the collection with whitespace formatting removed.
+        /// </returns>
+        protected virtual ReadOnlyCollection<ReadOnlyCollection<string>> CreateTable<T>(IEnumerable<T> collection, Func<T, IEnumerable<string>> selector)
+        {
+            return collection.Select(selector)
+                .Select(row => row
+                    .Select(element => RemoveFormatting(element))
+                    .ToList().AsReadOnly())
```
jonsequitur, line 244: Is this needed? Does the the outer `ToList` suffice?

src/System.CommandLine/Help/HelpBuilder.cs
```diff
@@ -204,103 +206,199 @@ private void AppendDescription(string description)
...
+        /// </summary>
+        /// <param name="table">The table of values to determine column widths for.</param>
+        /// <returns>A collection of column widths.</returns>
+        protected virtual ReadOnlyCollection<int> ColumnWidths(IEnumerable<IList<string>> table)
+        {
+            if (!table.Any())
+                return Array.AsReadOnly(Array.Empty<int>());
+
+            var columns = table.First().Count;
+            Debug.Assert(table.All(e => e.Count == columns), $"Every row in {nameof(table)} must have the same number of columns.");
```
jonsequitur, line 259: Please remove the `Debug.Assert`.

src/System.CommandLine/Help/HelpBuilder.cs
```diff
@@ -204,103 +206,199 @@ private void AppendDescription(string description)
...
+            var unsetWidth = -1;
+            var widths = new int[columns];
+            for (int i = 0; i < columns; ++i)
+                widths[i] = unsetWidth;
```
jonsequitur, line 263: Please add wrapping braces throughout.

src/System.CommandLine/Help/HelpBuilder.cs
```diff
@@ -204,103 +206,199 @@ private void AppendDescription(string description)
...
+                    .ToList().AsReadOnly())
+                .ToList().AsReadOnly();
+        }
+
+        /// <summary>
+        /// Allocate space for columns favoring minimal rows. Wider columns are allocated more space if it is available.
+        /// </summary>
+        /// <param name="table">The table of values to determine column widths for.</param>
+        /// <returns>A collection of column widths.</returns>
+        protected virtual ReadOnlyCollection<int> ColumnWidths(IEnumerable<IList<string>> table)
```
jonsequitur, line 253: Let's default this to private for now.

src/System.CommandLine/Help/HelpBuilder.cs
```diff
@@ -204,103 +206,199 @@ private void AppendDescription(string description)
...
+            return Array.AsReadOnly(widths);
+        }
 
-            foreach (var descriptionLine in descriptionLines.Skip(1))
+        /// <summary>
+        /// Writes a table of strings to the console.
+        /// </summary>
+        /// <param name="table">The table of values to write.</param>
+        /// <param name="columnWidths">The width of each column of the table.</param>
+        protected virtual void AppendTable(IEnumerable<IEnumerable<string>> table, IReadOnlyList<int> columnWidths)
```
jonsequitur, line 314: Let's default this to private for now.

src/System.CommandLine/Help/HelpBuilder.cs
```diff
@@ -204,103 +206,199 @@ private void AppendDescription(string description)
...
+            foreach (var row in table)
+                AppendRow(row, columnWidths);
+        }
+
+        /// <summary>
+        /// Writes a row of strings to the console with columns of the given width.
+        /// </summary>
+        /// <param name="row">The row of elements to write.</param>
+        /// <param name="columnWidths">The width of each column of the table.</param>
+        protected virtual void AppendRow(IEnumerable<string> row, IReadOnlyList<int> columnWidths)
```
jonsequitur, line 325: Let's default this to private for now.

src/System.CommandLine/Help/HelpBuilder.cs
```diff
@@ -204,103 +206,199 @@ private void AppendDescription(string description)
...
+        /// <param name="text">Text content to split into lines.</param>
+        /// <param name="width">Maximum number of characters allowed per line.</param>
         /// <returns>
-        /// Collection of lines of at most <see cref="maxLength"/> characters
-        /// generated from the supplied <see cref="text"/>
+        /// Collection of lines of at most <paramref name="width"/> characters
+        /// generated from the supplied <paramref name="text"/>.
         /// </returns>
-        protected virtual IReadOnlyCollection<string> SplitText(string text, int maxLength)
+        protected virtual ReadOnlyCollection<string> SplitText(string text, int width)
```
jonsequitur, line 364: What's the reason for changing this to a concrete type?

If it's useful to do this, we could also make this method private.

src/System.CommandLine/Help/HelpBuilder.cs
```diff
@@ -624,18 +722,31 @@ private bool ShouldDisplayArgumentHelp(ICommand? command)
             return command.Arguments.Any(ShouldShowHelp);
         }
 
-        private int GetConsoleWindowWidth()
+        private int GetConsoleWindowWidth(IConsole console)
         {
             try
             {
-                return System.Console.WindowWidth;
+                if (console is SystemConsole && System.Console.WindowWidth > 0)
```
jonsequitur, line 729: The thing to watch out for here is that `System.Console.WindowWidth` will throw there's no window (e.g., output is redirected). We generally avoid calling `System.Console` directly for this reason. You can't get this information unless there's actually a terminal (the `ITerminal` abstraction in `System.CommandLine.Rendering`.

I think we can probably away with the exception handler here altogether and do something like this instead:

```cs
private static int GetConsoleWindowWidth(SystemConsole console)
{
  return console.IsOutputRedirected
    ? int.MaxValue
    : System.Console.WindowWidth;
}
```

It might make sense to move this to the `SystemConsole` class in fact.

src/System.CommandLine/Help/HelpBuilder.cs
```diff
@@ -624,18 +722,31 @@ private bool ShouldDisplayArgumentHelp(ICommand? command)
...
+                    return int.MaxValue;
             }
-            catch (System.IO.IOException)
+            catch
             {
                 return int.MaxValue;
             }
         }
 
+        private string RemoveFormatting(string input)
```
jonsequitur, line 740: A more honest name might be something like `ShortenWhitespace`.

src/System.CommandLine/Help/HelpBuilder.cs
apogeeoak: A ReadOnlyCollection is not necessary but I used one to follow the [guidelines for collections](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/guidelines-for-collections).

I think a multidimensional array is better suited to represent a table. Originally I was going to implement the table view as a separate class with a multidimensional array backing but decided to implement it in place instead so I opted for a ReadOnlyCollection.

src/System.CommandLine/Help/HelpBuilder.cs
apogeeoak: I can remove the `Debug.Assert` but if `ColumnWidths` is called with an unequal number of columns an exception will be thrown at an indeterminate time. This should never happen in production so I thought an O(n) check for this was generally wasteful.

src/System.CommandLine/Help/HelpBuilder.cs
apogeeoak: I made it concrete to follow the [guidelines for collections](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/guidelines-for-collections).

I can return an interface but I need indexed access so I would have to return either `IList` or `IReadOnlyList`.

src/System.CommandLine/Help/HelpBuilder.cs
apogeeoak: I can move it.

src/System.CommandLine/Help/HelpBuilder.cs
jonsequitur: I believe that guidance predates the introduction of `IReadOnlyCollection<T>` and `IReadOnlyList<T>`. Generally, it's good to return a less specific type than a more specific type.

I'd avoid the multidimensional array as it's harder to reason about.

We have a table abstraction in System.CommandLine.Rendering. That project was started in part out of a desire to have better and more flexible help layouts. I wonder if there's an opportunity for reuse or convergence there. 

@Keboo?

src/System.CommandLine/Help/HelpBuilder.cs
jonsequitur: Would a compensation of some kind be preferable to an exception? Or can we make the incorrect state unrepresentable by the object model?

src/System.CommandLine/Help/HelpBuilder.cs
jonsequitur: `IReadOnlyList` is good.

src/System.CommandLine/Help/HelpBuilder.cs
apogeeoak: If the object model was a multidimensional array incorrect state would be unrepresentable as it would guarantee a consistent number of columns.

`CreateTable` will always return a table with a consistent number of columns but if `ColumnWidths` is called with arbitrary input then this cannot be guaranteed.

apogeeoak:
I can squash the commits on approval.

src/System.CommandLine/Help/HelpBuilder.cs
```diff
@@ -624,16 +734,22 @@ private bool ShouldDisplayArgumentHelp(ICommand? command)
...
+        private int GetConsoleWindowWidth(IConsole console)
         {
-            try
-            {
-                return System.Console.WindowWidth;
-            }
-            catch (System.IO.IOException)
-            {
+            if (console is SystemConsole systemConsole)
+                return systemConsole.GetConsoleWindowWidth();
```
jonsequitur, line 740: Please add braces.

jonsequitur:
Thanks, @apogeeoak! This is a really nice improvement.

apogeeoak:
I would still like to make some changes to this and then squash the commits if possible.

I would like to add in the braces I forgot to add and change `CreateTable` to immediately evaluate and return an `IReadOnlyList` instead of an `IEnumerable`. I think the immediate evaluation is worth the added memory since the result is queried twice and may be an expensive operation. I was also considering changing the caching function from `ToArray` to `ToList` but I do not know which is better.

jonsequitur:
Sorry for the over-eager merge. Further PRs to make optimizations are more than welcome.

apogeeoak:
Should I create a new PR or reuse this one?

jonsequitur:
Please create a new one.

PR #949, opened 2020-06-26
Add ParseResult.ValueForArgument and ValueForOption

jonsequitur:
This aims to further simplify looking up values from a `ParseResult`.

### DIFF
--- a/src/System.CommandLine/Parsing/ParseResult.cs
+++ b/src/System.CommandLine/Parsing/ParseResult.cs
@@ -88,15 +88,10 @@ namespace System.CommandLine.Parsing
         [return: MaybeNull]
         public T ValueForArgument<T>(Argument<T> argument)
         {
-            var result = FindResultFor(argument);
-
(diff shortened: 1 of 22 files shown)


src/System.CommandLine/Parsing/ParseResult.cs
```diff
@@ -85,6 +85,54 @@ public T ValueForOption<T>(string alias)
             }
         }
 
+        [return: MaybeNull]
+        public T ValueForArgument<T>(Argument<T> argument)
+        {
+            if (FindResultFor(argument) is {} result &&
+                result.GetValueOrDefault<T>() is {} t)
```
Keboo, line 92: nit: We are not being consistent with the spacing on the pattern `{}` vs `{ }`

jonsequitur: Formatting consistency is something we might want to set up globally and enforce with `dotnet-format`.

PR #950, opened 2020-06-28
Update Generic Host projects for .NET Standard 2.1 / .NET Core 3.0

fredrikhr:
There were some breaking changes that need to be addressed for the Generic Host infrastructure for .NET Core 3.0.

However, the previous Generic Host implementation will still continue to work for .NET Standard 2.0 projects. Therefore, the System.CommandLine.Hosting package will now be multi-targeting .NET Standard 2.0 and 2.1. .NET Core Apps targeting 3.0 or later will automatically resolve to use the .NET Standard 2.1 variant of the package.

## Details

* `Microsoft.Extensions.Hosting.IHostingEnvironment` has been replaced by  
   `Microsoft.Extensions.Hosting.IHostEnvironment`
* `Microsoft.Extensions.Hosting.IApplicationLifetime` has been replaced by  
   `Microsoft.Extensions.Hosting.IHostApplicationLifetime`

The replacing types listed above are fully compatible with the old types. Since the Generic Host heavily relies on DI the difference in types in the Constructor signature will most likely not affect any consumers of our API, since nobody really was expected to new the `InvocationLifetime` class manually anyways.

### DIFF
--- a/samples/HostingPlayground/HostingPlayground.csproj
+++ b/samples/HostingPlayground/HostingPlayground.csproj
@@ -7,17 +7,9 @@
     <LangVersion>8</LangVersion>
   </PropertyGroup>
 
-  
   <ItemGroup>
     <ProjectReference Include="..\..\src\System.CommandLine.Hosting\System.CommandLine.Hosting.csproj" />
     <ProjectReference Include="..\..\src\System.CommandLine\System.CommandLine.csproj" />
-  </ItemGroup>
-
-  
-  <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="3.1.4" />
-    <PackageReference Include="microsoft.extensions.hosting" Version="3.1.4" />
-    <PackageReference Include="Microsoft.Extensions.Logging" Version="3.1.4" />
   </ItemGroup>
 
 </Project>
--- a/src/System.CommandLine.Hosting.Tests/System.CommandLine.Hosting.Tests.csproj
+++ b/src/System.CommandLine.Hosting.Tests/System.CommandLine.Hosting.Tests.csproj
@@ -15,6 +15,9 @@
 
   <ItemGroup>
     <PackageReference Include="FluentAssertions" Version="5.10.3" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)'!='netcoreapp3.1'">
     <PackageReference Include="Microsoft.Extensions.Configuration.CommandLine" Version="2.2.0" />
   </ItemGroup>
 
--- a/src/System.CommandLine.Hosting/HostingExtensions.cs
+++ b/src/System.CommandLine.Hosting/HostingExtensions.cs
@@ -39,16 +39,15 @@ namespace System.CommandLine.Hosting
                 hostBuilder.UseInvocationLifetime(invocation);
                 configureHost?.Invoke(hostBuilder);
 
-                using (var host = hostBuilder.Build())
-                {
-                    invocation.BindingContext.AddService(typeof(IHost), _ => host);
+                using var host = hostBuilder.Build();
 
-                    await host.StartAsync();
+                invocation.BindingContext.AddService(typeof(IHost), _ => host);
 
-                    await next(invocation);
+                await host.StartAsync();
 
-                    await host.StopAsync();
-                }
+                await next(invocation);
+
+                await host.StopAsync();
             });
 
         public static CommandLineBuilder UseHost(this CommandLineBuilder builder,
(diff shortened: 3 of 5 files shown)


samples/HostingPlayground/HostingPlayground.csproj
```diff
@@ -7,17 +7,9 @@
...
   </ItemGroup>
 
-  
-  <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="3.1.4" />
-    <PackageReference Include="microsoft.extensions.hosting" Version="3.1.4" />
-    <PackageReference Include="Microsoft.Extensions.Logging" Version="3.1.4" />
-  </ItemGroup>
-
 </Project>
```
fredrikhr, line 15: In .NET Standard 2.1, the `Microsoft.Extensions.Hosting` package includes all of the necessary Generic Host infratrstructure (much like the ASP.NET Core `App` and `All` metapackages did in ASP.NET Core land). It is therefore no longer necessary to refer to single Generic Host packages individually. Doing so might actually cause version conflicts.

src/System.CommandLine.Hosting.Tests/System.CommandLine.Hosting.Tests.csproj
```diff
@@ -15,6 +15,9 @@
 
   <ItemGroup>
     <PackageReference Include="FluentAssertions" Version="5.10.3" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)'!='netcoreapp3.1'">
     <PackageReference Include="Microsoft.Extensions.Configuration.CommandLine" Version="2.2.0" />
```
fredrikhr, line 21: Previously in .NET Standard 2.0 (and thus also the `net462` target), individual Generic Host packages needed to be referenced individually, since `Microsoft.Extensions.Hosting` only contained the bare minimum.

`Microsoft.Extensions.Configuration.CommandLine` is actually trasitively references by `Microsoft.Extensions.Hosting` in .NET Standard 2.1 (and .NET Core 3.0 and later)

src/System.CommandLine.Hosting/HostingExtensions.cs
```diff
@@ -39,16 +39,15 @@ public static class HostingExtensions
...
 
-                    await host.StartAsync();
+                await host.StartAsync();
 
-                    await next(invocation);
+                await next(invocation);
 
-                    await host.StopAsync();
-                }
+                await host.StopAsync();
```
fredrikhr, line 50: Might as well follow Visual Studio's `Use simple using statement` (introduced in C# 8) suggestion when updating the package.

src/System.CommandLine.Hosting/InvocationLifetime.cs
```diff
@@ -7,6 +7,11 @@
 using Microsoft.Extensions.Logging.Abstractions;
 using Microsoft.Extensions.Options;
 
+#if NETSTANDARD2_0
+using IHostEnvironment = Microsoft.Extensions.Hosting.IHostingEnvironment;
+using IHostApplicationLifetime = Microsoft.Extensions.Hosting.IApplicationLifetime;
+#endif
+
```
fredrikhr, line 14: The types listed here are deprecated in `netstandard2.1`. But the new types do not exist in `netstandard2.0`. Using these conditional using-aliasing statements ensures that we use the new types from now on.

src/System.CommandLine.Hosting/InvocationLifetime.cs
```diff
@@ -18,8 +23,8 @@ public class InvocationLifetime : IHostLifetime
 
         public InvocationLifetime(
             IOptions<InvocationLifetimeOptions> options,
-            IHostingEnvironment environment,
-            IApplicationLifetime applicationLifetime,
+            IHostEnvironment environment,
+            IHostApplicationLifetime applicationLifetime,
```
fredrikhr, line 27: These arguments are expected to be passed via DI, so the change is not expected to affect consumers of the library.

src/System.CommandLine.Hosting/System.CommandLine.Hosting.csproj
```diff
@@ -11,9 +11,12 @@
     <DebugType>portable</DebugType>
   </PropertyGroup>
 
-  <ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)'=='netstandard2.0'">
     <PackageReference Include="Microsoft.Extensions.Hosting" Version="2.2.0" />
   </ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)'=='netstandard2.1'">
+    <PackageReference Include="Microsoft.Extensions.Hosting" Version="3.1.5" />
+  </ItemGroup>
```
fredrikhr, line 19: Multi-targeting package reference for .NET Standard 2.0 and 2.1.

I suggest that we will drop support for .NET Standard 2.0 at some point in the future for maintainability reasons. However, it might be prudent to temporarily support both TFMs in a transition period until .NET Core 2.1 is no longer covered by LTS support (scheduled for 2021-08-21) and .NET 5 (or maybe even .NET 6) has been firmly established.

Issue #952, opened 2020-06-28
Support explicit binding for parameters


KathleenDollard:
For type models, you can explicitly set Member/Symbol (property/Option or Argument) pairs. Constructor parameters can also be explicitly specified. 

You can't do this today for invocation parameters and we want this.

This is important for app models like StarFruit where we can improve performance (along with source generation) by avoiding the name lookup. And it gives tighter control in some scenarios where names might get odd (although some of those scenarios are pathological). 

It's straightforward to add a new dictionary and BindTo.. method.

The challenge is calling them. The current design in `ModelBindingCommandHandler.InvokeAsync` is to create a `ModelBinder` for each parameter. I want to change this so I can retrieve a pre-configured `ModelBinder` for either the group of parameters (some redefinition of what a model binder is) or per parameter. However, the current 'BindingContext' retrieval of a preconfigured `ModelBinder` is via only the type.  Obviously we might have multiple integer parameters, so this doesn't work. 

There are multiple approaches we could take. Broadly grouping a few: 

* Change how the BindingContext retrieves a model binder and allow a ModelBinder to be preconfigured with a ModelBinder for each parameter
* Alter ModelBinder to support parameters in a different way
* Create a new MethodBinder that parallels ModelBinder and shares code where logical

The design needs to accommodate combinations - properties in types containing invoked methods having their data filled prior to invocation and parameters on invoked methods being complex types that are populated. My gut is that there is a model that nests these in more complex ways. But when I have tried whiteboarding that it was more complex than inherited app models, but we probably want to stay open to that at least until F# models are understood. 

I lean heavily to the third approach, but want to discuss it to be sure I'm not overlooking possibilities at smaller tweaks.

PR #953, opened 2020-06-28
Explicit parameter binding

KathleenDollard:
WIP: This doesn't currently support complex parameter types (there is a failing test to illustrate). It does handle explicit binding directly to parameters and existing tests pass. 

Work on #952


KathleenDollard:
Replaced by #1009

PR #955, opened 2020-06-30
Allow Suggestions to be replaced

jmarolf:
prerequisite to enable https://github.com/dotnet/format/pull/711 to be merged

### DIFF
--- a/src/System.CommandLine.Tests/SuggestDirectiveTests.cs
+++ b/src/System.CommandLine.Tests/SuggestDirectiveTests.cs
@@ -22,10 +22,10 @@ namespace System.CommandLine.Tests
         public SuggestDirectiveTests()
         {
             _fruitOption = new Option<string>("--fruit");
-            _fruitOption.Argument.SuggestionSources.Add( new[] { "apple", "banana", "cherry" });
+            _fruitOption.Argument.Suggestions.Add( new[] { "apple", "banana", "cherry" });
 
             _vegetableOption = new Option<string>("--vegetable");
-            _vegetableOption.Argument.SuggestionSources.Add(new[] { "asparagus", "broccoli", "carrot" });
+            _vegetableOption.Argument.Suggestions.Add(new[] { "asparagus", "broccoli", "carrot" });
 
             _eatCommand = new Command("eat")
             {
--- a/src/System.CommandLine.Tests/SuggestDirectiveTests.cs
+++ b/src/System.CommandLine.Tests/SuggestDirectiveTests.cs
@@ -21,11 +21,11 @@ namespace System.CommandLine.Tests
 
         public SuggestDirectiveTests()
         {
-            _fruitOption = new Option<string>("--fruit");
-            _fruitOption.Argument.Suggestions.Add( new[] { "apple", "banana", "cherry" });
+            _fruitOption = new Option<string>("--fruit")
+                .AddSuggestions("apple", "banana", "cherry");
 
-            _vegetableOption = new Option<string>("--vegetable");
-            _vegetableOption.Argument.Suggestions.Add(new[] { "asparagus", "broccoli", "carrot" });
+            _vegetableOption = new Option<string>("--vegetable")
+                .AddSuggestions("asparagus", "broccoli", "carrot");
 
             _eatCommand = new Command("eat")
             {
--- a/src/System.CommandLine.Tests/SuggestDirectiveTests.cs
+++ b/src/System.CommandLine.Tests/SuggestDirectiveTests.cs
@@ -2,7 +2,6 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System.CommandLine.Builder;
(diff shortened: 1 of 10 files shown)


src/System.CommandLine/ArgumentExtensions.cs
```diff
@@ -41,6 +61,17 @@ public static class ArgumentExtensions
 
             return argument;
         }
+
+        public static TArgument WithAutomaticallyGenerateSuggestionsBasedOnTypee<TArgument>(
```
KathleenDollard, line 65: Typo: ...OnTypee

jonsequitur:
Please add some tests. One test that would really help clarify the scenario is the one we discussed where you're parsing to an enum allowing a number of "alias" values.

jmarolf:
> Please add some tests. One test that would really help clarify the scenario is the one we discussed where you're parsing to an enum allowing a number of "alias" values.

sure thing I was going to wait to see if there was any desire to take this API first

jmarolf:
updated with tests

jmarolf:
all I need really is the ability to disable `SuggestionSource.ForType(ArgumentType).GetSuggestions(textToMatch)` for enums

src/System.CommandLine.Tests/SuggestDirectiveTests.cs
```diff
@@ -273,5 +277,149 @@ public async Task It_does_not_repeat_suggestion_for_already_specified_bool_optio
...
+                    Argument = new Argument<LogLevel?>(parse: ParseLogLevel)
+                }
+            };
+
+            var console = new TestConsole();
+            var parser = new CommandLineBuilder(command)
+             .UseDefaults()
+             .Build();
+
+            var result = parser.Parse($"[suggest:13] \"--verbosity\"");
```
jonsequitur, line 296: You can simplify this by skipping the builder and just doing `command.Parse`.

src/System.CommandLine.Tests/SuggestDirectiveTests.cs
```diff
@@ -273,5 +277,149 @@ public async Task It_does_not_repeat_suggestion_for_already_specified_bool_optio
                    .Should()
                    .NotContain("--bool-option");
         }
+
+        [Fact]
+        public async Task Deafault_Enum_suggestions_match_Enum_member_Names_Even_with_custom_parser() {
```
jonsequitur, line 282: These tests can be a little more focused and specific by moving them to `SuggestionTests` and following the approach there, which is to call `GetSuggestions` directly. No need to assert on the console output.

src/System.CommandLine.Tests/SuggestDirectiveTests.cs
```diff
@@ -22,11 +24,11 @@ public class SuggestDirectiveTests
 
         public SuggestDirectiveTests()
         {
-            _fruitOption = new Option<string>("--fruit")
-                .WithSuggestions("apple", "banana", "cherry");
+            _fruitOption = new Option<string>("--fruit");
+            _fruitOption.Argument.SuggestionSources.Add( new[] { "apple", "banana", "cherry" });
```
jmarolf, line 28: This is the one place that feels like a take-back in the API. I could add a `AddSuggestions` to `Option` to improve this.

src/System.CommandLine.Tests/SuggestionTests.cs
```diff
@@ -997,6 +993,40 @@ public void When_there_are_multiple_arguments_then_suggestions_are_only_offered_
...
+            }
+
+            [Fact]
+            public void Enum_suggestions_can_be_configured_without_list_clear()
+            {
+                var command = new Command("the-command")
+                {
+                    new Argument<LogLevel?>()
+                    {
+                        SuggestionSources = { "q", "quiet", "m", "minimal", "n", "normal", "d", "detailed", "diag", "diagnostic" }
```
jonsequitur, line 1021: Seeing this written this way, it looks like it would read much more naturally as:

```cs
Suggestions = { "q", "quiet", "m", "minimal", "n", "normal", "d", "detailed", "diag", "diagnostic" }
```

src/System.CommandLine.Tests/System.CommandLine.Tests.csproj
```diff
@@ -21,6 +21,7 @@
   <ItemGroup>
     <PackageReference Include="ApprovalTests" Version="5.2.2" />
     <PackageReference Include="FluentAssertions" Version="5.10.3" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="3.1.5" />
```
jonsequitur, line 24: If we're just adding a dependency for the purpose of the one enum, it would be better just to introduce an enum or use one that's already defined in the BCL.

src/System.CommandLine.Tests/SuggestDirectiveTests.cs
jonsequitur: I like that idea.

PR #965, opened 2020-07-06
Add Environment Variable directive parsing

fredrikhr:
# TL;DR

* Adds a new extension method `UseEnvironmentVariableDirective()` to `CommandLineBuilder` enabling to control process-level environment variables for the invocation pipeline from command-line directives.  
   The new directive is: `env`.

# Details

As part of the discussion in #951 the need for an environment variable directive could be benefitial for invocations as well.

This PR adds a parser that intercepts the `env` directive and extracts a `=` separated key-value-pair from the directive value.

## Added to `UseDefaults()`

Since the ability to easily control process-level environment variables is quite useful, I propose that we should add the environment variable directive to the `UseDefaults()` extension method on `CommandLineBuilder`.

This directive is especially useful in shells (most notably on Windows) where specifying variables during process invocation is not possible without modifying the environment variables of the hosting shell.

# Behaviour

The value specified in the `env` directive is expected to contain a `=` separated key-value-pair. 

1. The value of the directive is split into two parts at the first ocurrence of `=`.  
    The first component is considered to be the environment variable to set, the second component is the new value for the variable.
2. Both the variable name and the value are trimmed, stripping leading and trailing whitespace.
3. `System.Envrionment.SetEnvironmentVariable(string variable, string value)` is invoked, setting the variable at process-level, overriding the existing value (if any).

## Invalid Directives

An `env`-directive is ignored if:

* The value of the directive (the text following the `:` in the directive argument) is an empty or white-space only string.
* The value of the directive does not contain at least one `=` character.
* The string preceeding the first `=` in the directive value is an empty or white-space only string.

## Limitations

This directive cannot modify the behaviour of the .NET Runtime that is controlled by environment variables. When the `env`-directive is parsed, the .NET Runtime Host has already booted the runtime and envaluated the environment variables at process launch.

Environment variables that are controlled by the .NET Generic Host can be affected by this directive, provided that the Generic Host is started **after** the invocation pipeline has parsed `env`-directives, or after the pipeline has finished.

## Invocation Pipeline precedence

The middle to parse the `env`-directives is added immediately following the `ExceptionHandler` middleware to ensure that environment variables are parsed as early in the pipeline as possible. `Envrionment.SetEnvironmentVariable` might throw an exception for weird combinatinations for variable names, and these exception should be handled and printed properly using the exception-handling middleware (if present in the pipeline).

### DIFF
--- a/src/System.CommandLine.Tests/EnvironmentVariableDirectiveTests.cs
+++ b/src/System.CommandLine.Tests/EnvironmentVariableDirectiveTests.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.CommandLine.Builder;
 using System.CommandLine.Invocation;
 using System.CommandLine.Parsing;
+using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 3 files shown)


src/System.CommandLine.Tests/EnvironmentVariableDirectiveTests.cs
```diff
@@ -0,0 +1,185 @@
...
+        {
+            bool asserted = false;
+            string variable = $"TEST_ENVIRONMENT_VARIABLE{randomizer.Next()}";
+            const string value = "This is a test";
+            var rootCommand = new RootCommand
+            {
+                Handler = CommandHandler.Create(() =>
+                {
+                    asserted = true;
+                    Assert.Equal(value, Environment.GetEnvironmentVariable(variable));
```
jonsequitur, line 29: Prefer the use of FluentAssertions.

src/System.CommandLine/Builder/CommandLineBuilderExtensions.cs
```diff
@@ -228,11 +228,36 @@ public static CommandLineBuilder CancelOnProcessTermination(this CommandLineBuil
...
+            builder.AddMiddleware((context, next) =>
+            {
+                if (context.ParseResult.Directives.TryGetValues("env", out var directives))
+                {
+                    foreach (var envDirective in directives)
+                    {
+                        var components = envDirective.Split(new[] { '=' }, count: 2);
+                        var variable = components.Length > 0 ? components[0].Trim() : string.Empty;
+                        if (string.IsNullOrEmpty(variable) || components.Length < 2)
+                            continue;
```
jonsequitur, line 243: Please add braces.

src/System.CommandLine.Tests/EnvironmentVariableDirectiveTests.cs
```diff
@@ -0,0 +1,185 @@
...
+{
+    public class EnvironmentVariableDirectiveTests
+    {
+        private static readonly Random randomizer = new Random(Seed: 456476756);
+
+        [Fact]
+        public static async Task Sets_environment_variable_to_value()
+        {
+            bool asserted = false;
+            string variable = $"TEST_ENVIRONMENT_VARIABLE{randomizer.Next()}";
```
jonsequitur, line 22: Rather than repeating this setup, you could assign the environment variable name to a field in the constructor.

KathleenDollard:
@fredrikhr Thanks for this!

src/System.CommandLine.Tests/EnvironmentVariableDirectiveTests.cs
fredrikhr: 0f75261

src/System.CommandLine/Builder/CommandLineBuilderExtensions.cs
fredrikhr: b755924

src/System.CommandLine.Tests/EnvironmentVariableDirectiveTests.cs
fredrikhr: 51bd07b

jonsequitur:
Thanks @fredrikhr!

PR #972, opened 2020-07-12
Improved lookups

jonsequitur:
This improves a number of lookups by shifting from iterations to `HashSet` and `Dictionary` lookups. 

It also adds a `SymbolResult.FindResultFor` method which can be used to simplify the usage of `ParseArgument<T>` by providing a more robust mechanism than traversing the result tree.

### DIFF
--- a/src/System.CommandLine/Option.cs
+++ b/src/System.CommandLine/Option.cs
@@ -56,7 +56,7 @@ namespace System.CommandLine
 
         private protected override void ChooseNameForUnnamedArgument(Argument argument)
         {
-            argument.Name = Aliases[0].ToLower();
+            argument.Name = Name.ToLower();
         }
     }
 }
--- a/src/System.CommandLine/Parsing/StringExtensions.cs
+++ b/src/System.CommandLine/Parsing/StringExtensions.cs
@@ -566,7 +566,8 @@ namespace System.CommandLine.Parsing
 
             for (var commandAliasIndex = 0; commandAliasIndex < command.RawAliases.Count; commandAliasIndex++)
             {
-                var commandAlias = command.RawAliases[commandAliasIndex];
+                var commandAlias = command.RawAliases.ToArray()[commandAliasIndex];
+                
                 tokens.Add(
                     commandAlias,
                     new Token(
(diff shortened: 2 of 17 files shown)


src/System.CommandLine/Parsing/StringExtensions.cs
```diff
@@ -566,7 +566,8 @@ IEnumerable<string> SplitLine(string line)
 
             for (var commandAliasIndex = 0; commandAliasIndex < command.RawAliases.Count; commandAliasIndex++)
             {
-                var commandAlias = command.RawAliases[commandAliasIndex];
+                var commandAlias = command.RawAliases.ToArray()[commandAliasIndex];
```
Keboo, line 569: Any reason to not use ElementAt here?
```suggestion
                var commandAlias = command.RawAliases.ElementAt(commandAliasIndex);
```

src/System.CommandLine/Parsing/StringExtensions.cs
```diff
@@ -579,7 +580,7 @@ IEnumerable<string> SplitLine(string line)
 
                     for (var childAliasIndex = 0; childAliasIndex < child.RawAliases.Count; childAliasIndex++)
                     {
-                        var childAlias = child.RawAliases[childAliasIndex];
+                        var childAlias = child.RawAliases.ToArray()[childAliasIndex];
```
Keboo, line 583: Any reason to not use ElementAt here?
```suggestion
                        var childAlias = child.RawAliases.ElementAt(childAliasIndex);
```

src/System.CommandLine/Option.cs
```diff
@@ -56,7 +56,7 @@ public virtual Argument Argument
 
         private protected override void ChooseNameForUnnamedArgument(Argument argument)
         {
-            argument.Name = Aliases[0].ToLower();
+            argument.Name = Name.ToLower();
```
Keboo, line 59: This is a subtle difference here. Assuming an explicit name is not specified, it appears that going to `Name` will switch to using the longest alias rather than the first one. Should we add another test to cover this behavior?

src/System.CommandLine/Parsing/StringExtensions.cs
jonsequitur: None at all. A coupe of these were stabs at moving to `for` loops that dead ended at the `HashSet`.

src/System.CommandLine/Option.cs
jonsequitur: I'm not too worried about testing emergent behaviors until we're out of beta. If someone was relying on this it will break them but I suspect the new behavior is more robust in any case.

PR #980, opened 2020-07-17
Keep newlines in descriptions

### LINKED ISSUE #960: Multiline description using help command
Hello there! 
-- ~~General Kenobi!~~ 

I have the following C# code:

```cs
                  {
                      new Option<bool>(new[] {"-d", "--debug"}, "Debug mode. Provides additional logging information while running."),
                      new Option<bool>(new[] {"-wi", "--what-if"}, "What-if mode. It provides additional logging information while running but not performs real actions.")
                  };

rootCommand.Description = "This application does blah-blah." +
                          "\nAnd bleh-bleh too." +
                          "\n\nAnd has a bunch of restrictions, like:" +
                          "\n\t- Bluh-Bluh;" +
                          "\n\t- Blih-Blih;" +
                          "\n\t- Bloh-Bloh.";

rootCommand.Handler = CommandHandler.Create<bool, bool>(SomeCoolRunProgramMethodThatIsNotNeededForThisIssue);
```

And if I call `help` like this:
```
picolino@pc C:\MyCoolSourceFolder\ThatCoolApp\AwesomeApp.exe -?
```

I get that:
```
AwesomeApp:
  This application does blah-blah. And bleh-bleh too. And has a bunch of restrictions, like: - Bluh-Bluh; - Blih-Blih; - Bloh-Bloh.

Usage:
  AwesomeApp [options]

Options:
  -d, --debug       Debug mode. Provides additional logging information while running.
  -wi, --what-if    What-if mode. It provides additional logging information while running but not performs real actions.
  --version         Show version information
  -?, -h, --help    Show help and usage information
```

So the problem is I can't use a multiline description when using `help` command.
Is that a bug or classic "Microsoft feature"? Any workarounds for needed behavior?

Appreciate for any help 🤗

NextTurn:
Trying to resolve #960. Shall I change the global behaviors or add an option somewhere?

### DIFF
--- a/src/System.CommandLine.Tests/ApprovalTests/Help/Approvals/HelpBuilderTests.Help_describes_default_values_for_complex_root_command_scenario.approved.txt
+++ b/src/System.CommandLine.Tests/ApprovalTests/Help/Approvals/HelpBuilderTests.Help_describes_default_values_for_complex_root_command_scenario.approved.txt
@@ -18,4 +18,7 @@ Options:
   -troda, --the-root-option-default-arg <the-root-option-arg>                                            the-root-option-default-arg-description [default: the-root-option-arg-value]
   -troea, --the-root-option-enum-arg <Read|ReadWrite|Write>                                              the-root-option-description [default: Read]
   -trorea, --the-root-option-required-enum-arg <Read|ReadWrite|Write> (REQUIRED)                         the-root-option-description [default: Read]
+  -tromld, --the-root-option-multi-line-description                                                      the-root-option
+                                                                                                         multi-line
+                                                                                                         description
 
--- a/src/System.CommandLine.Tests/ApprovalTests/Help/HelpBuilderTests.Approval.cs
+++ b/src/System.CommandLine.Tests/ApprovalTests/Help/HelpBuilderTests.Approval.cs
@@ -70,6 +70,9 @@ namespace System.CommandLine.Tests.Help
                         Description = "the-root-option-arg-description",
                     },
                     IsRequired = true
+                },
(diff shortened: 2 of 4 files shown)


picolino:
I think it must be a global behavior change.  
Because it is unpredictable that the system will remove `\n` or `\r\n` characters from description line.

jonsequitur:
Thanks for working on this, @NextTurn. 

I agree with @picolino that it should just be a global behavior. It keeps things simple and it means we have fewer combinations to test.

Speaking of tests, a test that's worth adding is that both `\r\n` and `\n` should be replaced regardless of the current value of `Environment.NewLine`. Someone might create help using string literals or resource files resulting in `\n` being evaluated on Windows or `\r\n` being evaluated on Linux.

NextTurn:
Almost there.

jonsequitur:
@NextTurn, could you please add an option or argument to the [ApprovalTest case]( https://github.com/jonsequitur/command-line-api/blob/9cdcb230ac5f111aa8b602ed12e62cef458fa1cb/src/System.CommandLine.Tests/ApprovalTests/Help/HelpBuilderTests.Approval.cs#L14-L14) and update the [associated approval file](https://github.com/jonsequitur/command-line-api/blob/96da131652807fb37a571aba20a4eea73e718e1c/src/System.CommandLine.Tests/ApprovalTests/Help/Approvals/HelpBuilderTests.Help_describes_default_values_for_complex_root_command_scenario.approved.txt#L1-L1).

jonsequitur (approved):

jonsequitur:
Thanks, @NextTurn!

mristin:
Hi @jonsequitur ,
Do you know when is the next release planned  (even alpha or beta) which will include this feature?

PR #982, opened 2020-07-20
Nit: Override Write(string?) in StandardStreamWriter

NextTurn:
`TextWriter` uses `string.ToCharArray` which is slow.

Besides, this is the only overload that is exposed through `IStandardStreamWriter`.


jonsequitur:
Thanks, @NextTurn!

PR #988, opened 2020-07-24
Extend CommandHandler to handle up to 11 arguments

### LINKED ISSUE #983: Introduce more than 7 arguments in `System.CommandLine.Invocation.CommandHandler.Create`
Hi,
I noticed that the `CommandHandler.Create` is limited to at most 7 arguments of the action. This corresponds to at most 7 options, which is a bit too little in a more complex applications.

Was there a reason for this limit (instead of pushing it higher to, say, 16)? What is your suggested way to handle more than 7 options?

Thanks a lot for the answers in advance! Please let me know if I should create a pull request to introduce a higher limit if there is no particular reason for the current one.

mristin:
This patch fixes `System.CommandLine.Invocation.CommandHandler.Create`
so that it can handle actions with up to 11 arguments. Current limit of
7 arguments is not enough for a more complex applications.

While users can alternatively bind arguments to arbitrary number of
properties, this requires refactoring the code once the limit is
hit. Additionally, some users are not aware of the feature and could
hit the wall unintentionally. With this patch, the arbitrary, but
reasonable limit of 11 arguments should cover most of the use cases
"in the wild".

Fix #983.

### DIFF
--- a/src/System.CommandLine/Invocation/CommandHandler.cs
+++ b/src/System.CommandLine/Invocation/CommandHandler.cs
@@ -46,6 +46,22 @@ namespace System.CommandLine.Invocation
             Action<T1, T2, T3, T4, T5, T6, T7> action) =>
             HandlerDescriptor.FromDelegate(action).GetCommandHandler();
 
+        public static ICommandHandler Create<T1, T2, T3, T4, T5, T6, T7, T8>(
+            Action<T1, T2, T3, T4, T5, T6, T7, T8> action) =>
+            HandlerDescriptor.FromDelegate(action).GetCommandHandler();
+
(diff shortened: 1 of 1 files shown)


jonsequitur:
Thanks, @mristin!

caccorsi:
Hi, does anyone have a small example of how to leverage this?  I see that a correction was merged to main long ago, and I'm using the latest AFAIK (2.0.0-beta4.22272.1) but I'm still only able to use 7 arguments max. I see there is a restructuring of the library. If this correction is not part of a release, if you would be so kind and publish a release having latest corrections for the existing architecture it would be most appreciated.

PR #999, opened 2020-07-28
Fix possible OverflowException

### LINKED ISSUE #997: Arithmetic overflow exception thrown during call for suggestions
I'm still working on a minimal reproducible example:

```bash
$ enuf lookup attrUnhandled exception. System.AggregateException: One or more errors occurred. (Arithmetic operation resulted in an overflow.)
 ---> System.OverflowException: Arithmetic operation resulted in an overflow.
   at System.Linq.Enumerable.Sum[TSource](IEnumerable`1 source, Func`2 selector)
   at System.CommandLine.Parsing.SymbolResult.MaximumArgumentCapacity()
   at System.CommandLine.Parsing.SymbolResult.get_RemainingArgumentCapacity()
   at System.CommandLine.Parsing.SymbolResult.get_IsArgumentLimitReached()
   at System.CommandLine.Parsing.ParseResultExtensions.GetSuggestions(ParseResult parseResult, Nullable`1 position)
   at System.CommandLine.Invocation.SuggestDirectiveResult.Apply(InvocationContext context)
   at System.CommandLine.Invocation.InvocationPipeline.GetResultCode(InvocationContext context)
   at System.CommandLine.Invocation.InvocationPipeline.InvokeAsync(IConsole console)
   at System.CommandLine.CommandExtensions.InvokeAsync(Command command, String[] args, IConsole console)
   --- End of inner exception stack trace ---
   at System.Threading.Tasks.Task`1.GetResultCore(Boolean waitCompletionNotification)
   at System.Threading.Tasks.Task`1.get_Result()
   at Enuf.Program.Main(String[] args) in /Users/mgwelch/Documents/repos/enuf/enuf.cli/Program.cs:line 93
```

michaelgwelch:
Reduce maximum possible arity of an arg to 100,000

To fix #997 (overflow exception when calculating
remainging capcity of arguments) this places an arbitray
cap on the maximum arity of an argument. In this commit
this was set to 100,000.

This is enforced in the constructor, and by removing the
setters on MinimumNumberOfValues and MaximumNumberOfValues.


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/command-line-api?pullRequest=999) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/command-line-api?pullRequest=999) before we can accept your contribution.<br/><br/>:x: michaelgwelch [sign now](https://cla.dotnetfoundation.org/dotnet/command-line-api?pullRequest=999)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/command-line-api?pullRequest=999) it.</sub>

michaelgwelch:
I don't understand all the failed checks. Does this indicate an issue with the pipeline(s)?

michaelgwelch:
Looks like some of the tests dealing with Arity and MaxCapacity may have been impacted. I should have run all tests before opening PR. I'll investigate.

michaelgwelch:
Well, Duh, there's a very stupid mistake in my code. Will fix.

src/System.CommandLine.Tests/SymbolResultTests.cs
```diff
@@ -9,6 +9,36 @@ namespace System.CommandLine.Tests
 {
     public class SymbolResultTests
     {
+
```
michaelgwelch, line 12: I'm not sure this is the best class for this test. The defect was in SymbolResult but exposed via GetSuggestions.

michaelgwelch:
If reviewers have suggestions for

1. Best test class to target
1. A better setup that you think can trigger the defect with less setup and is easier to understand
1. A better fix idea

I'll gladly update the PR and tests.

jonsequitur:
It wasn't immediately clear to me why this triggers with suggestions but not parsing, but `ParserTests` seems like a reasonable place for the test. 

Another possible fix would be to use a significantly lower value than `int.MaxValue` for the default upper arity for collection types.

michaelgwelch:
> Another possible fix would be to use a significantly lower value than int.MaxValue for the default upper arity for collection types.

I started down that path. Would you be comfortable having this lower value capped in the constructor for ArgumentArity and the MaximumNumberOfValues property setter as well? If not, someone could just pass those in and we have the same issue.

I've started to do the research on what a sane value would be. Here's info from POSIX world: https://www.in-ulm.de/~mascheck/various/argmax/#maximum_number

This [article](https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessa?redirectedfrom=MSDN) seems to indicate 32767 for a maximum length for command line. 

I had gone ahead and set the default max number of arguments to 100,000 which seems sufficiently large yet small enough not to overflow. (Assuming no one could set it significantly higher thru constructor or property setter)

jonsequitur:
I think setting the cap at some arbitrary value makes sense. 100k arguments ought to be enough for anybody. 😄

michaelgwelch:
I will redo the commit taking that approach.

michaelgwelch:
> 100k arguments ought to be enough for anybody.

based on some pretty famous words.

src/System.CommandLine/ArgumentArity.cs
```diff
@@ -21,13 +23,18 @@ public ArgumentArity(int minimumNumberOfValues, int maximumNumberOfValues)
...
+
             MinimumNumberOfValues = minimumNumberOfValues;
             MaximumNumberOfValues = maximumNumberOfValues;
         }
 
-        public int MinimumNumberOfValues { get; set; }
+        public int MinimumNumberOfValues { get; }
 
-        public int MaximumNumberOfValues { get; set; }
+        public int MaximumNumberOfValues { get; }
```
michaelgwelch, line 37: I wasn't sure if there was a good reason for making this class mutable. Obviously the setters could be put back and just add checking. I noticed that the other constraints from the constructor were missing on the setters and just decided they were unlikely to be used and removed them

src/System.CommandLine/Parsing/SymbolResult.cs
```diff
@@ -44,9 +44,12 @@ Parent switch
...
             MaximumArgumentCapacity() - Tokens.Count;
 
-        internal int MaximumArgumentCapacity() =>
-            Symbol.Arguments()
-                  .Sum(a => a.Arity.MaximumNumberOfValues);
+        internal int MaximumArgumentCapacity()
+        {
+            var maximumArgumentCapacity = Symbol.Arguments()
+                .Sum(a => a.Arity.MaximumNumberOfValues);
+            return (int) Math.Min(maximumArgumentCapacity, int.MaxValue);
```
michaelgwelch, line 51: Arguably, this change could be removed; but theoretically someone could go out of their way to cause an overflow.

I'm fine removing this change if you think it's ridiculous to keep it around.

michaelgwelch:
@jonsequitur This PR has been updated with the suggested change.

src/System.CommandLine/ArgumentArity.cs
jonsequitur: Yeah, I'm all for making it immutable.

PR #1004, opened 2020-08-03
Added missing 'string[] args' to Main method in code samples

toxicgorilla:
`args` parameter was missing in code samples


jonsequitur:
Thanks, @toxicgorilla!

PR #1009, opened 2020-08-05
Explicit parameter binding

KathleenDollard:
Allows differently named parameters and paired options/arguments


KathleenDollard:
Fixes #952

src/System.CommandLine/Binding/ModelBinder.cs
```diff
@@ -219,13 +212,12 @@ public void UpdateInstance<T>(T instance, BindingContext bindingContext)
                         {
                             if (parameterDescriptor.HasDefaultValue)
                                 boundValue = BoundValue.DefaultForValueDescriptor(parameterDescriptor);
-                            else if (parameterDescriptor.AllowsNull && 
+                            else if (parameterDescriptor.AllowsNull &&
                                 ShouldPassNullToConstructor(constructorDescriptor.Parent, constructorDescriptor))
                                 boundValue = BoundValue.DefaultForType(valueDescriptor);
```
jonsequitur, line 217: It looks like this block should also be part of the extracted `GetBoundValue` method.

src/System.CommandLine/Invocation/InvocationExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+namespace System.CommandLine.Invocation
+{
+    public static class InvocationExtensions
+    {
+        public static void BindParameter(this ICommandHandler handler, ParameterInfo param, Option option)
+        {
+            // check for nulls
+            if (!(handler is ModelBindingCommandHandler bindingHandler))
+            {
+                throw new InvalidOperationException("Cannot bind to this type of handler");
```
jonsequitur, line 12: Maybe we should just make `ModelBindingCommandHandler` public?

KathleenDollard:
Replaced with #1018

Issue #872, opened 2020-04-20
Create a Roslyn analyzer to enforce matching argument/option names with command handler parameter names


jonsequitur:
The most common issue people encounter with the System.CommandLine API is when option or argument names/aliases do not match the parameters in a command handler:

```csharp
var command= new RootCommand
{
    new Option<int>("--aaa")
};
command.Handler = CommandHandler.Create<int>(zzz => /*  zzz is always going to be 0  */); 
```

We haven't seen an approach that enforces a relationship between the parameters and the parser at compile time without an enormous increase in the complexity of the code, so we took a more dynamic approach, based on ASP.NET MVC controller methods. The consequence is that these mismatches can't be flagged by the compiler, and until you're aware of the convention, it's easy to get this wrong.

Programmatic validation that all parameters are matchable on all command handlers at runtime has been proposed. This approach would be very expensive, especially when complex objects rather than scalar values are being bound, and since this won't vary after the code is compiled, it's not ideal to pay this cost every time a program runs. 

An analyzer might be a more appropriate tool for the job, since it only needs to be validated when the code is being written. It could also include quick fixes for renaming parameters to match options and arguments.

jeremyong:
> This approach would be very expensive, especially when complex objects rather than scalar values are being bound, and since this won't vary after the code is compiled, it's not ideal to pay this cost every time a program runs.

It seems reasonable to restrict the full validation to `Debug` builds if performance is a concern, or at the very least, implement a "light version" and "full version" of the reflection-based validation so that the light version kicks in at runtime. An option to disable validation (passed to the CommandHandler) may also make sense.

jeredm:
Another case to allow in an analyzer would be for any "magic" variables like `console`.
```
var addCommmand = new Command("add-item", "Adds a grocery item to the list")
{
    new Option<int>(new string[] { "--quantity", "-q" }, () => 1, "The number of items to add"),
    new Argument<string>("name", "The name of the grocery item")
};
addCommmand.Handler = CommandHandler.Create<IConsole, string, int>(async (console, name, quantity) =>
    await shoppingList.AddItemsAsync(console, name, quantity));
```

jonsequitur:
@jeremyong The reason we haven't approached this with something that only runs by default in debug mode is that this introduces a significant behavioral difference between debug and release builds. Unless someone is verifying both builds, it creates a possibility of shipping serious bugs.

jeremyong:
[...]

PR #1012, opened 2020-08-06
CommandHandler.Create without name matching

jonsequitur:
The approach I'm experimenting with here is to avoid the most common gotcha in the System.CommandLine API, which is the name-based matching of options/arguments to their corresponding parameters in a command handler.

Given:

```csharp
var stringOption = new Option<string>("--the-string");
var intOption = new Option<int>("--the-int");
var filesArg = new Argument<FileInfo[]>("the-files");
var command = new RootCommand
{
    stringOption,
    intOption,
    filesArg
};

```

Handler with name-based matching:

```csharp
command.Handler = 
    // You have to specify the types and they have to match the options/arguments:
    CommandHandler.Create<string, int, FileInfo[]>( 
        // The names must match some alias of the symbol:
        (theString, theInt, theFiles) => 
        {
            
        });
```

Handler with the new proposed API: 

```csharp
command.Handler = 
    // Types are inferred:
    CommandHandler.Create(
        stringOption, intOption, filesArg,
        // The names don't matter:
        (s, i, fs) => 
        {
            
        });
```

Pros:

* No more failed binding because of naming mismatches.
* Type inference means it's all strongly-typed without having to repeat information, and no worries about a type mismatch.

Cons / design problems to be solved:

* No current solution for complex model binding, since parameters map one-to-one with options/arguments, meaning we need a lot more overloads. (People have asked for this anyway since the complex model binding is non-obvious.)
* No current solution for service parameters, e.g. `IConsole`, `ParseResult`, etc.

Please discuss.

### DIFF
--- a/src/System.CommandLine/Binding/IValueDescriptor.cs
+++ b/src/System.CommandLine/Binding/IValueDescriptor.cs
@@ -13,4 +13,9 @@ namespace System.CommandLine.Binding
 
         object? GetDefaultValue();
     }
+
+    public interface IValueDescriptor<T> : IValueDescriptor
+    {
+        // FIX: (IValueDescriptor) 
+    }
 }
--- a/src/System.CommandLine.Tests/Binding/ModelBindingCommandHandlerTests.BindingByName.cs
+++ b/src/System.CommandLine.Tests/Binding/ModelBindingCommandHandlerTests.BindingByName.cs
@@ -1,0 +1,167 @@
+﻿// // Copyright (c) .NET Foundation and contributors. All rights reserved.
+// // Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.CommandLine.Binding;
+using System.CommandLine.Invocation;
+using System.CommandLine.IO;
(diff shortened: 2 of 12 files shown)


ohadschn:
The [implementation](https://github.com/dotnet/command-line-api/pull/1012/files#diff-3f83ebde8e030ac48edd1f0b478ef766R144) looks exactly like my workaround [here](https://github.com/dotnet/command-line-api/issues/872#issuecomment-658862025) so I'm definitely on board.

src/System.CommandLine/Binding/IValueDescriptor.cs
```diff
@@ -13,4 +13,9 @@ public interface IValueDescriptor
 
         object? GetDefaultValue();
     }
+
+    public interface IValueDescriptor<T> : IValueDescriptor
+    {
+        // FIX: (IValueDescriptor)
```
Keboo, line 19: Does this needs to be addressed?

Keboo (approved):

PR #1013, opened 2020-08-06
Add Localization Features package for System.CommandLine

fredrikhr:


src/System.CommandLine.Localization/LocalizedHelpBuilder.cs
```diff
@@ -47,7 +82,8 @@ private Option GetLocalizedOption(IOption option)
             Localize(lopt, option);
 
             lopt.Name = option.Name;
-            lopt.Argument = GetLocalizedArgument(option.Argument);
+            if (!(option.Argument.Arity is { MaximumNumberOfValues: 0, MinimumNumberOfValues: 0 }))
+                lopt.Argument = GetLocalizedArgument(option.Argument);
```
jonsequitur, line 86: Nit: Curly braces please.

fredrikhr: 46fe1fb

samples/LocalizationPlayground/CultureEnvironmentCommandLineExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+        internal static CommandLineBuilder UseCultureEnvironment(
+            this CommandLineBuilder builder)
+        {
+            return builder.UseMiddleware(async (context, next) =>
+            {
+                if (Environment.GetEnvironmentVariable("DOTNET_SYSTEM_GLOBALIZATION_CULTURE") is string culture &&
+                        !string.IsNullOrEmpty(culture))
+                    CultureInfo.CurrentCulture = CultureInfo.CurrentUICulture = CultureInfo.GetCultureInfo(culture);
+
+                var execCtx = ExecutionContext.Capture();
```
jonsequitur, line 20: Is there an approach that can work without using `ExecutionContext`? It's fairly heavy in terms of performance.

fredrikhr: Yeah, this is just a quick and dirty hack to achieve this in the playground, but I am using the same trick in #951, so we should discuss it there... When we merge #951 this won't be necessary anyways. In the meantime is it okay to just keep this here, since it's just the playground?

fredrikhr: @jonsequitur I improved it slightly in 8587a5c3

samples/LocalizationPlayground/Properties/launchSettings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "profiles": {
+    "LocalizationPlayground": {
+      "commandName": "Project",
+      "commandLineArgs": "[env:DOTNET_SYSTEM_GLOBALIZATION_CULTURE=de-DE] -c 3 .NET"
```
wli3, line 5: This env should not be needed

wli3: There should be a .net core leveled env to do the same thing

jonsequitur: @wli3  We were hoping to align to whatever environment variables exist already in the SDK but the only one I could find was for specifying the invariant culture, not for specifying other cultures. https://docs.microsoft.com/en-us/dotnet/core/run-time-config/globalization

wli3: my memory is wrong. https://github.com/dotnet/sdk/blob/b1223209644d900702287faea8e9b71f95ec49f8/src/Cli/dotnet/UILanguageOverride.cs#L38 

we had a CLI specific flag. If we need a flag for this library that's fine.

wli3:
I don't think xlf is generated by Arcade. We need to do that so it can fit into existing localization system.

But that's also a headache, we would require the contributor to run a build with arcade (which is disabled in "dev mode") and then checkin the new xlfs. Maybe the "right way" is we could have a bot to run and append a checkin for PRs touching resx files.

jonsequitur:
Related: #1125

PR #1018, opened 2020-08-09
Explicit binding 4

KathleenDollard:
Significant updates to binding so method binding and model binding use the same logic and same code paths. 

Draft due to rebase issues.


src/System.CommandLine.Tests/Binding/ModelBinderConstructorTests.cs
```diff
@@ -0,0 +1,143 @@
...
+            }
+
+            public object BoundValue { get; }
+
+            public void Apply(InvocationContext context)
+            {
+            }
+        }
+
+        private class ClassForCaptureMethod<T>
```
jonsequitur, line 119: A clearer name for this might be something like `TestCommandHandler<T>`.

src/System.CommandLine.Tests/Binding/ModelBinderConstructorTests.cs
```diff
@@ -0,0 +1,143 @@
...
+        //[InlineData(typeof(ClassWithSetter<int>))]
+        //[InlineData(typeof(ClassWithCtorParameter<string>))]
+        //[InlineData(typeof(ClassWithSetter<string>))]
+        [InlineData(typeof(FileInfo))]
+        [InlineData(typeof(FileInfo[]))]
+        [InlineData(typeof(string[]))]
+        [InlineData(typeof(List<string>))]
+        [InlineData(typeof(int[]))]
+        [InlineData(typeof(List<int>))]
+        public async Task Constructor_receives_option_arguments_bound_to_the_specified_type(
```
jonsequitur, line 63: Maybe `Model_constructor_receives_option_arguments_bound_to_the_specified_type`?

src/System.CommandLine.Tests/Binding/ModelBinderConstructorTests.cs
```diff
@@ -0,0 +1,143 @@
...
+
+            public void Invoke() { }
+        }
+
+        private class ClassForCreateBase
+        {
+            public object? Value { get; protected set; }
+
+        }
+        private class ClassForCreate<T> : ClassForCreateBase
```
jonsequitur, line 134: Maybe `TestModel` and `TestModel<T>`?

src/System.CommandLine/Binding/BindingContext.cs
```diff
@@ -63,6 +63,15 @@ public ModelBinder GetModelBinder(IValueDescriptor valueDescriptor)
             return new ModelBinder(valueDescriptor);
         }
 
+        internal ModelBinder GetModelBinder(Type type)
+        {
+            if (_modelBindersByValueDescriptor.TryGetValue(type, out ModelBinder binder))
+            {
+                return binder;
+            }
+            return new ModelBinder(type);
```
jonsequitur, line 72: Is it deliberate that we don't cache this?

src/System.CommandLine/Binding/BoundValue.cs
```diff
@@ -5,6 +5,7 @@ namespace System.CommandLine.Binding
 {
     public class BoundValue
     {
+        // ?? Why have an internal constructor on a public readonly class?
```
jonsequitur, line 8: We don't want people to create instances of this.

src/System.CommandLine/Binding/ServiceProviderValueSource.cs
```diff
@@ -5,12 +5,12 @@ namespace System.CommandLine.Binding
...
-        public bool TryGetValue(
-            IValueDescriptor valueDescriptor,
-            BindingContext? bindingContext,
-            out object? boundValue)
+        public bool TryGetValue(IValueDescriptor valueDescriptor,
+                                BindingContext? bindingContext,
+                                out object? boundValue)
         {
             boundValue = bindingContext?.ServiceProvider.GetService(valueDescriptor.ValueType);
+            // ?? Why return true if the service isn't found?
```
jonsequitur, line 13: This is because `null` is an acceptable return value for `ServiceProvider.GetService`.

src/System.CommandLine/Invocation/InvocationExtensions.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System.Reflection;
+
+namespace System.CommandLine.Invocation
+{
+    public static class InvocationExtensions
+    {
+        public static void BindParameter(this ICommandHandler handler, ParameterInfo param, Option option)
```
jonsequitur, line 7: These shouldn't be necessary, correct?

src/System.CommandLine/Binding/BindingContext.cs
KathleenDollard: Method removed as dead code

src/System.CommandLine.Tests/Binding/ModelBinderConstructorTests.cs
KathleenDollard: Done

src/System.CommandLine/Binding/BoundValue.cs
KathleenDollard: It does not appear that BoundValue escape the context of System.CommandLine. I think we should deliberately consider the needs of any application models that need something other than to connect a ValueSource to a Option/Argument ahead of any evaluation. IOW, we may need this public in the future, but it does not appear that we need it public now. 

This is probably part of a broader discussion of checking our public API surface.

src/System.CommandLine/Binding/ServiceProviderValueSource.cs
KathleenDollard: OK.

This led to me reviewing how we are using the return value in other ValueSources, and I think we should separately look at the name "ParseResultMatchingValueSource". It seems a name closer to MatchByName might be helpful

src/System.CommandLine/Invocation/InvocationExtensions.cs
KathleenDollard: I think it is needed, although it might not be in the right place. That extension method drives the following syntax: 

```
            handler.BindParameter(parameter, argument);
            command.Handler = handler;
```

The handler is managed as an ICommandHandler, so putting the method on the interface isn't sensible due to the ParameterInfo/PropertyInfo. Thus, an extension method seemed the right approach.

src/System.CommandLine/Binding/ServiceProviderValueSource.cs
jonsequitur: Yeah, that seems helpful in the name, as is the fact that it matches from within the parse results, not e.g. the service provider.

src/System.CommandLine/Invocation/InvocationExtensions.cs
KathleenDollard: Or were you just expecting to always do it on the argument and this is on the option?

jonsequitur: I think making `ModelBindingCommandHandler` public makes more sense than extending the interface but throwing if it's the wrong implementation.

PR #1020, opened 2020-08-13
adds ParseResult parameter for suggestions

### LINKED ISSUE #1019: Access to ParseResult when providing suggestions
I try to implement suggestions in a tool of mine, where one parameter is dependent on another. Imagine following scenarios:
- Add a user to group
  `addusertogroup -user <user> -group <group>`
  Suggestions for `group` should only only include groups the user `user` is not already member of.
- Delete an Azure resource from a resource group
  `deleteresource -resourcegroup <resouregroup> -resource <resource>`
  Suggestions for `resource` should only include resources in the specified `resouregroup`.

Unfortunately, I can't see a way to access already passed parameters when providing suggestions. Do I miss something?

I've implemented a proof-of-concept to pass `ParseResult` to get suggestions. However, this is a **breaking change**. E.g.
```diff
- public delegate IEnumerable<string> SuggestDelegate(string? textToMatch);
+ public delegate IEnumerable<string> SuggestDelegate(ParseResult? parseResult, string? textToMatch);
```
You can see my complete PoC at https://github.com/thoemmi/command-line-api/commit/91aa52baf816cc609d15626f9695ebaea043e91c.

Do you see an simpler solution?

thoemmi:
This PR adds support for accessing other options when providing suggestions. Imagine following scenarios:
- Add a user to group
  `addusertogroup -user <user> -group <group>`
  Suggestions for `group` should only only include groups the user `user` is not already member of.
- Delete an Azure resource from a resource group
  `deleteresource -resourcegroup <resouregroup> -resource <resource>`
  Suggestions for `resource` should only include resources in the specified `resouregroup`.

When implementing suggestion, you won't get the `textToParse` only, but also `ParseResult`, which allows access to all options parsed so far.

**Please note that this is a breaking change.**

This fixes #1019

### DIFF
--- a/src/System.CommandLine.Tests/SuggestDirectiveTests.cs
+++ b/src/System.CommandLine.Tests/SuggestDirectiveTests.cs
@@ -25,7 +25,7 @@ namespace System.CommandLine.Tests
                 .AddSuggestions("apple", "banana", "cherry");
 
             _vegetableOption = new Option<string>("--vegetable")
-                .AddSuggestions(_ => new[] { "asparagus", "broccoli", "carrot" });
+                .AddSuggestions("asparagus", "broccoli", "carrot");
 
             _eatCommand = new Command("eat")
             {
--- a/src/System.CommandLine.Tests/SuggestDirectiveTests.cs
+++ b/src/System.CommandLine.Tests/SuggestDirectiveTests.cs
@@ -25,7 +25,7 @@ namespace System.CommandLine.Tests
                 .AddSuggestions("apple", "banana", "cherry");
 
             _vegetableOption = new Option<string>("--vegetable")
-                .AddSuggestions("asparagus", "broccoli", "carrot");
+                .AddSuggestions((_, __) => new[] { "asparagus", "broccoli", "carrot" });
 
             _eatCommand = new Command("eat")
             {
--- a/src/System.CommandLine.Tests/SuggestionTests.cs
+++ b/src/System.CommandLine.Tests/SuggestionTests.cs
@@ -137,7 +137,7 @@ namespace System.CommandLine.Tests
                 new Command("andmyothersubcommand"),
             };
 
-            var suggestions = command.GetSuggestions(null, "my");
+            var suggestions = command.GetSuggestions("my");
 
             suggestions.Should().BeEquivalentSequenceTo("mysubcommand", "andmyothersubcommand", "andmythirdsubcommand");
         }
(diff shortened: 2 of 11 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/command-line-api?pullRequest=1020) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/command-line-api?pullRequest=1020) before we can accept your contribution.<br/><br/>:x: thoemmi [sign now](https://cla.dotnetfoundation.org/dotnet/command-line-api?pullRequest=1020)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/command-line-api?pullRequest=1020) it.</sub>

src/System.CommandLine.Tests/SuggestDirectiveTests.cs
```diff
@@ -25,7 +25,7 @@ public SuggestDirectiveTests()
                 .AddSuggestions("apple", "banana", "cherry");
 
             _vegetableOption = new Option<string>("--vegetable")
-                .AddSuggestions(_ => new[] { "asparagus", "broccoli", "carrot" });
+                .AddSuggestions("asparagus", "broccoli", "carrot");
```
jonsequitur, line 28: Although it could be a lot more apparent, this is deliberately using the delegate-based approach as opposed to the `params` approach on line 25. So please revert this in case it represents a loss of scenario coverage.

src/System.CommandLine.Tests/SuggestionTests.cs
```diff
@@ -137,7 +137,7 @@ public void Command_Suggest_with_text_to_match_orders_by_match_position_then_alp
                 new Command("andmyothersubcommand"),
             };
 
-            var suggestions = command.GetSuggestions("my");
+            var suggestions = command.GetSuggestions(null, "my");
```
jonsequitur, line 140: We can avoid the need to do this by overloading `GetSuggestions` via an extension method. This way we're not introducing a source-breaking change.

thoemmi: Fair enough. Would it be ok to make `SymbolExtensions` public and keep the existing extension methods as internal?

src/System.CommandLine.Tests/SuggestDirectiveTests.cs
thoemmi: Ok. Only this one, or should I revert all my changes from delegate to `params`?

jonsequitur:
Thanks, @thoemmi!

PR #1026, opened 2020-08-21
Explicit parameter binding and other binding updates

KathleenDollard:
Significant updates to binding so method binding and model binding use the same logic and same code paths.


src/System.CommandLine.Tests/Binding/ModelBinderConstructorTests.cs
```diff
@@ -0,0 +1,143 @@
...
+using Xunit;
+using common = System.CommandLine.Tests.Binding.ModelBindingCommandHandlerTests;
+
+namespace System.CommandLine.Tests.Binding
+{
+    public class ModelBinderConstructorTests
+    {
+
+        [Theory]
+        //[InlineData(typeof(ClassWithCtorParameter<int>))]
```
jonsequitur, line 23: Rather than  comment these out can you add the `Skip` property to them? This way they'll show up in test results and be easier to not to lose track of.

src/System.CommandLine.Tests/Binding/ModelBinderTests.cs
```diff
@@ -79,6 +79,39 @@ public class ModelBinderTests
...
+
+            object valueReceivedValue = ((dynamic)instance).Value;
+            var expectedValue = new FileInfo(commandLine);
+
+            valueReceivedValue.Should().BeOfType<FileInfo>();
+            var fileInfoValue = valueReceivedValue as FileInfo;
+            fileInfoValue.FullName.Should().Be(expectedValue.FullName);
+            // The following fails when it attempts to compare the Length of the file. I have
+            // no idea why this previously worked. 
+            //valueReceivedValue.Should().BeEquivalentTo(expectedValue);
```
jonsequitur, line 112: Doing a property-by-property comparison on a `FileInfo` is probably not useful. I would just compare the properties that matter.

KathleenDollard: I just dropped that line of code. I was hesitant to just throw away a test, but the presence of the file is not relevant to this test.

src/System.CommandLine/Binding/ModelBinder.cs
```diff
@@ -37,208 +26,210 @@ internal ModelBinder(IValueDescriptor valueDescriptor)
...
+
+            static bool IsNullable(Type type)
+            {
+                return type.IsGenericType &&
+                       type.GetGenericTypeDefinition() == typeof(Nullable<>);
+            }
+        }
 
-                    instance = fromModelBinder;
+        private (bool success, object? newInstance, bool anyNonDefaults) GetSimpleModelValue(
```
Keboo, line 115: It appears that the `success` and `anyNonDefaults` always have the same value. Would it be cleaner to have this be a bool returning method with an out parameter for the `newInstance`?

src/System.CommandLine/Binding/ModelBinder.cs
```diff
@@ -37,208 +26,210 @@ internal ModelBinder(IValueDescriptor valueDescriptor)
...
 
-            return values;
+            return (values, anyNonDefaults);
         }
 
-        private IValueSource GetValueSource(
-            IDictionary<IValueDescriptor, IValueSource>? bindingSources,
-            BindingContext bindingContext,
-            IValueDescriptor valueDescriptor)
+        internal static IValueSource GetValueSource(IDictionary<IValueDescriptor, IValueSource>? bindingSources,
```
Keboo, line 229:
```suggestion
        private static IValueSource GetValueSource(IDictionary<IValueDescriptor, IValueSource>? bindingSources,
```

src/System.CommandLine/Binding/ModelBinder.cs
```diff
@@ -261,21 +252,88 @@ public void UpdateInstance<T>(T instance, BindingContext bindingContext)
...
+                    return (new BoundValue(parameterDescriptor.GetDefaultValue(), valueDescriptor, valueSource), false);
+                }
+                // Logic dropped here - misnamed and purpose unclear: ShouldPassNullToConstructor(constructorDescriptor.Parent, constructorDescriptor))
+                return (BoundValue.DefaultForType(valueDescriptor), false);
+            }
+            return (null, false);
+        }
+
+
+        protected ConstructorDescriptor FindModelConstructorDescriptor(ConstructorInfo constructorInfo)
```
Keboo, line 298:
```suggestion
        private ConstructorDescriptor FindModelConstructorDescriptor(ConstructorInfo constructorInfo)
```

src/System.CommandLine/Binding/ModelBinder.cs
```diff
@@ -37,208 +26,210 @@ internal ModelBinder(IValueDescriptor valueDescriptor)
...
-        }
-
-        public void BindConstructorArgumentFromValue(ParameterInfo parameter,
-            IValueDescriptor valueDescriptor)
-        {
-            if (!(parameter.Member is ConstructorInfo constructor))
-                throw new ArgumentException(paramName: nameof(parameter),
-                    message: "Parameter must be declared on a constructor.");
-
+            var constructor = FindConstructorOrThrow(parameter, "Parameter must be declared on a constructor.");
```
Keboo, line 33:
```suggestion
            var constructor = parameter.Member as ConstructorInfo ?? throw new ArgumentException(paramName: nameof(parameter),
                      message: "Parameter must be declared on a constructor.");
```

This also allows the removal of `FindConstructorOrThrow`

src/System.CommandLine/Binding/ModelBinder.cs
```diff
@@ -293,4 +351,18 @@ public AnonymousValueDescriptor(Type modelType)
             public override string ToString() => $"{ValueType}";
         }
     }
+
+    internal class ConstructorAndArgs
```
Keboo, line 355: This looks like it could be made private and nested within `ModelBinder`

src/System.CommandLine/Binding/ModelDescriptor.cs
```diff
@@ -35,7 +35,7 @@ protected ModelDescriptor(Type modelType)
         public IReadOnlyList<IValueDescriptor> PropertyDescriptors =>
             _propertyDescriptors ??=
                 ModelType.GetProperties(CommonBindingFlags)
-                         .Where(p => p.CanWrite)
+                         .Where(p => p.CanWrite && p.SetMethod.IsPublic)
```
Keboo, line 38: Why limit it to only settings properties that have public setters?

src/System.CommandLine/Binding/ParameterDescriptor.cs
```diff
@@ -32,22 +32,16 @@ public bool AllowsNull
...
-                        _allowsNull = true;
-                    }
+                    _allowsNull = CalculateAllowsNull(_parameterInfo);
                 }
-
                 return _allowsNull ?? false;
             }
         }
 
+        public static bool CalculateAllowsNull(ParameterInfo parameterInfo)
```
Keboo, line 41:
```suggestion
        public static bool GetAllowsNull(ParameterInfo parameterInfo) 
```

src/System.CommandLine/Invocation/InvocationExtensions.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System.Reflection;
+
+namespace System.CommandLine.Invocation
+{
+    public static class InvocationExtensions
+    {
+        public static void BindParameter(this ICommandHandler handler, ParameterInfo param, Option option)
+        {
+            // check for nulls
```
Keboo, line 9: Is this a TODO to check incoming parameters for null?

src/System.CommandLine/Invocation/ModelBindingCommandHandler.cs
```diff
@@ -76,5 +76,34 @@ public async Task<int> InvokeAsync(InvocationContext context)
 
             return await CommandHandler.GetResultCodeAsync(result, context);
         }
+
+        internal void BindParameter(ParameterInfo param, Argument argument)
+        {
+            var _ = argument ?? throw new InvalidOperationException("You must specify an argument to bind");
```
Keboo, line 82:
```suggestion
            var _ = argument ?? throw new ArgumentNullException(nameof(argument));
```

src/System.CommandLine/Invocation/ModelBindingCommandHandler.cs
```diff
@@ -76,5 +76,34 @@ public async Task<int> InvokeAsync(InvocationContext context)
...
+            BindValueSource(param, new SpecificSymbolValueSource(argument));
+        }
+
+        internal void BindParameter(ParameterInfo param, Option option)
+        {
+            var _ = option ?? throw new InvalidOperationException("You must specify an option to bind");
```
Keboo, line 88:
```suggestion
            var _ = option ?? throw new ArgumentNullException(nameof(option));
```

src/System.CommandLine/Binding/ModelDescriptor.cs
jonsequitur: This is less surprising than if we were to set non-public members.

src/System.CommandLine/Binding/ParameterDescriptor.cs
jonsequitur: Or even just `AllowsNull`?

Keboo: I believe that conflicts with the property of the same name above.

KathleenDollard:
Closed in favor of #1032

Issue #1035, opened 2020-09-07
SymbolSet.GetByAlias does not work when alias is to a symbol after it's added to the parent symbol


jonsequitur:
![image](https://user-images.githubusercontent.com/547415/92411027-12bc2880-f0fb-11ea-8e22-6ef535d45da7.png)

Issue #340, opened 2018-12-18
After changing a subcommand name, AliasedSet Contains doesn't find it


KathleenDollard:
To reproduce...

Create a command with subcommands (children via AddCommand)
`command.Children.Contains(<childname>)` is true
Change the Name property of a child
`command.Children.Contains(<newChildName>)` is false

You can confirm the names of the children in the watch window. 

It appears that the keys in the Aliased Set's HashSet do not respond to the mutation.

A few of the options:

- Figure out a way to update the AliasSet when the name changes. I do not see a way to accomplish this
- Not allow names to be changed at all
- Only allow names to be changed via a method on the AliasSet

KathleenDollard:
Thinking more about this, we might consider a "freeze" of things when they are attached to other things. 

For me, I simply quit changing names, but we definitely want some guardrails here. 

I've worked around this, so this is not blocking me.

jonsequitur:
The current intent is that:

1) `Command`, `Option`, and `Argument` are mutable prior to a `CommandLineConfiguration` being created, and 
2) the `CommandLineConfiguration`, which exposes a hierarchy of `ICommand`, `IOption`, and `IArgument`, should be immutable. 

On the mutable side, changing a name should be reflected in the API. If a reference to these instances was held and they're then mutated, the `CommandLineConfiguration` should not be affected.

This issue is definitely a bug in regards to point 1. 

We've discussed cloning objects in order to enforce point 2 but this hasn't been implemented yet.

Issue #260, opened 2018-10-17
remove or improve RawAliases APIs


jonsequitur:
The "raw alias" concept is unclear. It can be clarified or removed.

From https://github.com/dotnet/System.CommandLine/pull/252#discussion_r225593830:
> I suspect RawAliases can become an internal implementation detail and we can collapse the HasAlias and HasRawAlias methods down to something like:
> 
> ```csharp 
> bool HasAlias(string alias, string prefix = null);
> ```
> I'm not sure how clear the intention is there. The gist is that HasAlias tends to be an end-user use case (i.e. don't make me think about prefixes when looking at parse results) and HasRawAlias tends to be an internal use-case (i.e. pre-compute all of the valid tokens and put them in a hashset for more efficient lookup during parse.)
>

PR #1040, opened 2020-09-14
Keep alias lookups up to date with additions and name changes 

jonsequitur:
This will fix the following issues:

* #260
* #340
* #1035

Changes include:

- [x] Remove `Argument.AddAlias` method, since Arguments don't really support aliases
- [x] Remove `RawAliases`
- [x] Change the meaning of `Aliases` to mean the literal aliases, including prefixes. They were only ever different for `Option`. The most important significance of this change can be seen in many of the tests which is that you can no longer query for an option without the prefix and get a positive result (e.g. given a parse result containing option `-x`, `HasOption("x")` now returns `false`, `CommandResult.ValueForOption("x")` won't find anything, etc.)
- [x] Add `Option.HasAliasIgnorePrefix`
- [x] Refactor and make perf improvements.


src/System.CommandLine/Command.cs
```diff
@@ -10,12 +10,29 @@
...
+        Symbol, 
+        ICommand, 
+        IEnumerable<Symbol>
     {
         private readonly SymbolSet _globalOptions = new SymbolSet();
 
-        public Command(string name, string? description = null) : base(new[] { name }, description)
+        public Command(string name, string? description = null) : base(description)
         {
+            Name = name;
```
mikem8361, line 22: Won't this add the alias in the Name property setter? Doesn't it make the below AddAlias redundant?

src/System.CommandLine/Command.cs
```diff
@@ -10,12 +10,29 @@
...
+        }
+
+        public override string Name
+        {
+            // FIX: (Name) 
+            get => base.Name;
+            set
+            {
+                base.Name = value;
+                AddAlias(Name);
```
mikem8361, line 34: Does the original base.Name need to be removed from the aliases? Like you do in the Options?

mikem8361: Never mind. It looks like it will get removed in the base (Symbol) class's Name property.

mikem8361 (approved):

src/System.CommandLine/DictionaryExtensions.cs
```diff
@@ -22,5 +22,36 @@ internal static class DictionaryExtensions
...
+            }
+        }
+
+        public static bool TryRemove<TKey, TValue>(
+            this IDictionary<TKey, TValue> source,
+            TKey key)
+        {
+            if (source.ContainsKey(key))
+            {
+                source.Remove(key);
```
Keboo, line 48: IDictionary.Remove already returns a bool. Is this method needed?

src/System.CommandLine/Option.cs
```diff
@@ -8,18 +8,34 @@
...
+        public Option(string[] aliases, string? description = null) : base(description)
         {
+            if (aliases is null)
+            {
+                throw new ArgumentNullException(nameof(aliases));
+            }
+
+            if (!aliases.Any())
+            {
+                throw new ArgumentException("An option must have at least one alias.");
```
Keboo, line 31:
```suggestion
                throw new ArgumentException("An option must have at least one alias.", nameof(aliases));
```

src/System.CommandLine/Symbol.cs
```diff
@@ -145,17 +122,38 @@ public bool HasAlias(string alias)
...
         ISymbolSet ISymbol.Children => Children;
+
+        internal Action<ISymbol>? OnNameOrAliasChanged;
+
+        [DebuggerStepThrough]
+        private protected void ThrowIfAliasIsInvalid(string alias)
+        {
+            if (string.IsNullOrWhiteSpace(alias))
+            {
+                throw new ArgumentException("An alias cannot be null, empty, or consist entirely of whitespace.");
```
Keboo, line 147:
```suggestion
                throw new ArgumentException("An alias cannot be null, empty, or consist entirely of whitespace.", nameof(alias));
```

src/System.CommandLine/Symbol.cs
```diff
@@ -145,17 +122,38 @@ public bool HasAlias(string alias)
...
+            }
+
+            for (var i = 0; i < alias!.Length; i++)
+            {
+                if (char.IsWhiteSpace(alias[i]))
+                {
+                    throw new ArgumentException($"{GetType().Name} alias cannot contain whitespace: \"{alias}\"");
```
Keboo, line 154:
```suggestion
                    throw new ArgumentException($"{GetType().Name} alias cannot contain whitespace: \"{alias}\"", nameof(alias));
```

src/System.CommandLine/Symbol.cs
```diff
@@ -145,17 +122,38 @@ public bool HasAlias(string alias)
...
+
+        [DebuggerStepThrough]
+        private protected void ThrowIfAliasIsInvalid(string alias)
+        {
+            if (string.IsNullOrWhiteSpace(alias))
+            {
+                throw new ArgumentException("An alias cannot be null, empty, or consist entirely of whitespace.");
+            }
+
+            for (var i = 0; i < alias!.Length; i++)
```
Keboo, line 150: Is the null forgiveness operator needed here? The parameter is declared non-nullable.

mikem8361:
Can you let me know what build version these changes made it into?

PR #1044, opened 2020-09-19
Differentiate symbol types

jonsequitur:
I'm looking for feedback on this change. Does it make sense? Is it helpful?

The key change is to introduce the interface `IIdentifierSymbol : ISymbol` and to have `Command` and `Option` implement this new interface, while `Argument` implements only `ISymbol`.

Distinctions between these types are:

* An `IIdentifierSymbol` can have aliases, so the `HasAlias` method and `Aliases` property are there. `Argument` never really supported these so removing them from `Argument` seems more honest.
* `Argument.Name` is used for help output and for binding, but _not_ for token matching during parsing.
* `IIdentifierSymbol.Aliases` are used for help output, binding, and token matching.


Keboo (approved):

KathleenDollard:
Finally! a sane way to unravel this naming issue. Looks great.

Issue #1027, opened 2020-08-22
Trailing spaces break parsing in a response file


AndrewSav:
If you have a partial command line like `--option1 value1 --option2 value2 --options3 value3` and you copy and paste it to a response file, and then break it up to a line per token so that the response file works, if you are not careful enough you can leave trailing spaces behind like this:

```
--option1<space>
value1<space>
--option2<space>
value2<space>
--options3<space>
value3
```

This is not get parsed as expected with `program @responseFile` and gives validation errors instead, similar to this:

```
Unrecognized command or argument '--option1 '
Unrecognized command or argument 'value1 '
```

etc.

jonsequitur:
Thanks for the fix, @hackf5!

PR #1045, opened 2020-09-25
Resolved bug #1027

hackf5:
The tests are overly verbose as I'm using VS2019 not 17 and wasn't sure exactly what was allowed.

### DIFF
--- a/src/System.CommandLine.Tests/ResponseFileTests.cs
+++ b/src/System.CommandLine.Tests/ResponseFileTests.cs
@@ -336,5 +336,31 @@ namespace System.CommandLine.Tests
             result.FindResultFor(option3).GetValueOrDefault().Should().Be(3);
             result.Errors.Should().BeEmpty();
         }
+
+        [Fact]
+        public void When_response_file_options_or_arguments_contain_trailing_spaces_they_are_ignored()
+        {
+            var responseFile1 = ResponseFile("--option1 ", "value1 ", "--option2 ", "2 ");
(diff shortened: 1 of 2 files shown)


src/System.CommandLine.Tests/ResponseFileTests.cs
```diff
@@ -336,5 +336,31 @@ public void Response_files_can_refer_to_other_response_files()
             result.FindResultFor(option3).GetValueOrDefault().Should().Be(3);
             result.Errors.Should().BeEmpty();
         }
+
+        [Fact]
+        public void When_response_file_options_or_arguments_contain_trailing_spaces_they_are_ignored()
```
jonsequitur, line 341: This looks like three separate tests and breaking them up will make it a little clearer.

Trailing spaces for options:
```csharp
var responseFile1 = ResponseFile("--option1 ", "value1 ", "--option2 ", "2 ");
```

Leading spaces for options:
```csharp
var responseFile2 = ResponseFile(" --option1", " value1", " --option2", " 2");
```

Trailing *and* leading spaces for arguments:
```csharp
var responseFile3 = ResponseFile(" --option1 ", " value1 ", " --option2 ", " 2 ");
```

hackf5: Agreed and done. Wasn't sure how granular the tests should be, but I realised after committing that the name was wrong anyway.

jonsequitur: Nice to add in tabs as well!

jonsequitur (approved):

Issue #970, opened 2020-07-10
Incorrect ANSI escape code


JayJeckel:
While moving a bunch of in-house console apps to dotnet core using System.CommandLine, I've come across a few bugs in the System.CommandLine.Rendering.Ansi class.

* Ansi.Color.Background class isn't marked static.
* Ansi.Cursor.Move.NextLine() is missing the open bracket `[` after the escape, ie `$"{Esc}{line}E";` should be `$"{Esc}[{line}E";`.

According to the ms docs [Console Virtual Terminal Sequences](https://docs.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences?redirectedfrom=MSDN) and the wikipedia [ANSI escape code](https://en.wikipedia.org/wiki/ANSI_escape_code) article, the following also need correcting.

* Ansi.Cursor.Scroll.UpOne() should be `A` instead of `D`. Up one line is `A`, while `D` is left one character.
* Ansi.Cursor.Scroll.DownOne() should be `B` instead of `M`. Down one line is `B`, while `M` is the reverse of `\n` and moves up one line.
* Ansi.Cursor.SavePosition should be `s` instead of `7`.
* Ansi.Cursor.RestorePosition should be `u` instead of `8`.

hackf5:
What you say may be correct, however the current implementation works, apart from `Ansi.Cursor.Move.NextLine()`, which is indeed broken.

I've added a new sample to the `RenderingPlayground` that demonstrates this. If you get this branch https://github.com/hackf5/command-line-api/tree/bug-970-ansi-escape-codes and run the `RenderingPlayground` sample it will load the new `Cursor` sample by default. This sample demonstrates that the cursor currently works as expected (I've fixed `NextLine()`) on Windows at least.

I would be happy to tidy this up and add some unit tests in order to close this bug.

hackf5:
I've reread the spec and 

- Ansi.Cursor.SavePosition = `7`
- Ansi.Cursor.RestorePosition = `8`

are fine; see [here](https://docs.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences#simple-cursor-positioning).

However, to be explicit I think that scroll up and down are better represented as:

- Up = `[S`
- Down = `[T`

since this exactly matches the [spec](https://docs.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences#viewport-positioning).

I've updated the sample.

I will update the test cases later and submit a PR.

adamsitnik:
[...]

PR #1048, opened 2020-09-26
Bug #970 ansi escape codes

hackf5:
In order to have a valid discussion on this issue I felt it was necessary to add a new sample to the `RenderingPlayground` that used the cursor methods; this can be accessed by running with `--sample Cursor`.

The original implementation worked for the most part, however:

- `Move.NextLine` was wrong, as stated in the bug report, and this has been fixed.
- `Scroll.UpOne / DownOne` were partially correct, in that one direction worked, but not both. I think that the problem here was that short sequences were being used, but there is only one scroll short sequence. I've changed it to use the full scroll sequences, for which there is one in either direction. The suggestion in the bug report looks wrong.

It is worth noting that scroll scrolls the content not the screen, so it goes in the opposite direction to that which you are expecting. This behaviour is maintained in the API, however I've swapped them round so the seem natural in the sample. Also scrolling wipes out the parts of the screen that are scrolled. This seems to be the expected behaviour.

I've added tests (for what they are worth) and marked the non-static classes as static.

### DIFF
--- a/samples/RenderingPlayground/Program.cs
+++ b/samples/RenderingPlayground/Program.cs
@@ -28,7 +28,7 @@ namespace RenderingPlayground
         public static void Main(
 #pragma warning disable CS1573 // Parameter has no matching param tag in the XML comment (but other parameters do)
             InvocationContext invocationContext,
-            SampleName sample = SampleName.Cursor,
+            SampleName sample = SampleName.Dir,
             int? height = null,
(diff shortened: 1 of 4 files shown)


jonsequitur (approved):

jonsequitur:
Thanks, @hackf5! Nice addition.

Issue #1051, opened 2020-10-02
Support "Or" Argument 


tdashworth:
Would it be possible to support two (or more?) optional arguments that together must be required. For example:

`dotnet clean <PROJECT>|<SOLUTION>`

where you must pass either a project or a solution. It this case they would be the same type `FileInfo` but I'm looking for a `FileInfo` and `string`. 

Otherwise is it possible to setup two commands? One `clitool subcommand1 <string-arg>` and another `clitool subcommand1 <file-arg>`?

jonsequitur:
You can do this with a `ParseArgument<T>` delegate, for example:

```csharp

public class FileInfoOrString
{
    public FileInfoOrString(FileInfo fileInfo) => FileInfoArg = fileInfo;
    public FileInfoOrString(String s) => StringArg = s;
    public FileInfo FileInfoArg {  get; }
    public String StringArg {  get; }
}

var root = new RootCommand
{
    new Argument<FileInfoOrString>("FileInfo|String", parse: argResult => {
        var arg = argResult.Tokens.Single().Value;
        if (File.Exists(arg))
        {
            return new FileInfoOrString(new FileInfo(arg));
        } 

        return new FileInfoOrString(arg);
    })
};


```

tdashworth:
Thanks! Exactly what I need 😀

tdashworth:
I was struggling to use the argument name "FileInfo|String" and for it to bind to the command handler. I have made a PR for a small change to replace the "|" with "or" so that it allows a friendlier parameter name to the handler.

PR #1053, opened 2020-10-03
Replace | with or when searching for Argument value

tdashworth:
This allows the CommandHandler to be registered with `fileOrString` for an Argument named `File|String`.

#1051

### DIFF
--- a/src/System.CommandLine.Tests/Invocation/CommandHandlerTests.cs
+++ b/src/System.CommandLine.Tests/Invocation/CommandHandlerTests.cs
@@ -415,5 +415,96 @@ namespace System.CommandLine.Tests.Invocation
             testClass.boundName.Should().Be("Gandalf");
             testClass.boundAge.Should().Be(425);
         }
+
+        [Fact]
+        public async Task Method_parameters_on_the_invoked_method_are_bound_to_matching_argument_names()
+        {
+            string boundName = default;
+            int boundAge = default;
+
+            void Execute(string name, int age)
+            {
+                boundName = name;
+                boundAge = age;
+            }
+
+            var command = new Command("command");
+            command.AddArgument(new Argument<int>("age"));
+            command.AddArgument(new Argument<string>("name"));
+            command.Handler = CommandHandler.Create<string, int>(Execute);
+
+            await command.InvokeAsync("command 425 Gandalf", _console);
+
+            boundName.Should().Be("Gandalf");
+            boundAge.Should().Be(425);
+        }
+
+
+        [Fact]
+        public async Task Method_parameters_on_the_invoked_method_can_be_bound_to_hyphenated_argument_names()
+        {
+            string boundFirstName = default;
+
+            void Execute(string firstName)
+            {
+                boundFirstName = firstName;
+            }
+
+            var command = new Command("command")
+            {
+                new Argument<string>("first-name")
+            };
+            command.Handler = CommandHandler.Create<string>(Execute);
+
+            await command.InvokeAsync("command Gandalf", _console);
+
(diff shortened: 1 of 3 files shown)


jonsequitur:
This is a pretty cool idea. I worry that string-based conventions like this can be a bit hard to discover, but in this case, not knowing about it wouldn't get in the way. Not knowing about the name-based matching in general, on the other hand, has been the most common pitfall in our design, thus #1012. 

Adding @keboo and @KathleenDollard for their thoughts.

Keboo:
No real objection from me but it would be nice to see some tests on this.

tdashworth:
@Keboo Will look to add tests in the coming days 😀

dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/command-line-api?pullRequest=1053) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/command-line-api?pullRequest=1053) before we can accept your contribution.<br/><br/>:x: tdashworth [sign now](https://cla.dotnetfoundation.org/dotnet/command-line-api?pullRequest=1053)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/command-line-api?pullRequest=1053) it.</sub>

tdashworth:
Hey - any chance of a review now I've added some tests? Thanks

src/System.CommandLine.Tests/Invocation/CommandHandlerTests.cs
```diff
@@ -415,5 +415,96 @@ public async Task Method_parameters_on_the_invoked_member_method_are_bound_to_ma
...
+            command.Handler = CommandHandler.Create<string, int>(Execute);
+
+            await command.InvokeAsync("command 425 Gandalf", _console);
+
+            boundName.Should().Be("Gandalf");
+            boundAge.Should().Be(425);
+        }
+
+        [Fact]
+        public async Task Method_parameters_on_the_invoked_method_are_bound_to_matching_argument_names_with_pipe_in()
```
jonsequitur, line 488: Other than this test, it looks like the added tests are duplicates.

tdashworth: The other tests are to similar to others but use Arguments only not Options. Happy to remove but when debugging they seemed to be a different code path?

jonsequitur: Oh, I see. I missed the distinction. It's fine to add these, then.

src/System.CommandLine/Parsing/CommandResultExtensions.cs
```diff
@@ -36,7 +36,7 @@ public static class CommandResultExtensions
             {
                 foreach (var argument in commandResult.Command.Arguments)
                 {
-                    if (valueName.IsMatch(argument.Name))
+                    if (valueName.IsMatch(argument.Name) || valueName.IsMatch(argument.Name.Replace("|","or")))
```
jonsequitur, line 39: This change should maybe be moved to the `IsMatch` method.

tdashworth: Done. I had considered not doing to before to allow the or with and without the replacement. After realising that a `|` is invalid in a parameter name - this shouldn't cause any breaking changes.

tdashworth:
Can this be reviewed again, please?

jonsequitur (approved):

Issue #1077, opened 2020-11-12
Allow opt out of multiple arguments per option flag


sfoslund:
The old parser used in the .NET CLI had behavior such that to specify multiple values for an option you had to include the option flag before each one:
`--source source1 --source source2 argument`
With System.CommandLine, we're able to specify options one after another without re-specifying the option flag:
`--source source1 source2 argument`
This creates confusion in the following case:
`--source source1 argument`
With existing CLI behavior, `argument` is treated as a command level argument whereas with existing system command line behavior it's treated as another argument to the `--source` option. It would be great is there's a way to opt out of this behavior to avoid breaking changes to the CLI. cc @jonsequitur

PR #1079, opened 2020-11-17
Allow opting out of multiple arguments per option flag

sfoslund:
Fixes https://github.com/dotnet/command-line-api/issues/1077. This contains a change to IOption, which is obviously not ideal so let me know if you have an alternate approach you would suggest @jonsequitur

### DIFF
--- a/src/System.CommandLine.Tests/OptionTests.cs
+++ b/src/System.CommandLine.Tests/OptionTests.cs
@@ -358,20 +358,7 @@ namespace System.CommandLine.Tests
         }
 
         [Fact]
-        public void When_Name_is_set_to_its_current_value_then_it_is_not_removed_from_aliases()
-        {
-            var option = new Option("--name");
-
-            option.Name = "name";
-
-            option.HasAlias("name").Should().BeTrue();
-            option.HasAlias("--name").Should().BeTrue();
-            option.Aliases.Should().Contain("--name");
-            option.Aliases.Should().Contain("name");
-        }
-
-        [Fact]
-        public void Single_option_arg_is_matched_when_disallowing_multiple_args_per_option_flag()
+        public void Single_option_arg_is_matched_when_disallowing_multiple_args_per_option_token()
         {
             var option = new Option<string[]>("--option") { AllowMultipleArgumentsPerOptionFlag = false };
             var command = new Command("the-command") { option };
@@ -385,7 +372,7 @@ namespace System.CommandLine.Tests
         }
 
         [Fact]
-        public void Multiple_option_args_are_matched_with_multiple_option_flags_when_disallowing_multiple_args_per_option_flag()
+        public void Multiple_option_args_are_matched_with_multiple_option_tokens_when_disallowing_multiple_args_per_option_token()
         {
             var option = new Option<string[]>("--option") { AllowMultipleArgumentsPerOptionFlag = false };
             var command = new Command("the-command") { option };
@@ -395,6 +382,19 @@ namespace System.CommandLine.Tests
             var optionResult = result.ValueForOption(option);
 
             optionResult.Should().BeEquivalentTo(new[] { "1", "2" });
+        }
+
+        [Fact]
+        public void When_Name_is_set_to_its_current_value_then_it_is_not_removed_from_aliases()
+        {
+            var option = new Option("--name");
+
+            option.Name = "name";
+
+            option.HasAlias("name").Should().BeTrue();
+            option.HasAlias("--name").Should().BeTrue();
+            option.Aliases.Should().Contain("--name");
+            option.Aliases.Should().Contain("name");
         }
 
         protected override Symbol CreateSymbol(string name) => new Option(name);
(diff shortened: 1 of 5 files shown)


src/System.CommandLine.Tests/ParserTests.MultipleArguments.cs
```diff
@@ -176,6 +176,36 @@ public void arity_ambiguities_can_be_differentiated_by_type_convertibility()
                       .BeEquivalentTo(new[] { "one", "two" },
                                       options => options.WithStrictOrdering());
             }
+
+            [Theory]
+            [InlineData(true)]
+            [InlineData(false)]
+            public void Multiple_arguments_can_be_specified_after_option(bool allowMultipleArguments)
```
jonsequitur, line 183: We typically align the test name to the scenario. There are two different scenarios here, which the use of `Theory` kind of hides. I also think the existing behavior (your `false` case) is already tested.

So I'd suggest only testing the new behavior, and letting the test name describe it explicitly.

src/System.CommandLine.Tests/ParserTests.MultipleArguments.cs
```diff
@@ -176,6 +176,36 @@ public void arity_ambiguities_can_be_differentiated_by_type_convertibility()
                       .BeEquivalentTo(new[] { "one", "two" },
                                       options => options.WithStrictOrdering());
             }
```
jonsequitur, line 178: These tests probably belong under `OptionTests`. (Per the pedantic distinction from earlier, the new functionality is about multiple tokens for a single option argument. Options don't support multiple `Argument` instances, which is what this test class is otherwise referring to.)

src/System.CommandLine.Tests/ParserTests.MultipleArguments.cs
sfoslund: No problem, just pushed these changes.

src/System.CommandLine.Tests/OptionTests.cs
```diff
@@ -357,6 +357,33 @@ public void Option_T_default_value_is_validated()
...
+            var result = command.Parse("--option 1 2");
+
+            var optionResult = result.ValueForOption(option);
+
+            optionResult.Should().BeEquivalentTo(new[] { "1" });
+            result.UnmatchedTokens.Should().BeEquivalentTo(new[] { "2" });
+        }
+
+        [Fact]
+        public void Multiple_option_args_are_matched_with_multiple_option_flags_when_disallowing_multiple_args_per_option_flag()
```
jonsequitur, line 375:
```suggestion
        public void Multiple_option_args_are_matched_with_multiple_option_tokens_when_disallowing_multiple_args_per_option_token()
```

src/System.CommandLine.Tests/OptionTests.cs
```diff
@@ -357,6 +357,33 @@ public void Option_T_default_value_is_validated()
                 .BeEquivalentTo(new[] { "ERR" });
         }
 
+        [Fact]
+        public void Single_option_arg_is_matched_when_disallowing_multiple_args_per_option_flag()
```
jonsequitur, line 361:
```suggestion
        public void Single_option_arg_is_matched_when_disallowing_multiple_args_per_option_token()
```

jonsequitur:
I'd like to come up with simpler naming if possible than `AllowMultipleArgumentsPerOptionFlag` and also keep #1083 in mind, though let's not worry about fixing #1083 in this PR. 

For example, if we allow multiple occurrences of the option token, but the arity is `ExactlyOne`, then what does this mean? Given the current behavior, it would be a bit of a contradiction, in that multiple occurrences are "allowed" but then result in a parse error. But we could say that in the `ExactlyOne` arity case, allowing multiple option tokens enables #1083, then that might be reasonable. 

Does the default for `AllowMultipleArgumentsPerOptionFlag` (`true`) seem correct for both of these use cases?

Thoughts? 

@KathleenDollard  @Keboo

Keboo:
I like the idea, but I do agree that perhaps a simpler names is in order. Given the ambiguity of the case called out by #1077, I am tempted to say that the default should be flipped to require the option to be specified (not 100% on this, but I think making ambiguity and opt-in feature makes more sense). 

As for a name perhaps something like: `SetForEachArgument` with a default of `true`

jonsequitur:
> Given the ambiguity of the case called out by #1077, I am tempted to say that the default should be flipped to require the option to be specified

This would be a breaking change.

sfoslund:
I'm not sure how common this behavior is so I don't know that the default needs to be changed, but the name `SetForEachArgument` works for me, I can update this PR if everyone is in agreement on it.

jonsequitur:
I'm hoping for a name that works for both this and #1083.

sfoslund:
> I'm hoping for a name that works for both this and #1083. (I'm not sure this is possible but I can dream.)

Are you planning on making that behavior opt in too? My understanding was that this would not require a feature flag.

jonsequitur:
I think it can likely become the default.

jonsequitur (approved):

PR #1080, opened 2020-11-17
Reduce allocations

paulomorgado:
# Change

Reduce allocations in:

* `string` comparisons
* `List<T>` growth
* **LINQ**
  * Enumerators
  * Delegates
  * Closures

### DIFF
--- a/src/System.CommandLine.Rendering/Ansi.cs
+++ b/src/System.CommandLine.Rendering/Ansi.cs
@@ -51,7 +51,7 @@ namespace System.CommandLine.Rendering
                 public static AnsiControlCode LightCyan { get; } = $"{Esc}[106m";
                 public static AnsiControlCode LightGray { get; } = $"{Esc}[107m";
 
-                public static AnsiControlCode Rgb(byte r, byte g, byte b) => $"{Esc}[48;2;{r};{g};{b}m";
+                public static AnsiControlCode Rgb(byte r, byte g, byte b) => $"{Esc}[48;2;{r.ToString()};{g.ToString()};{b.ToString()}m";
             }
 
             [DebuggerStepThrough]
@@ -76,7 +76,7 @@ namespace System.CommandLine.Rendering
                 public static AnsiControlCode LightCyan { get; } = $"{Esc}[96m";
                 public static AnsiControlCode LightGray { get; } = $"{Esc}[97m";
 
-                public static AnsiControlCode Rgb(byte r, byte g, byte b) => $"{Esc}[38;2;{r};{g};{b}m";
+                public static AnsiControlCode Rgb(byte r, byte g, byte b) => $"{Esc}[38;2;{r.ToString()};{g.ToString()};{b.ToString()}m";
             }
         }
 
@@ -87,13 +87,13 @@ namespace System.CommandLine.Rendering
             [DebuggerStepThrough]
             public static class Move
             {
-                public static AnsiControlCode Up(int lines = 1) => $"{Esc}[{lines}A";
-                public static AnsiControlCode Down(int lines = 1) => $"{Esc}[{lines}B";
-                public static AnsiControlCode Right(int columns = 1) => $"{Esc}[{columns}C";
-                public static AnsiControlCode Left(int columns = 1) => $"{Esc}[{columns}D";
-                public static AnsiControlCode NextLine(int line = 1) => $"{Esc}[{line}E";
+                public static AnsiControlCode Up(int lines = 1) => $"{Esc}[{lines.ToString()}A";
+                public static AnsiControlCode Down(int lines = 1) => $"{Esc}[{lines.ToString()}B";
+                public static AnsiControlCode Right(int columns = 1) => $"{Esc}[{columns.ToString()}C";
+                public static AnsiControlCode Left(int columns = 1) => $"{Esc}[{columns.ToString()}D";
+                public static AnsiControlCode NextLine(int line = 1) => $"{Esc}[{line.ToString()}E";
                 public static AnsiControlCode ToUpperLeftCorner { get; } = $"{Esc}[H";
-                public static AnsiControlCode ToLocation(int? left = null, int? top = null) => $"{Esc}[{top ?? 1};{left ?? 1}H";
+                public static AnsiControlCode ToLocation(int? left = null, int? top = null) => $"{Esc}[{top.GetValueOrDefault(1)};{left.GetValueOrDefault(1).ToString()}H";
             }
 
             [DebuggerStepThrough]
--- a/src/System.CommandLine.Suggest/FileSuggestionRegistration.cs
+++ b/src/System.CommandLine.Suggest/FileSuggestionRegistration.cs
@@ -4,6 +4,7 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Text;
 using static System.Environment;
 
 namespace System.CommandLine.Suggest
@@ -30,7 +31,7 @@ namespace System.CommandLine.Suggest
             }
 
             var userProfile = GetFolderPath(SpecialFolder.UserProfile);
-            
+
             _registrationConfigurationFilePath = Path.Combine(userProfile, RegistrationFileName);
         }
 
@@ -47,11 +48,19 @@ namespace System.CommandLine.Suggest
                 return null;
             }
 
-            var completionTarget =
-                File.ReadAllLines(_registrationConfigurationFilePath).LastOrDefault(line =>
-                                                                                        line.StartsWith(soughtExecutable.FullName, StringComparison.OrdinalIgnoreCase));
+            string completionTarget = null;
+            using (var sr = new StreamReader(_registrationConfigurationFilePath, Encoding.UTF8))
+            {
+                while (sr.ReadLine() is string line)
+                {
+                    if (line.StartsWith(soughtExecutable.FullName, StringComparison.OrdinalIgnoreCase))
+                    {
+                        completionTarget = line;
+                    }
+                }
+            }
 
-            if (completionTarget == null)
+            if (completionTarget is null)
             {
                 // Completion provider not found!
                 return null;
@@ -66,13 +75,17 @@ namespace System.CommandLine.Suggest
 
             if (_registrationConfigurationFilePath != null && File.Exists(_registrationConfigurationFilePath))
             {
-                allRegistration
-                    .AddRange(File
-                        .ReadAllLines(_registrationConfigurationFilePath)
-                        .Select(l => l.Trim())
-                        .Where(l => l.Any())
-                        .Select(item => new Registration(item))
-                    );
+                using (var sr = new StreamReader(_registrationConfigurationFilePath, Encoding.UTF8))
+                {
+                    string line;
+                    while ((line = sr.ReadLine()) != null)
+                    {
+                        if (!string.IsNullOrWhiteSpace(line))
+                        {
+                            allRegistration.Add(new Registration(line.Trim()));
+                        }
+                    }
+                }
             }
 
             return allRegistration;
--- a/src/System.CommandLine/Binding/Binder.cs
+++ b/src/System.CommandLine/Binding/Binder.cs
@@ -7,15 +7,45 @@ namespace System.CommandLine.Binding
 {
     internal static class Binder
     {
-        internal static bool IsMatch(this string parameterName, string alias) =>
-            string.Equals(alias?.RemovePrefix()
-                              .Replace("-", "")
-                              .Replace("|", "or"),
-                          parameterName,
-                          StringComparison.OrdinalIgnoreCase);
+        internal static bool IsMatch(this string parameterName, string alias)
+        {
+            var parameterIndex = 0;
+            for (var aliasIndex = alias.AdvancePrefix(); aliasIndex < alias.Length; aliasIndex++)
+            {
+                var aliasChar = alias[aliasIndex];
+
+                if (aliasChar == '-')
+                {
+                    continue;
+                }
+
+                var parameterNameChar = parameterName[parameterIndex];
+
+                if (aliasChar == '|')
+                {
+                    if ((parameterName.Length < parameterIndex + 2) || ((parameterNameChar | 32) != 'o') || ((parameterName[parameterIndex + 1] | 32) != 'r'))
+                    {
+                        return false;
+                    }
+
+                    parameterIndex += 2;
+
+                    continue;
+                }
+
+                if (parameterNameChar != aliasChar && char.ToUpperInvariant(parameterNameChar) != char.ToUpperInvariant(aliasChar))
+                {
+                    return false;
+                }
+
+                parameterIndex++;
+            }
+
+            return true;
+        }
 
         internal static bool IsMatch(this string parameterName, IOption symbol) =>
-            parameterName.IsMatch(symbol.Name) || 
+            parameterName.IsMatch(symbol.Name) ||
             symbol.HasAlias(parameterName);
 
         internal static bool IsNullable(this Type t)
--- a/src/System.CommandLine/Help/HelpBuilder.cs
+++ b/src/System.CommandLine/Help/HelpBuilder.cs
@@ -5,7 +5,6 @@ using System.Collections.Generic;
 using System.CommandLine.IO;
 using System.CommandLine.Parsing;
 using System.Linq;
-using System.Text;
 using System.Text.RegularExpressions;
 using static System.CommandLine.Help.DefaultHelpText;
 
@@ -255,22 +254,22 @@ namespace System.CommandLine.Help
         /// </summary>
         /// <param name="table">The table of values to determine column widths for.</param>
         /// <returns>A collection of column widths.</returns>
-        private IReadOnlyList<int> ColumnWidths(IEnumerable<IReadOnlyList<string>> table)
+        private IReadOnlyList<int> ColumnWidths(IReadOnlyList<IReadOnlyList<string>> table)
         {
-            if (!table.Any())
+            if (table.Count == 0)
             {
                 return Array.Empty<int>();
             }
 
-            var columns = table.First().Count;
+            var columns = table[0].Count;
             var unsetWidth = -1;
             var widths = new int[columns];
-            for (int i = 0; i < columns; ++i)
+            for (var i = 0; i < columns; ++i)
             {
                 widths[i] = unsetWidth;
             }
             var maxWidths = new int[columns];
-            for (int i = 0; i < columns; ++i)
+            for (var i = 0; i < columns; ++i)
             {
                 maxWidths[i] = table.Max(row => row[i].Length);
             }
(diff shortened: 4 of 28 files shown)


src/System.CommandLine/TypeExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.Runtime.CompilerServices;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace System.CommandLine
+{
+    /// <summary>
+    /// Provides extensions to <see cref="Type"/>.
+    /// </summary>
+    public static class TypeExtensions
```
jonsequitur, line 13: These should be kept internal. Maybe include them alongside similar methods in `Binder.cs`?

src/System.CommandLine/StringBuilderPool.cs
```diff
@@ -0,0 +1,68 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace System.CommandLine
+{
+    public class StringBuilderPool
```
jonsequitur, line 10: This should probably be internal.

paulomorgado: Should I repeat it across all projects that need it? Or àdd `InternalsVisibleToAttribute`?

src/System.CommandLine/TypeExtensions.cs
paulomorgado: I don't think this is binder specific.

These are extension methods on `Type` that are used by the binder.

It might be a good idea to move all extension methods on `Type` to this class.

src/System.CommandLine/StringBuilderPool.cs
paulomorgado: @jonsequitur, I was about to make `StringBuilderPool` internal and add to the other projects as a link, but that defeats the whole purpose of the shared pool.

src/System.CommandLine/Parsing/ParseResultVisitor.cs
```diff
@@ -34,19 +34,27 @@ internal class ParseResultVisitor : SyntaxVisitor
...
             {
-                _unmatchedTokens.AddRange(unmatchedTokens.Select(t => t.Value));
+                foreach (var unmatchedToken in unmatchedTokens)
+                {
+                    _unmatchedTokens.Add(unmatchedToken.Value);
+                }
             }
 
-            _errors = new List<ParseError>();
+            _errors = new List<ParseError>(2);
```
Keboo, line 57: Shouldn't this be?
```suggestion
            _errors = new List<ParseError>(_tokenizeResult.Errors.Count + parseErrors.Count);
```

src/System.CommandLine/TypeExtensions.cs
jonsequitur: Within the scope of this project I believe they're only used for binding, which is why e.g. `IsNullable` is also there.

src/System.CommandLine.Rendering/TestTerminal.cs
```diff
@@ -247,6 +247,8 @@ public IEnumerable<TextRendered> RenderOperations()
             {
                 yield return new TextRendered(buffer.ToString(), position);
             }
+
+            StringBuilderPool.Default.Return(buffer);
```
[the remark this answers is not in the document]
paulomorgado: Thought of that but was not sure how much would that complicate the state machine.

src/System.CommandLine.Rendering/WrappingExtensions.cs
```diff
@@ -70,6 +70,8 @@ public static IEnumerable<string> SplitForWrapping(this string text)
             {
                 yield return sb.ToString();
             }
+
+            StringBuilderPool.Default.Return(sb);
```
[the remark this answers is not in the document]
paulomorgado: Thought of that but was not sure how much would that complicate the state machine.

src/System.CommandLine/StringBuilderPool.cs
jonsequitur: I typically avoid `InternalsVisibleToAttribute` because the impact of it tends to spread out over time.

We've used linked files in a few places but because this one is stateful, that approach might defeat the purpose.

Because of the lower level of maturity for the other libraries, maybe we should skip this optimization for now? Other ideas?

src/System.CommandLine/Binding/Binder.cs
```diff
@@ -7,15 +7,50 @@ namespace System.CommandLine.Binding
...
+                if (aliasChar == '-')
+                {
+                    continue;
+                }
+
+                var parameterNameChar = parameterName[parameterIndex];
+
+                if (aliasChar == '|')
+                {
+                    if ((parameterName.Length < parameterIndex + 2) || ((parameterNameChar | 32) != 'o') || ((parameterName[parameterIndex + 1] | 32) != 'r'))
```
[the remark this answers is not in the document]
paulomorgado: If you set that bit (00100000) on a character and it becomes an **o** or **r**, it's because it was, respectively, an **O** or **R**. It's a no-op for **o** or **r** and it doesn't matter for other characters.

src/System.CommandLine/Help/HelpBuilder.cs
```diff
@@ -377,48 +377,116 @@ private void AppendRow(IEnumerable<string> row, IReadOnlyList<int> columnWidths)
...
+
+                return lines;
+            }
+            finally
+            {
+                StringBuilderPool.Default.Return(lineBuilder);
             }
 
-            return lines;
+            void appendLine()
```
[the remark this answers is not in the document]
paulomorgado: It's a local function, not a member method.

src/System.CommandLine/Help/HelpBuilder.cs
```diff
@@ -377,48 +377,116 @@ private void AppendRow(IEnumerable<string> row, IReadOnlyList<int> columnWidths)
...
+            {
+                lines.Add(lineBuilder.ToString(lineBuilderOffset, (isWhiteSpace ? lineBuilder.Length - 1 : lineBuilder.Length) - lineBuilderOffset));
+            }
+
+            void truncateLineAndAppendIfNeeded()
```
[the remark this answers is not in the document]
paulomorgado: It's a local function, not a member method.

src/System.CommandLine/Help/HelpBuilder.cs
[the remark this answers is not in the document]
Keboo: Agreed, however the other other local functions are using PascalCase. It would be good to maintain consistency.

src/System.CommandLine/Binding/Binder.cs
```diff
@@ -7,15 +7,50 @@ namespace System.CommandLine.Binding
 {
     internal static class Binder
     {
-        internal static bool IsMatch(this string parameterName, string alias) =>
-            string.Equals(alias?.RemovePrefix()
-                              .Replace("-", "")
-                              .Replace("|", "or"),
-                          parameterName,
-                          StringComparison.OrdinalIgnoreCase);
+        internal static bool IsMatch(this string parameterName, string alias)
```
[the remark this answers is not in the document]
paulomorgado: This was misidentified as a change because it was a expression bodied member and it was changed into block body method. That's why I always put the arrow in the next line if it's not a one-liner.

The API was already defined with non-nullable strings but was accepting null strings. Removed the null checks.

src/System.CommandLine/Parsing/ParseResultVisitor.cs
paulomorgado: Good call! Missed that and the allocation of the enumerator and delegate in the next line

src/System.CommandLine/Binding/Binder.cs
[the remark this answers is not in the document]
jonsequitur: Out of curiosity, what's the performance implication of this versus a `char.Equals` or `string.Equals`?

One aspect of this particular feature that's worth calling out is that it's a localization problem, so the implementation is going to need to change to take translations into account at some point. That might disrupt the optimization, and at the very least it makes the logic here a little harder to follow.

src/System.CommandLine/Help/HelpBuilder.cs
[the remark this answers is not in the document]
paulomorgado: Maybe it's the others that are wrong. 😄

src/System.CommandLine/StringBuilderPool.cs
paulomorgado: Leave it public?

I understand the concern of leaking a API that is not the concern of the package.

I just think that using [`ObjectPoolProviderExtensions`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.objectpool.objectpoolproviderextensions) might be overkill because of the lifetime of the of the pool and pooled objects.

Relying in an hypothetical global `StringBuilder` pool would be even worst.

src/System.CommandLine/TypeExtensions.cs
paulomorgado: I tend to place them on <extended_type>Extensions classes.

If they are super specific, I don't define them as extension methods - maybe one day we can have private extensions.

src/System.CommandLine/Binding/Binder.cs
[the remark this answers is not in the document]
paulomorgado: It's not about `char.Equals` or `string.Equals`.

The original code was:

```C#
internal static bool IsMatch(this string parameterName, string alias) =>
        string.Equals(alias?.RemovePrefix()
                          .Replace("-", "")
                          .Replace("|", "or"),
                      parameterName,
                      StringComparison.OrdinalIgnoreCase);
```

It's creating 3 extra strings for an `OrdinalIgnoreCase` comparison.

Every time there's an avoidable string creation, a bucket of CO2 is thrown into the atmosphere. 😄 

Now that we're streaming the string, there are more efficient ways to compare ordinal characters.

The original code is not particularly localizable. By the time we get to that, there might be a code generator for regular expressions that may allow for a better localization.

But, the truth is, I hadn't thought about that.

src/System.CommandLine/StringBuilderPool.cs
jonsequitur: I don't think it should be public because then people will use it, and we'll have to maintain it. I also don't think we should add dependencies. 

It's unclear if `System.CommandLine.DragonFruit` or `System.CommandLine.Rendering` will reach a GA in their current form so I'm less concerned with optimizing their performance than I am with making sure the `System.CommandLine` surface area is clean.

paulomorgado: What do you think of taking a dependency on [Microsoft.Extensions.ObjectPool](https://www.nuget.org/packages/Microsoft.Extensions.ObjectPool/)?

I could keep it as a pooled weak references.

jonsequitur: I'd like to avoid adding a dependency.

src/System.CommandLine.Rendering/Ansi.cs
```diff
@@ -51,7 +51,7 @@ public static class Background
                 public static AnsiControlCode LightCyan { get; } = $"{Esc}[106m";
                 public static AnsiControlCode LightGray { get; } = $"{Esc}[107m";
 
-                public static AnsiControlCode Rgb(byte r, byte g, byte b) => $"{Esc}[48;2;{r};{g};{b}m";
+                public static AnsiControlCode Rgb(byte r, byte g, byte b) => $"{Esc}[48;2;{r.ToString()};{g.ToString()};{b.ToString()}m";
```
jonsequitur, line 54: Why the explicit `ToString` calls?

src/System.CommandLine.Suggest/FileSuggestionRegistration.cs
```diff
@@ -52,9 +53,18 @@ public Registration FindRegistration(FileInfo soughtExecutable)
...
             }
 
-            var completionTarget =
-                File.ReadAllLines(_registrationConfigurationFilePath).LastOrDefault(line =>
-                                                                                        line.StartsWith(soughtExecutable.FullName, StringComparison.OrdinalIgnoreCase));
+            string completionTarget = null;
+            using (var sr = new StreamReader(_registrationConfigurationFilePath, Encoding.UTF8))
+            {
+                string? line;
+                while ((line = sr.ReadLine()) != null)
```
jonsequitur, line 60: Or maybe do away with the separate variable declaration and do this?

```suggestion
                while (sr.ReadLine() is { } line)
```

src/System.CommandLine.Suggest/FileSuggestionRegistration.cs
```diff
@@ -52,9 +53,18 @@ public Registration FindRegistration(FileInfo soughtExecutable)
...
+                {
+                    if (line.StartsWith(soughtExecutable.FullName, StringComparison.OrdinalIgnoreCase))
+                    {
+                        completionTarget = line;
+                    }
+                }
+            }
 
             if (completionTarget == null)
```
jonsequitur, line 69: Maybe change to `if (completionTarget is not null)`?

src/System.CommandLine/Symbol.cs
```diff
@@ -59,22 +59,37 @@ private protected void AddArgumentInner(Argument argument)
...
+                            }
+                        }
+                        break;
+                }
+            }
+
+            return suggestions
+                .OrderBy(symbol => symbol!.IndexOfCaseInsensitive(textToMatch))
+                .ThenBy(symbol => symbol, StringComparer.OrdinalIgnoreCase)
+                .ToList();
```
jonsequitur, line 92: Is pre-emptively calling `ToList` useful here?

src/System.CommandLine.Rendering/Ansi.cs
paulomorgado: To avoid boxing and unboxing.

src/System.CommandLine.Suggest/FileSuggestionRegistration.cs
paulomorgado: I was under the impressions that this was targeting C# 7.3.

In that case, I would prever `while (sr.ReadLine() is string line)` in this case.

src/System.CommandLine.Suggest/FileSuggestionRegistration.cs
paulomorgado: Don't you mean `if (completionTarget is null)`?

src/System.CommandLine/Symbol.cs
paulomorgado: Not if it's iterated only once, which it seems to be the case. I'll remove it.

src/System.CommandLine.Suggest/FileSuggestionRegistration.cs
jonsequitur: We should update to latest, I think.

src/System.CommandLine.Suggest/FileSuggestionRegistration.cs
jonsequitur: Oops. Yeah.

src/System.CommandLine.Suggest/FileSuggestionRegistration.cs
paulomorgado: I prefer to be explicit. `latest` means different things for different builds of Visual Studio.

jonsequitur: That's fine. Any reason not to support C# 9 then?

paulomorgado: C# 9.0 is the default only for .NET 5.0 ([language defaults](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version#defaults)).

It must be done carefully to support .NET Framework and .NET Standard 2.0 (Visual Studio will default to C# 7.3) because some features depend on types that are only on .NET Core 3.x/.NET Standard 2.1/.NET 5.0, but it can be done.

jonsequitur:
@paulomorgado Would you mind rebasing this onto `main`?

paulomorgado:
@jonsequitur,

Done!
All tests passing!
We should upgrade C# to 9.0, because some constraints require it.

src/System.CommandLine/Parsing/ParseResultVisitor.cs
```diff
@@ -34,20 +34,33 @@ internal class ParseResultVisitor : SyntaxVisitor
             _tokenizeResult = tokenizeResult;
             _rawInput = rawInput;
 
-            _unparsedTokens = new List<string>();
-            if (unparsedTokens?.Count > 0)
+            var unparsedTokensCount = unparsedTokens?.Count ?? 0;
+            _unparsedTokens = unparsedTokensCount == 0 ? new List<string>() : new List<string>(unparsedTokensCount);
```
Keboo, line 38:
```suggestion
            _unparsedTokens = unparsedTokensCount == 0 ? new List<string>(0) : new List<string>(unparsedTokensCount);
```

src/System.CommandLine/Parsing/ParseResultVisitor.cs
```diff
@@ -34,20 +34,33 @@ internal class ParseResultVisitor : SyntaxVisitor
...
             {
-                _unparsedTokens.AddRange(unparsedTokens.Select(t => t.Value));
+                foreach (var unparsedToken in unparsedTokens!)
+                {
+                    _unparsedTokens.Add(unparsedToken.Value);
+                }
+            }
+
+            var unmatchedTokensCount = unmatchedTokens?.Count ?? 0;
+            _unmatchedTokens = unmatchedTokensCount == 0 ? new List<string>() : new List<string>(unmatchedTokensCount);
```
Keboo, line 48:
```suggestion
            _unmatchedTokens = unmatchedTokensCount == 0 ? new List<string>(0) : new List<string>(unmatchedTokensCount);
```

Keboo (approved):

src/System.CommandLine/Parsing/ParseResultVisitor.cs
paulomorgado: Interesting point.

The [parameterless constructor for `List<T>`](https://source.dot.net/#System.Private.CoreLib/List.cs,37) does exactly the same as the [constructor with initial capacity when the initial capacity is `0`](https://source.dot.net/#System.Private.CoreLib/List.cs,52) with the added (very minimal) cost of [checking for negative capacity](https://source.dot.net/#System.Private.CoreLib/List.cs,48).

But this might be an option with improved readability and maintainability:

    _unparsedTokens = new List<string>(unparsedTokensCount);

src/System.CommandLine/Parsing/ParseResultVisitor.cs
paulomorgado: Interesting point.

The [parameterless constructor for `List<T>`](https://source.dot.net/#System.Private.CoreLib/List.cs,37) does exactly the same as the [constructor with initial capacity when the initial capacity is `0`](https://source.dot.net/#System.Private.CoreLib/List.cs,52) with the added (very minimal) cost of [checking for negative capacity](https://source.dot.net/#System.Private.CoreLib/List.cs,48).

But this might be an option with improved readability and maintainability:

    _unmatchedTokens = new List<string>(unmatchedTokensCount);

jonsequitur (approved):

elgonzo:
Unfortunately, the change to the `Binder.IsMatch` method introduced with this pull request makes binding of option arguments to properties of complex model classes unreliable. See issue report here: https://github.com/dotnet/command-line-api/issues/1195

PR #1090, opened 2020-11-21
Fix complete registration issue where items with spaces are registere…

### LINKED ISSUE #1085: Can't get suggestions to work
I've spend a lot of time trying to get suggestions to work but to no avail.  I tried both on Windows and Ubuntu Linux using .Net Core 3.1 SDK.

I'm using the latest Nuget Package and followed the instructions [here](https://github.com/dotnet/command-line-api/blob/main/docs/dotnet-suggest.md).

Using dotnet-suggest (1.1.142701)  and System.CommandLine 2.0.0-beta1.20371.2 

There are some things I find confusing and would appreciate some help. The bash script added to my .bashrc contains this:
```
complete -F _dotnet_bash_complete `dotnet-suggest list`
```
My understanding is that is going to register a completion handler, which in turn calls a function called ` _dotnet_bash_complete`, which has been provided.

`dotnet-suggest list` returns a list of strings. Looking at the code, it defaults to all the dotnet tools, in two forms. One with dash, one without:
```
dotnet-try
dotnet try
dotnet-suggest
dotnet suggest
dotnet-trace
dotnet trace
```
It also looks for additional "user" paths registered in a special file here
```
~/.dotnet-suggest-registration.txt
```
These items can be registered manually, or it seems, automatically the first time you run your System.Console based app. However, it only puts the **path** in the list.

Next, I can inspect what has been registered as follows:
```
complete | grep _dot
complete -F _dotnet_bash_complete dotnet-suggest
complete -F _dotnet_bash_complete dotnet
complete -F _dotnet_bash_complete trace
complete -F _dotnet_bash_complete suggest
complete -F _dotnet_bash_complete dotnet-try
complete -F _dotnet_bash_complete try
complete -F _dotnet_bash_complete dotnet-trace
```
This seems consistent with the output of  `dotnet-suggest list` and what we're telling `complete` to register - but surely this is not what we wanted? 

My understanding is, we're telling the shell "If you type `trace` then call  `_dotnet_bash_complete`. The same for all the other registered commands. This in itself seems wrong - I don't want to call the script for items such as `try` and `trace`.

Could someone explain the flow for a console app called **myapp**? Surely it must actually register the string `myapp` (and not just the path where it's located) so that when you hit tab, it will run the script and in turn call dotnet-suggest, which calls myapp with the correct arguments to get help and then return to the user.

I may well be missing something here! Help appreciated.

pieter-venter:
…d separately

Call a function that changes the value of IFS to newline only. This ensures IFS is
restored to whatever it was before and would work any various platforms.

Fix #1085

### DIFF
--- a/src/System.CommandLine.Suggest/dotnet-suggest-shim.bash
+++ b/src/System.CommandLine.Suggest/dotnet-suggest-shim.bash
@@ -24,6 +24,11 @@ _dotnet_bash_complete()
     fi
 }
 
-complete -F _dotnet_bash_complete `dotnet-suggest list`
+_dotnet_bash_register_complete()
+{
+    local IFS=$'\n'
+    complete -F _dotnet_bash_complete `dotnet-suggest list`
+}
+_dotnet_bash_register_complete
 export DOTNET_SUGGEST_SCRIPT_VERSION="1.0.0"
 # dotnet suggest shell complete script end
--- a/src/System.CommandLine.Suggest/dotnet-suggest-shim.bash
+++ b/src/System.CommandLine.Suggest/dotnet-suggest-shim.bash
@@ -30,5 +30,5 @@ _dotnet_bash_register_complete()
     complete -F _dotnet_bash_complete `dotnet-suggest list`
 }
 _dotnet_bash_register_complete
-export DOTNET_SUGGEST_SCRIPT_VERSION="1.0.0"
+export DOTNET_SUGGEST_SCRIPT_VERSION="1.0.1"
 # dotnet suggest shell complete script end


src/System.CommandLine.Suggest/dotnet-suggest-shim.bash
```diff
@@ -24,6 +24,11 @@ _dotnet_bash_complete()
...
 }
 
-complete -F _dotnet_bash_complete `dotnet-suggest list`
+_dotnet_bash_register_complete()
+{
+    local IFS=$'\n'
+    complete -F _dotnet_bash_complete `dotnet-suggest list`
+}
+_dotnet_bash_register_complete
 export DOTNET_SUGGEST_SCRIPT_VERSION="1.0.0"
```
jonsequitur, line 33: Please update the script version to `1.0.1`.

pieter-venter: Done

jonsequitur (approved):

PR #1091, opened 2020-11-22
Fix dotnet-suggest issue that prevents files from being registered

### LINKED ISSUE #1085: Can't get suggestions to work

pieter-venter:
Fix #1085

### DIFF
--- a/src/System.CommandLine.Suggest.Tests/SuggestionRegistrationTest.cs
+++ b/src/System.CommandLine.Suggest.Tests/SuggestionRegistrationTest.cs
@@ -35,6 +35,55 @@ namespace System.CommandLine.Suggest.Tests
         }
 
         [Fact]
+        public void Missing_suggestion_can_not_be_found()
+        {
+            ISuggestionRegistration suggestionProvider = GetSuggestionRegistration();
+            
+            var path = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
+            try
+            {
+                Directory.CreateDirectory(path);
+                var unregisteredFile = Path.Combine(path, "im-not-registered");
+                File.WriteAllText(unregisteredFile, "");
+                
+                var foundRegistration = suggestionProvider.FindRegistration(new FileInfo(unregisteredFile));
+
+                foundRegistration
+                    .Should()
+                    .BeNull();
+            }
+            finally
+            {
+                Directory.Delete(path, true);
+            }
+        }     
+        
+        [Fact]
+        public void Added_suggestion_can_be_found()
+        {
+            ISuggestionRegistration suggestionProvider = GetSuggestionRegistration();
+            
+            var path = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
+            try
+            {
+                Directory.CreateDirectory(path);
+                var registeredFile = Path.Combine(path, "im-registered");
+                File.WriteAllText(registeredFile, "");
+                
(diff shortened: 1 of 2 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/command-line-api?pullRequest=1091) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/command-line-api?pullRequest=1091) before we can accept your contribution.<br/><br/>:x: pieter-venter [sign now](https://cla.dotnetfoundation.org/dotnet/command-line-api?pullRequest=1091)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/command-line-api?pullRequest=1091) it.</sub>

pieter-venter:
@jonsequitur I probably should have created separate issues in retrospect, but this MR is also required to fix the second issue detailed on #1085. LMK once you have some time to review and if it needs more work. Happy to help.

jonsequitur:
Thanks, @pieter-venter! This looks good.

jonsequitur (approved):

PR #1093, opened 2020-11-24
Ensure longer option alias is parsed when aliases are prefixes

sfoslund:


sfoslund:
Closing this after discussion with Jon, not needed.

PR #1096, opened 2020-11-25
XML documentation for RootCommand, Command and Argument classes

pieter-venter:
I would like to help resolve issue #130  

After doing some research, this seems to be the best reference to follow as a standard:
https://github.com/dotnet/dotnet-api-docs/wiki
and I followed it closely.

It looks like @KathleenDollard has made some XML comments in RootCommand and I would appreciate feedback on my attempt here. I worked on the `Command` class as a start.


jonsequitur (dismissed):

src/System.CommandLine/Command.cs
```diff
@@ -10,40 +10,83 @@
 
 namespace System.CommandLine
 {
+    /// <summary>
+    /// Represents a specific action that the application performs.
+    /// </summary>
+    /// <remarks>
+    /// Use the Command object for actions that correspond to a specific string (the command name). See
+    /// <see cref="RootCommand"/> for simple applications that only have one action. For example, <c>dotnet run</>
```
jonsequitur, line 18:
```suggestion
    /// <see cref="RootCommand"/> for simple applications that only have one action. For example, <c>dotnet run</c>
```

src/System.CommandLine/Command.cs
```diff
@@ -10,40 +10,83 @@
...
+        /// Initializes a new instance of the Command class.
+        /// </summary>
+        /// <param name="name">The name of the command.</param>
+        /// <param name="description">The description of the command, shown in help.</param>
         public Command(string name, string? description = null) : base(name, description)
         {
         }
 
+        /// <summary>
+        /// Represents all the arguments associated with the command.
```
jonsequitur, line 38:
```suggestion
        /// Represents all of the arguments associated with the command.
```

src/System.CommandLine/Command.cs
```diff
@@ -10,40 +10,83 @@
...
+        /// </summary>
         public IEnumerable<Argument> Arguments => 
             Children.OfType<Argument>();
 
+        /// <summary>
+        /// Represents all the options associated with the command, including global options.
```
jonsequitur, line 44:
```suggestion
        /// Represents all of the options associated with the command, including global options.
```

src/System.CommandLine/Command.cs
```diff
@@ -10,40 +10,83 @@
...
+        /// </summary>
         public IEnumerable<Option> Options =>
             Children.OfType<Option>()
                     .Concat(Parents
                             .OfType<Command>()
                             .SelectMany(c => c.GlobalOptions));
-
+        /// <summary>
+        /// Represents all the global options associated with the command
```
jonsequitur, line 52:
```suggestion
        /// Represents all of the global options associated with the command
```

src/System.CommandLine/Command.cs
```diff
@@ -10,40 +10,83 @@
...
         public IEnumerable<Option> GlobalOptions => _globalOptions.OfType<Option>();
 
+        /// <summary>
+        /// Adds an <see cref="Argument"/> to the command.
+        /// </summary>
+        /// <param name="argument">The argument to add to the command.</param>
         public void AddArgument(Argument argument) => AddArgumentInner(argument);
 
+        /// <summary>
+        /// Adds a subcommand to the command. Commands can be nested to an arbitrary depth.
```
jonsequitur, line 63: Maybe `Commands can be nested to an arbitrary depth` should be in `remarks`?

src/System.CommandLine/Command.cs
```diff
@@ -10,40 +10,83 @@
...
         public void AddCommand(Command command) => AddSymbol(command);
 
+        /// <summary>
+        /// Adds an <see cref="Option"/> to the command.
+        /// </summary>
+        /// <param name="option">The option to add to the command.</param>
         public void AddOption(Option option) => AddSymbol(option);
 
+        /// <summary>
+        /// Adds a global <see cref="Option"/> to the command. Global options are applied to all commands.
```
jonsequitur, line 75: Global options are only applied to subcommands, recursively, but not to parent commands.

src/System.CommandLine/Command.cs
```diff
@@ -10,40 +10,83 @@
...
         public void AddGlobalOption(Option option)
         {
             _globalOptions.Add(option);
             Children.AddWithoutAliasCollisionCheck(option);
         }
-
+        
+        /// <summary>
+        /// Adds a global <see cref="Option"/> to the command. Global options are applied to all commands. A
+        /// return value indicates if the option alias is already in use.
```
jonsequitur, line 86:
```suggestion
        /// return value indicates whether the option alias is already in use.
```

src/System.CommandLine/Command.cs
```diff
@@ -87,12 +143,28 @@ private protected override void AddSymbol(Symbol symbol)
...
         public bool TreatUnmatchedTokensAsErrors { get; set; } = true;
 
+        /// <summary>
+        /// Gets or sets the <see cref="ICommandHandler"/> for the command. The handler represents the action
+        /// that will be performed when the command is invoked.
+        /// </summary>
         public ICommandHandler? Handler { get; set; }
 
+        /// <summary>
+        /// Represents all the symbols associated with the command.
```
jonsequitur, line 166:
```suggestion
        /// Represents all of the symbols associated with the command.
```

jonsequitur:
Thanks for helping with this. It's a big task and every bit helps.

I suggested a few minor changes.

src/System.CommandLine/Command.cs
pieter-venter: OK, moved it

src/System.CommandLine/Command.cs
pieter-venter: I moved this to the remarks too, and expanded the text to be clear that it applies to the current command, subcommands but not parent commands. See what you think.

pieter-venter:
@jonsequitur Would you prefer it if I make a separate PR for each class? Writing these docs are surprisingly hard. I find myself rewriting things multiple times to make them consistent etc. May have a PR open for a very long time if I keep adding stuff.

jonsequitur:
Smaller PRs are always welcome. They're easier to review and quicker to merge.

Thanks!

pieter-venter:
> Smaller PRs are always welcome. They're easier to review and quicker to merge.
> 
> Thanks!

OK so shall we leave this PR "as is" and you can merge when happy?
I'll only commit additional PR feedback here.

I'm working on some others but will do that in a separate branches and create PRs when ready.

src/System.CommandLine/Argument.cs
```diff
@@ -119,6 +135,9 @@ public List<ISuggestionSource> Suggestions
             }
         }
 
+        /// <summary>
+        /// Gets or sets the type of the argument.
```
jonsequitur, line 139: Maybe something like `Gets or sets the <see cref="System.Type" /> that the argument will be converted to.`?

src/System.CommandLine/Argument.cs
```diff
@@ -146,8 +165,17 @@ private protected override string DefaultName
 
         internal List<ValidateSymbol<ArgumentResult>> Validators { get; } = new List<ValidateSymbol<ArgumentResult>>();
 
+        /// <summary>
+        /// Adds a custom <see cref="ValidateSymbol{T}(ArgumentResult)"/> to the argument. Validators can be used
+        /// to create custom validation logic.
```
jonsequitur, line 170:
```suggestion
        /// to provide custom errors based on user input.
```

src/System.CommandLine/Argument.cs
```diff
@@ -146,8 +165,17 @@ private protected override string DefaultName
...
+        /// </summary>
+        /// <param name="validate">The delegate to validate the symbols during parsing.</param>
```
jonsequitur, line 172:
```suggestion
        /// <param name="validate">The delegate to validate the parsed argument.</param>
```

src/System.CommandLine/Argument.cs
```diff
@@ -146,8 +165,17 @@ private protected override string DefaultName
...
         public void AddValidator(ValidateSymbol<ArgumentResult> validator) => Validators.Add(validator);
 
+        /// <summary>
+        /// Gets the default value for the argument.
+        /// </summary>
+        /// <returns>Returns the default value for the argument, if defined. Null otherwise.</returns>
```
jonsequitur, line 178: I believe by default it returns the default for the `ArgumentType`, not `null`. For example it might return `0` if `ArgumentType` is `int`.

src/System.CommandLine/Command.cs
```diff
@@ -87,18 +148,37 @@ private protected override void AddSymbol(Symbol symbol)
...
+        /// <summary>
+        /// Gets or sets a value that indicates whether unmatched tokens should be treated as errors. For example,
+        /// if set to <c>true</c> and an extra command or argument is provided, validation will fail.
+        /// </summary>
         public bool TreatUnmatchedTokensAsErrors { get; set; } = true;
 
+        /// <summary>
+        /// Gets or sets the <see cref="ICommandHandler"/> for the command. The handler represents the action
+        /// that will be performed when the command is invoked.
+        /// </summary>
```
jonsequitur, line 167: It might be worth adding a remark that command line input specifying commands that don't have handlers will result in parse errors.

src/System.CommandLine/Argument.cs
pieter-venter: I like it - and I just learned something new.

src/System.CommandLine/Argument.cs
pieter-venter: The "Null otherwise" is definitely incorrect, good catch.

However, looking at the code again, it will throw if you haven't set a default value. If you set a default value (or factory), then it doesn't need to find the default value for the type.

I can't see an obvious scenario where you can make this method return the default value for the ArgumentType type. Help!

src/System.CommandLine/RootCommand.cs
```diff
@@ -8,14 +8,19 @@
 namespace System.CommandLine
 {
     /// <summary>
-    /// A command representing an application entry point.
+    /// Represents the main action that the application performs.
```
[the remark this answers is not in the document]
pieter-venter: FYI This was done so the text is more in line with the Command.cs summary. The idea was to capture the difference between an app with one purpose vs ones with subcommands. A little tricky to capture everything succinctly.

The way I understand it (please help if I'm wrong)
1. RootCommand only, no handler = pointless
2. RootCommand + handler = app does one thing.
3. RootCommand + handler, SubCommand no handler = app does one thing, subcommand pointless.
4. RootCommand, no handler, SubCommand + handler = command must be specified.

Ideas welcome.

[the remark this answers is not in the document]
jonsequitur: Re: 1, 2, and 4, agreed. For 3, there's also the case to consider where there's no handler on the first level of subcommands but there are handlers on the next layer of subcommands, e.g.

```console
> dotnet add  # This is an error
> dotnet add package <package name> # This is not an error
```

src/System.CommandLine/Command.cs
pieter-venter: OK I also added a reference to the `CommandHandler.Create` method. For me, this was one of the trickiest things "discover". We may want to expand the documentation even more. Take a look.

jonsequitur (approved):

PR #1097, opened 2020-11-25
update tfm to net5.0

brettfo:
~~Leaving as draft until a [full internal build](https://dev.azure.com/dnceng/internal/_build/results?buildId=999891&view=logs&s=6884a131-87da-5381-61f3-d7acc3b91d76) has completed.~~ Internal ubuntu build passed.

Filed #1206 to handle linux cancellation test failure.


brettfo:
[This test](https://github.com/dotnet/command-line-api/blob/0581713635ab5930125c47d8a384d6594679849e/src/System.CommandLine.Tests/Invocation/CancelOnProcessTerminationTests.cs#L25) is failing because it's not getting the expected `OperationCanceledException` and the return code 42; it's instead returning code 130 which corresponds to Ctrl+C.  Looking around to see if/what changed between .NET 3 and 5.

src/System.CommandLine.Benchmarks/System.CommandLine.Benchmarks.csproj
```diff
@@ -9,8 +9,8 @@
     <UseSharedCompilation>false</UseSharedCompilation>
 
     <!-- Supported target frameworks -->
-    <TargetFrameworks Condition="'$(TargetFrameworks)' == '' AND '$(OS)' == 'Windows_NT'">net461;netcoreapp3.1;</TargetFrameworks>
-    <TargetFrameworks Condition="'$(TargetFrameworks)' == ''">netcoreapp3.1;</TargetFrameworks>
+    <TargetFrameworks Condition="'$(TargetFrameworks)' == '' AND '$(OS)' == 'Windows_NT'">net461;net5.0;</TargetFrameworks>
+    <TargetFrameworks Condition="'$(TargetFrameworks)' == ''">net5.0;</TargetFrameworks>
```
jonsequitur, line 13: Maybe this should be an additive change to we can compare against  the .NET Core 3.1 performance?

Not sure how interesting that really is.

jonsequitur (approved):

brettfo:
Annoyingly this test passes locally.

PR #1099, opened 2020-11-28
XML documentation for ArgumentArity class

pieter-venter:


src/System.CommandLine/ArgumentArity.cs
```diff
@@ -7,10 +7,21 @@
 
 namespace System.CommandLine
 {
+    /// <inheritdoc />
     public class ArgumentArity : IArgumentArity
     {
+        /// <summary>
+        /// The overall maximum number of values that can be provided to any argument.
+        /// </summary>
```
jonsequitur, line 15: I think this should actually be an internal `const` and not require a doc comment.

If you'd like to make the change, feel free to just specify it literally in the tests currently referencing it.

src/System.CommandLine/ArgumentArity.cs
```diff
@@ -72,14 +85,29 @@ public ArgumentArity(int minimumNumberOfValues, int maximumNumberOfValues)
             return null;
         }
 
+        /// <summary>
+        /// Must not have any arguments.
```
jonsequitur, line 89: Maybe something like: "An arity that does not allow any values."

src/System.CommandLine/ArgumentArity.cs
```diff
@@ -72,14 +85,29 @@ public ArgumentArity(int minimumNumberOfValues, int maximumNumberOfValues)
...
+        /// </summary>
         public static IArgumentArity Zero => new ArgumentArity(0, 0);
 
+        /// <summary>
+        /// May have one argument, but no more than one.
```
jonsequitur, line 94: Since we have the properties `MinimumNumberOfValues` and `MaximumNumberOfValues`, these comments should probably refer to "values" rather than "arguments".

src/System.CommandLine/ArgumentArity.cs
pieter-venter: OK just to confirm, change the modifier to `internal` and remove the XML comment.
Since that makes it inaccessible to the tests, you want to change this to a literal (12 places in 2 classes)?
``` csharp
[InlineData(0, ArgumentArity.MaximumArity, typeof(string[]))]
``` 
to
``` csharp
[InlineData(0, 100_000, typeof(string[]))]
```
or maybe better to add a const to each of the tests and keep things a bit DRY?

``` csharp
const int MaximumArity = 100_000;


[InlineData(0, MaximumArity, typeof(string[]))]
```

pieter-venter: On second thought, I guess you mean just remove the `public` modifier and make it a private const to the class?

jonsequitur: Yes. I don't think DRY in test code is as important as keeping the library API surface smaller, and in the case of these tests, the specific value isn't very important. Basically, this is a large number that's lower than `int.MaxValue` so we don't overflow.

pieter-venter: Done

src/System.CommandLine/ArgumentArity.cs
pieter-venter: Done

src/System.CommandLine/ArgumentArity.cs
pieter-venter: Agreed. I've changed them all.

jonsequitur (approved):

PR #1100, opened 2020-11-28
XML documentation for Argument<T> class

pieter-venter:


src/System.CommandLine/Argument{T}.cs
```diff
@@ -20,6 +29,13 @@ public Argument()
             Description = description;
         }
 
+        /// <summary>
+        /// Initializes a new instance of the Argument class.
+        /// </summary>
+        /// <param name="name">The name of the Argument.</param>
```
jonsequitur, line 35:
```suggestion
        /// <param name="name">The name of the argument.</param>
```

src/System.CommandLine/Argument{T}.cs
```diff
@@ -45,6 +66,13 @@ public Argument(Func<T> getDefaultValue) : this()
             SetDefaultValueFactory(() => getDefaultValue());
         }
 
+        /// <summary>
+        /// Initializes a new instance of the Argument class.
+        /// </summary>
+        /// <param name="name">The name of the argument.</param>
+        /// <param name="parse">A custom argument parser.</param>
+        /// <param name="isDefault"><c>true</c> to use parser result as default value.</param>
```
jonsequitur, line 74: This should probably include a reference to the `parse` parameter on this constructor, as it might be a little ambiguous otherwise.

src/System.CommandLine/Argument{T}.cs
```diff
@@ -82,6 +110,11 @@ public Argument(Func<T> getDefaultValue) : this()
             };
         }
 
+        /// <summary>
+        /// Initializes a new instance of the Argument class.
+        /// </summary>
+        /// <param name="parse">A custom argument parser.</param>
+        /// <param name="isDefault"><c>true</c> to use parser result as default value.</param>
```
jonsequitur, line 117: Same as above.

src/System.CommandLine/Argument{T}.cs
pieter-venter: Done

src/System.CommandLine/Argument{T}.cs
pieter-venter: Changed it, hopefully this is what you meant.

jonsequitur (approved):

PR #1101, opened 2020-11-28
XML documentation for IArgument interface

pieter-venter:


src/System.CommandLine/IArgument.cs
```diff
@@ -6,11 +6,17 @@
 
 namespace System.CommandLine
 {
+    /// <summary>
+    /// Defines a symbol with an arity.
```
jonsequitur, line 10: There might be some useful terminology to reuse here: https://github.com/dotnet/command-line-api/blob/main/docs/Syntax-Concepts-and-Parser.md#arguments-and-arity.

In this case I'd say something like `A symbol representing a value passed to an option or command`.

sharwell: Agree that @jonsequitur 's suggestion is clearer than current comment. Two things to consider:

1. A symbol ~representing~ **defining** a value
2. Consider linking `<see cref="IOption"/>` and `<see cref="ICommand"/>` if applicable, or if the interfaces have different names you could also use the form `<see cref="IOptionInterfaceName">option</see>`

jonsequitur (approved):

PR #1102, opened 2020-11-28
XML documentation for IArgumentArity interface

pieter-venter:


src/System.CommandLine/IArgumentArity.cs
```diff
@@ -3,10 +3,19 @@
 
 namespace System.CommandLine
 {
+    /// <summary>
+    /// Defines the arity of an argument.
```
jonsequitur, line 7: We could add a remark here indicating what this actually means, e.g. https://github.com/dotnet/command-line-api/blob/main/docs/Syntax-Concepts-and-Parser.md#arguments-and-arity

pieter-venter: Done. The explanation seems to fit verbatim so I added it in as-is.

jonsequitur (approved):

jonsequitur:
Thanks @pieter-venter!

PR #1103, opened 2020-11-29
XML documentation for ICommand interface

pieter-venter:


src/System.CommandLine/ICommand.cs
```diff
@@ -5,12 +5,24 @@
 
 namespace System.CommandLine
 {
+    /// <summary>
+    /// Defines a symbol with arguments and options.
```
jonsequitur, line 9: We might want to align to the definitions [here](https://github.com/dotnet/command-line-api/blob/main/docs/Syntax-Concepts-and-Parser.md#arguments-and-arity) and maybe add a remark explaining that this is also sometimes known as a "verb".

pieter-venter: @jonsequitur I would like some input on this. My thought process:

I kept the interface descriptions very abstract here, just describing literally what they advertise. After all, we can't be sure what an implementation of ICommand will do other than it has arguments and options (and one getter).  Admittedly, this can lead to some pretty unspecific summaries as can be seen here. The implementations, like Command.cs, aligns closer with the [docs](https://github.com/dotnet/command-line-api/blob/main/docs/Syntax-Concepts-and-Parser.md#commands), because we _know_ what it does.

We _can_ provide comprehensive documentation in the interface and then use inheritdoc in the implementation to avoid copy/paste. However, in many instances this is just not possible since the implementation derives from multiple interfaces.

In your option, what should be in the ICommand summary and what should be in Command? Should they be mostly identical or is there some way to more clearly define what to put in each? It seems more appropriate to me that the fact that a subcommand is sometimes called a verb is on the Command description.

My feeling was most users of this library are going to care about the docs for Command, Argument, Option etc and less about ICommand, IArgument. Another option is to just remove the summary doc from the Interfaces that are either just going to be a duplicate of the implementation or otherwise difficult to describe.

src/System.CommandLine/ICommand.cs
```diff
@@ -5,12 +5,24 @@
...
+        /// Gets a value that indicates whether unmatched tokens should be treated as errors.
+        /// </summary>
         bool TreatUnmatchedTokensAsErrors { get; }
 
+        /// <summary>
+        /// Gets the arguments for the command.
+        /// </summary>
         IEnumerable<IArgument> Arguments { get; }
         
+        /// <summary>      `
```
sharwell, line 23:
```suggestion
        /// <summary>
```

src/System.CommandLine/ICommand.cs
sharwell: If you have any meetings on this, can you invite me? Documentation strategies is a personal passion project.

jonsequitur: My take is the descriptions should be mostly identical. The reason we separated out `ICommand`, etc. from their concrete implementations wasn't because we expect someone to implement them, but because we wanted a way to implement some form of immutability once the parser is constructed, without adding the performance overhead of cloning or freezing. It's not an ideal solution and we don't have tests covering other implementations of these interfaces, so there's not much purpose in overgeneralizing.

jonsequitur (approved):

PR #1104, opened 2020-11-29
XML documentation for IIdentifierSymbol interface

pieter-venter:


src/System.CommandLine/IIdentifierSymbol.cs
```diff
@@ -5,10 +5,21 @@
 
 namespace System.CommandLine
 {
+    /// <summary>
+    /// Defines a symbol with aliases.
```
jonsequitur, line 9:
```suggestion
    /// Defines a symbol which is an identifier in the command line grammar, such as a command or option.
```

src/System.CommandLine/IIdentifierSymbol.cs
```diff
@@ -5,10 +5,21 @@
...
+        /// <summary>
+        /// Gets the aliases for the symbol.
+        /// </summary>
         IReadOnlyCollection<string> Aliases { get; }
 
+        /// <summary>
+        /// Determines whether the alias has already been defined.
+        /// </summary>
+        /// <param name="alias">The alias to search for.</param>
+        /// <returns><c>true</c> if the alias has already been defined; otherwise <c>false</c>.</returns>
```
sharwell, line 22:
```suggestion
        /// <returns><see langword="true"/> if the alias has already been defined; otherwise, <see langword="false"/>.</returns>
```

💡 If you type `true` and then press <kbd>Ctrl</kbd>+<kbd>Space</kbd>, the IDE will automatically expand it to `<see langword="true"/>`.

jonsequitur (approved):

PR #1105, opened 2020-11-29
XML documentation for ISymbol interface

pieter-venter:


src/System.CommandLine/ISymbol.cs
```diff
@@ -6,16 +6,34 @@
 
 namespace System.CommandLine
 {
+    /// <summary>
+    /// Defines a named symbol that resides in a hierarchy with parent and child symbols.
```
jonsequitur, line 10: Confusingly enough, we have a `NamedSymbol` class which `Argument` doesn't implement. That probably needs renaming (since `Argument` actually does have a `Name`).

The `IIdentifierSymbol` is worth thinking about here. Basically, options and arguments represent identifiers in the command-line's grammar; arguments do not.

jonsequitur: See #1122

jonsequitur (approved):

PR #1108, opened 2020-11-30
XML documentation for IOption interface

pieter-venter:


src/System.CommandLine/IOption.cs
```diff
@@ -5,12 +5,24 @@
 
 namespace System.CommandLine
 {
+    /// <summary>
+    /// Defines a symbol with an argument.
```
jonsequitur, line 9: We can align this to the terminology here: https://github.com/dotnet/command-line-api/blob/main/docs/Syntax-Concepts-and-Parser.md#arguments-and-arity.

src/System.CommandLine/IOption.cs
```diff
@@ -5,12 +5,24 @@
...
+        /// </summary>
         IArgument Argument { get; }
 
+        /// <summary>
+        /// Gets a value that indicates whether the option is required.
+        /// </summary>
         bool IsRequired { get; }
 
+        /// <summary>
+        /// Gets a value that indicates whether multiple argument values are allowed.
```
jonsequitur, line 24: This is a little more specific and a little hard to explain so it'll deserve a remark with an example, I think. 

What this is really indicating is whether multiple argument tokens can be passed for a single option token, like this (when it's set to `true`):

```csharp
> my.exe --option 1 2 3
```

or whether (when set to `false`) people need to repeat the option token, like this:

```csharp
> my.exe --option 1 --option 2 --option  3
```

In both cases, assuming `--option` is an `Option<int[]>`, the value will be `new [] { 1, 2, 3 }`. 

The naming could possibly be made clearer. Feel free to make suggestions.

jonsequitur (approved):

PR #1109, opened 2020-11-30
XML documentation ISuggestionSource interface

pieter-venter:


src/System.CommandLine/Suggestions/ISuggestionSource.cs
```diff
@@ -6,8 +6,17 @@
 
 namespace System.CommandLine.Suggestions
 {
+    /// <summary>
+    /// Defines the functionality to suggest values.
```
jonsequitur, line 10: Maybe something like "Provides suggestions for tab completion and example values for help", to give a sense of when we might suggest values.

jonsequitur (approved):

jonsequitur:
Thanks, @pieter-venter!

PR #1113, opened 2020-12-01
invert version selector properties

brettfo:
To temporarily work around issue dotnet/arcade#6616, this PR inverts the version selector stuff so that by default `AutoGenerateAssemblyVersion` = `true` and `PreReleaseVersionLabel` is empty to enable publishing a stable package to a non-isolated feed.

The next immediate section effectively reverts those values if `UseGlobalToolVersion` isn't `true` (which is all projects _except_ `dotnet-suggest.csproj`) and finally the next 2 sections override those values again if `UseAlphaVersion` or `UseBetaVersion` is true.

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -6,22 +6,18 @@
   </PropertyGroup>
   <!-- Repo Version Information -->
   <PropertyGroup>
-    <!-- Default version number. -->
+    <!-- Default version number to global tool. -->
     <VersionPrefix>0.1.0</VersionPrefix>
-    <PreReleaseVersionLabel>beta</PreReleaseVersionLabel>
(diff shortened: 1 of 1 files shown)


epananth:
You are still using https://github.com/dotnet/command-line-api/blob/main/src/System.CommandLine/Directory.Build.props to use 
UseBetaVersion. This will set PreReleaseVersionLabel back to "beta".

Can you run a build with this change so that we can what the value of PreReleaseVersionLabel? I'm guessing it going to be beta again and it will fail.

jonsequitur (approved):

brettfo:
[Internal build](https://dev.azure.com/dnceng/internal/_build/results?buildId=905159&view=results) has the expected properties and the generated package versions were:

image

compared to the last good build:

image

PR #1124, opened 2020-12-10
fix #1050

### LINKED ISSUE #1050: `--version` does not short circuit flow (unlike `--help` which does)
I noticed a difference in behaviour and like to have a clarification if it is a bug or works as intended.
If you invoke an app with `--help [command]` the command is not executed and the help of the command is shown.
If you invoke an app with `--version [command]` the command is executed and no version information is shown.

This is somewhat confusing.

Some example code:
```csharp
using System;
using System.CommandLine;
using System.IO;
using System.Threading.Tasks;

namespace Test.CommandLine
{
    class Program
    {
        {
            var rootCommand = new RootCommand("Description of rootCommand");

            AddSubCommand(rootCommand);

            // Parse the incoming args and invoke the handler
            return await rootCommand.InvokeAsync(args);
        }

        private static void AddSubCommand(RootCommand rootCommand)
        {
            var subCommand = new Command("sub", "Description of sub command")
            {
                new Option<DirectoryInfo>("--foo", "a dummy directory")
                {
                    IsRequired = true,
                    Argument = new Argument<DirectoryInfo>()
                    {
                        Arity = ArgumentArity.ExactlyOne
                    }
                },
                new Option("--bar", "a dummy flag")
                {
                    IsRequired = false,
                    Argument = new Argument<bool>()
                    {
                        Arity = ArgumentArity.ZeroOrOne
                    }
                }
            };

            subCommand.Handler = CommandHandler.Create<DirectoryInfo, bool, IConsole>((foo, bar, console) =>
            {
                console.Out.Write($"sub command was executed with values:{Environment.NewLine}\t{nameof(foo)}: \"{foo}\"{Environment.NewLine}\t{nameof(bar)}: {bar}");
            });

            rootCommand.AddCommand(subCommand);
        }
    }
}
```

Examples of invocation and shown result:
```console
.\App.exe --help sub --foo "C:\Windows"
sub:
  Description of sub command

Usage:
  Test.CommandLine sub [options]

Options:
  --foo <foo> (REQUIRED)    a dummy directory
  --bar                     a dummy flag
  -?, -h, --help            Show help and usage information
```

```console
.\App.exe --version sub --foo "C:\Windows"
sub command was executed with values:
        foo: "C:\Windows"
        bar: False
```

jonsequitur:
This fixes #1050.

### DIFF
--- a/src/System.CommandLine/Builder/CommandLineBuilderExtensions.cs
+++ b/src/System.CommandLine/Builder/CommandLineBuilderExtensions.cs
@@ -458,20 +458,41 @@ namespace System.CommandLine.Builder
         public static CommandLineBuilder UseVersionOption(
             this CommandLineBuilder builder)
         {
-            if (builder.Command.Children.GetByAlias("--version") != null)
+            var command = builder.Command;
+
+            if (command.Children.GetByAlias("--version") != null)
             {
                 return builder;
             }
 
-            var versionOption = new Option("--version", "Show version information");
+            var versionOption = new Option<bool>(
+                "--version",
+                description: "Show version information",
+                parseArgument: result =>
+                {
+                    if (result.FindResultFor(command)?.Children.Count > 1)
+                    {
+                        result.ErrorMessage = "--version option cannot be combined with other arguments.";
+                        return false;
+                    }
 
-            builder.AddOption(versionOption);
+                    return true;
+                });
+
+            command.AddOption(versionOption);
 
             builder.AddMiddleware(async (context, next) =>
             {
-                if (context.ParseResult.HasOption(versionOption))
+                if (context.ParseResult.FindResultFor(versionOption) is { } result)
                 {
-                    context.Console.Out.WriteLine(_assemblyVersion.Value);
+                    if (result.ArgumentConversionResult.ErrorMessage is { })
+                    {
+                        context.InvocationResult = new ParseErrorResult();
+                    }
+                    else
+                    {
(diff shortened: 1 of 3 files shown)


src/System.CommandLine/Builder/CommandLineBuilderExtensions.cs
```diff
@@ -458,20 +458,41 @@ public static TBuilder UseHelpBuilder<TBuilder>(this TBuilder builder, Func<Bind
...
                 return builder;
             }
 
-            var versionOption = new Option("--version", "Show version information");
+            var versionOption = new Option<bool>(
+                "--version",
+                description: "Show version information",
+                parseArgument: result =>
+                {
+                    if (result.FindResultFor(command)?.Children.Count > 1)
```
Keboo, line 473: Does the --help also produce an error if combined with other options?
Not really opposed to this behavior, but I do think the --version and --help behavior should match.

Keboo (approved):

src/System.CommandLine/Builder/CommandLineBuilderExtensions.cs
jonsequitur: `--help` is a little different in that it's valid for every single command, whereas `-version` isn't. So:

```console
> command subcommand --version #invalid
> command subcommand --help #valid
```

But for the cases where there additional tokens after `--help` we should maybe consider a similar error.

PR #1131, opened 2020-12-11
Allowing for customizing error result codes

### LINKED ISSUE #331: Default error exit codes
When System.CommandLine encounters an error, the program stops with an exit code 1.

It would be nice if this would be configurable.

Perhaps it makes sense for the package to return different values for different types of errors by default (e.g. unhandled exception vs parser error). Posix shells use 'exit codes' higher than 128 for indicating process termination by signal. Picking values below that is preferable.

Keboo:
Fixes #331

### DIFF
--- a/src/System.CommandLine.Tests/ParseDirectiveTests.cs
+++ b/src/System.CommandLine.Tests/ParseDirectiveTests.cs
@@ -82,5 +82,25 @@ namespace System.CommandLine.Tests
 
             exitCode.Should().Be(1);
         }
+
+        [Fact]
+        public async Task When_there_are_errors_then_parse_directive_sets_exit_code_to_custom_value()
+        {
+            var command = new RootCommand
+            {
+                new Option("-x")
+                {
+                    Argument = new Argument<int>()
+                }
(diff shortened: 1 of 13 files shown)


src/System.CommandLine.Tests/ParseDirectiveTests.cs
```diff
@@ -82,5 +82,25 @@ public async Task When_there_are_errors_then_parse_directive_sets_exit_code_1()
 
             exitCode.Should().Be(1);
         }
+
+        [Fact]
+        public async Task When_there_are_errors_then_parse_directive_sets_exit_code_to_custom_value()
+        {
+            var command = new RootCommand
+            {
+                new Option("-x")
```
jonsequitur, line 91: Please use `Option<int>` so there's less code to refactor assuming we remove the `Argument` property.

Keboo: Good point. Fixed up all of the tests in this file.

jonsequitur (approved):

PR #1139, opened 2020-12-17
WIP showing localizability.

Keboo:
@sharwell and @jonsequitur Looking to get feedback before pushing further.

The initial discussion suggested using [LocalizableResourceString](https://github.com/dotnet/roslyn/blob/cf55f3a58e47298426fa971d3bd9d8857c746c65/src/Compilers/Core/Portable/Diagnostic/LocalizableResourceString.cs) from [Microsoft.CodeAnalysis.Common](https://www.nuget.org/packages/Microsoft.CodeAnalysis.Common/) based on the documentation [here](https://github.com/dotnet/roslyn/blob/master/docs/analyzers/Localizing%20Analyzers.md). 

A few things I ran into while doing this:
1. LocalizableResourceString is not CLS Compliant (something that was only recently enabled with  #1130)
2. Arcade appears to be interesting things when there are resx files present. It seems to be working correctly given all of the satellite assemblies are generated, but I will need someone more familiar with that setup to confirm.
3. Reading through the documentation on the `LocalizableResourceString` I _think_ it is designed to solve a problem that we don't have here. Specifically it offers support for switching the output language of the compiler using the [preferreduilang](https://docs.microsoft.com/dotnet/csharp/language-reference/compiler-options/preferreduilang-compiler-option). The internals of which end up just calling into the `ResourceManager` to get the localized string. Unlike analyzers, that function more like plugins to the compilers, we simply have a library so I _think_ we might be better of simply going to the generated `ResourceManager` directly. This also saves on needing to take a dependency on `Microsoft.CodeAnalysis.Common`


sharwell:
> LocalizableResourceString is not CLS Compliant

Is there a need to enable CLS compliance? My understanding is this is generally unnecessary and even when implemented is unlikely to provide value to consumers. /cc @jaredpar 

> Unlike analyzers, that function more like plugins to the compilers, we simply have a library so I think we might be better of simply going to the generated ResourceManager directly.

If one of the goals is configuring the resource language (even if that means `invariant`) from the command line, we may be in a position that looks more like Roslyn's.

> This also saves on needing to take a dependency on Microsoft.CodeAnalysis.Common

Rather than depend on Microsoft.CodeAnalysis.Common, I was suggesting we create a copy of LocalizableResourceString as part of this repository.

jaredpar:
> Is there a need to enable CLS compliance? My understanding is this is generally unnecessary and even when implemented is unlikely to provide value to consumers

I don't know if there is an official policy on this. In general though I don't believe our repos, outside of possibly runtime, are using `CLSCompliance`. Possible I'm just looking at the outliers but don't believe so. 

> Marking the assembly as CLSCompliant(false) resolves downstream warnings (even in CLS compliant applications) and also makes the compilation process faster

Indeed, this has a measurable impact on compiler performance.

jonsequitur:
I think it's fine to remove the CLS compliance.

Keboo:
Ok more updates with some additional questions.
1. Is my copy of LocalizableResourceString correct. The one from the Roselyn lib also includes a bunch of stuff for `IObjectWritable` that I removed. Do we need that?
2. How high up should the LocalizableResourceString be propagated. Right now with the single resource I have done, we are converting to a string in the validation delegate. Should the validation delegate be changed to use LocalizableResourceString or remain as System.String?
3. Is it still acceptable to use the ValidationMessages class as the entry point for getting these messages?
4. Should we rename ValidationMessages or does this name still seem appropriate?

src/System.CommandLine/Properties/Resources.Designer.cs
```diff
@@ -0,0 +1,72 @@
...
+// <auto-generated>
+//     This code was generated by a tool.
+//     Runtime Version:4.0.30319.42000
+//
+//     Changes to this file may cause incorrect behavior and will be lost if
+//     the code is regenerated.
+// </auto-generated>
+//------------------------------------------------------------------------------
+
+namespace System.CommandLine.Properties {
```
sharwell, line 11: ❔ Is this repository using Arcade? If so, this file is not required.

jonsequitur: Can you elaborate on that? It does use Arcade, however Arcade is not required for developers. We have an environment variable that can disable many of the Arcade customizations on developer machines.

src/System.CommandLine/ValidationMessages.cs
```diff
@@ -29,15 +29,19 @@ public virtual string ExpectsOneArgument(SymbolResult symbolResult)
...
+            Token token,
             int providedNumberOfValues,
             int maximumNumberOfValues) =>
             token.Type == TokenType.Command
                 ? $"Command '{token}' expects no more than {maximumNumberOfValues} arguments, but {providedNumberOfValues} were provided."
                 : $"Option '{token}' expects no more than {maximumNumberOfValues} arguments, but {providedNumberOfValues} were provided.";
 
-        public virtual string DirectoryDoesNotExist(string path) =>
-            $"Directory does not exist: {path}";
+        public virtual LocalizableResourceString DirectoryDoesNotExist(string path) =>
```
sharwell, line 39:
```suggestion
        public virtual LocalizableString DirectoryDoesNotExist(string path) =>
```

src/System.CommandLine/System.CommandLine.csproj
```diff
@@ -28,4 +28,19 @@
...
+      <DependentUpon>Resources.resx</DependentUpon>
+    </Compile>
+  </ItemGroup>
+
+  <ItemGroup>
+    <EmbeddedResource Update="Properties\Resources.resx">
+      <Generator>ResXFileCodeGenerator</Generator>
+      <LastGenOutput>Resources.Designer.cs</LastGenOutput>
+    </EmbeddedResource>
+  </ItemGroup>
```
sharwell, line 44:
```suggestion
  <ItemGroup>
    <EmbeddedResource Update="Properties\Resources.resx" GenerateSource="true" />
  </ItemGroup>
```
```

**Comment by sharwell** [comment]
The main decision to make at this point is the source of language used for messages:

* You can use the current thread's culture info
* You can use some other application-specific source

If the former is acceptable, then this pull request is overly complex and we can just reference resource strings wherever we need a message. If the latter is important, we can start to look into the API impact of a pull request like this one. It will likely help to bring in someone who helped with the compiler localization process to identify the key issues for decision making.

File: src/System.CommandLine/Properties/Resources.Designer.cs (hunk 1 above)
Comment (Keboo) [reply src/System.CommandLine/Properties/Resources.Designer.cs]: Based on testing locally, there is stuff in the MSBuild pipeline that handles generating the needed code behind (similar to the what the .Designer.cs was doing with the embedded tool). Doesn't look like there is any additional overhead for developers so I just pushed the update.

**Comment by jonsequitur** [comment]
> The main decision to make at this point is the source of language used for messages:
> * You can use the current thread's culture info
> * You can use some other application-specific source

I think the first option is acceptable. The host, or a common middleware, can set the thread culture as needed for application-specific use cases.

File: src/System.CommandLine/LocalizableResourceString.cs [hunk 4]
```diff
@@ -0,0 +1,83 @@
...
+using System.Linq;
+using System.Resources;
+
+namespace System.CommandLine
+{
+    /// <summary>
+    /// A localizable resource string that may possibly be formatted differently depending on culture.
+    /// Copied from https://github.com/dotnet/roslyn/blob/cf55f3a58e47298426fa971d3bd9d8857c746c65/src/Compilers/Core/Portable/Diagnostic/LocalizableResourceString.cs
+    /// </summary>
+    public sealed class LocalizableResourceString : LocalizableString
```
jonsequitur, line 16: Is there any reason these need to be in the root namespace? I'd prefer to keep it uncluttered.

Keboo: Nope, I am happy to move it. Any suggestions on a name?

jonsequitur: Maybe move it into the `System.CommandLine.Help` namespace so as not to introduce a new namespace just for this?

sharwell:
@Keboo my recommendation is to mark this pull request as a draft and close it. If we need this functionality, we can bring it back. Then open a new pull request that just does the following:

1. Adds the .resx file, xlf files, and project file change
2. Modifies `ValidationMessages.DirectoryDoesNotExist` to do this:

    ```csharp
    => string.Format(Resources.DirectoryDoesNotExist, path)`
    ```

PR #1141, opened 2020-12-18
Making the debug directive opt-in

### LINKED ISSUE #468: provide a way for end users to disable [debug] directive
In production environments the blocking behavior of the `[debug]` directive could be a liability. Consider enabling or disabling it via an environment variable or other mechanism.

Keboo:
One possible solution for addressing the debug directive.

Some questions:
1. Is the environment variable name `SYSTEM_COMMANDLINE_DEBUG_PROCESSES` acceptable?
2. Currently the messages if you specify the debug directive but have not opted-in with the environment variable are written to standard error. Is that correct?
3. Should error conditions like that short-circuit or continue calling through the pipeline? If they should short-circuit what should the exit code be?
4. I added an optional timeout, with a default of infinite, should this default be changed?

Fixes #468

### DIFF
--- a/src/System.CommandLine/Builder/CommandLineBuilderExtensions.cs
+++ b/src/System.CommandLine/Builder/CommandLineBuilderExtensions.cs
@@ -21,7 +21,8 @@ namespace System.CommandLine.Builder
     public static class CommandLineBuilderExtensions
     {
         private static readonly Lazy<string> _assemblyVersion =
-            new Lazy<string>(() => {
+            new Lazy<string>(() =>
+            {
                 var assembly = Assembly.GetEntryAssembly() ?? Assembly.GetExecutingAssembly();
                 var assemblyVersionAttribute = assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>();
                 if (assemblyVersionAttribute is null)
@@ -65,7 +66,7 @@ namespace System.CommandLine.Builder
         }
 
         public static TBuilder AddGlobalOption<TBuilder>(
-            this TBuilder builder, 
+            this TBuilder builder,
             Option option)
             where TBuilder : CommandBuilder
         {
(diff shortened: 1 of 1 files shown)


src/System.CommandLine/Builder/CommandLineBuilderExtensions.cs
```diff
@@ -205,20 +206,47 @@ public static CommandLineBuilder CancelOnProcessTermination(this CommandLineBuil
...
                 {
-                    var process = Diagnostics.Process.GetCurrentProcess();
+                    const string environmentVariableName = "SYSTEM_COMMANDLINE_DEBUG_PROCESSES";
 
-                    var processId = process.Id;
+                    string debuggableProcessNames = GetEnvironmentVariable(environmentVariableName);
+                    if (string.IsNullOrWhiteSpace(debuggableProcessNames))
+                    {
+                        context.Console.Error.WriteLine("Debug directive specified, but no process names are listed as allowed for debug.");
+                        context.Console.Error.WriteLine($"Add your process name to the '{environmentVariableName}' environment variable.");
```
jonsequitur, line 225: This looks like it deserves a bit more detail, as the convention around multiple process names isn't evident.

jonsequitur:
> 1. Is the environment variable name SYSTEM_COMMANDLINE_DEBUG_PROCESSES acceptable?

We've been tending toward a `DOTNET` prefix as the general audience of end users may not be aware of System.CommandLine, which is an implementation detail. That said, this is a developer feature, so I'd say that seems fine.  

> 2. Currently the messages if you specify the debug directive but have not opted-in with the environment variable are written to standard error. Is that correct?

I think these should be written to standard out, if they're not actually errors, but...

> 3. Should error conditions like that short-circuit or continue calling through the pipeline? If they should short-circuit what should the exit code be?

It might be clearer if this just errors out. People won't be using it accidentally, so they probably don't want the command to run.

> 4. I added an optional timeout, with a default of infinite, should this default be changed?

Given the additional guard rails, maybe the timeout is just more complexity than we need?

jonsequitur (approved):

PR #1145, opened 2020-12-19
Perf tuning

jonsequitur:


src/System.CommandLine/Binding/ArgumentConverter.cs
```diff
@@ -106,59 +106,87 @@ internal static class ArgumentConverter
...
+            return Success(argument, values);
 
-            return Success(argument, value);
+            static IList CreateList(Type itemType, int capacity)
+            {
+                if (itemType == typeof(string))
+                {
+                    return new List<string>(capacity);
+                }
+                else if (itemType == typeof(int))
```
Keboo, line 163: Why are we optimizing for int?

src/System.CommandLine/Binding/ArgumentConverter.cs
```diff
@@ -323,5 +350,45 @@ internal static T GetValueOrDefault<T>(this ArgumentConversionResult result)
                     return default!;
             }
         }
+
+        public static bool TryConvertBoolArgument(ArgumentResult argumentResult, out object? value)
```
Keboo, line 354: Is there a reason the out parameter is `object?` not `bool`?

```suggestion

        public static bool TryConvertBoolArgument(ArgumentResult argumentResult, out bool value)
```

src/System.CommandLine/Binding/Binder.cs
```diff
@@ -19,29 +20,53 @@ internal static bool IsMatch(this string parameterName, string alias)
...
+                parameterNameIndex++;
             }
 
             return true;
+
+            static int IndexAfterPrefix(string alias)
+            {
+                if (alias.Length > 0)
+                {
+                    if (alias[0] == '-' && (alias.Length > 1 && alias[1] == '-'))
```
Keboo, line 54:
```suggestion
                    if (alias[0] == '-' && alias.Length > 1 && alias[1] == '-')
```

src/System.CommandLine/Parsing/SymbolResult.cs
```diff
@@ -19,6 +20,11 @@ public abstract class SymbolResult
             Symbol = symbol ?? throw new ArgumentNullException(nameof(symbol));
 
             Parent = parent;
+
+            if (parent?.Root is {} root)
+            {
+                _root = root;
+            }
```
Keboo, line 27: Is thee any reason we can't directly assign rather than do the null check?
```suggestion
            _root = parent?.Root;
```

Keboo (approved):

src/System.CommandLine/Binding/ArgumentConverter.cs
jonsequitur: It's an experiment. `Activator.CreateInstance` sticks out as a perf problem, so I was trying to see if there are ways to reduce that hit. Of course this approach is optimized for a specific and maybe not common use case, so we can't really extend this to many types and maybe shouldn't bother at all.

Other ideas?

src/System.CommandLine/Binding/ArgumentConverter.cs
jonsequitur: Good catch! Just a refactoring artifact.

PR #1147, opened 2020-12-20
Fix1125 localization

Keboo:
Based on feedback in #1139 
This includes just the resx files as well and updating ValidationMessage to use them.

/cc: @sharwell


jonsequitur (approved):

sharwell (commented):
This approach seems like a safe start.

Issue #1128, opened 2020-12-11
Consider making Option.Argument an internal implementation detail


jonsequitur:
Making `Option.Argument` an internal implementation detail would reduce some ambiguities and potentially make the API cleaner. 

See: #836, #1010.

PR #1148, opened 2020-12-24
Making argument internal to option

Keboo:
Still need to review new public API of option after this change.
Looking for feedback on new properties and constructors.

Fixes: #1128

### DIFF
--- a/src/System.CommandLine.DragonFruit.Tests/ConfigureFromMethodTests.cs
+++ b/src/System.CommandLine.DragonFruit.Tests/ConfigureFromMethodTests.cs
@@ -220,7 +220,7 @@ namespace System.CommandLine.DragonFruit.Tests
             var options = handlerMethod.BuildOptions();
 
             options.Should()
-                   .NotContain(o => o.Argument.ArgumentType == type);
+                   .NotContain(o => o.ArgumentType == type);
         }
 
         [Fact]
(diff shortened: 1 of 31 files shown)


src/System.CommandLine/Option.cs
```diff
@@ -15,12 +15,29 @@ public class Option :
...
-        {
-        }
-
-        public Option(string[] aliases, string? description = null) : base(description)
+        public Option(
+            string alias,
+            string? description = null,
+            Type? argumentType = null,
+            Func<object?>? getDefaultValue = null,
+            IArgumentArity? arity = null)
```
jonsequitur, line 23: Should this parameter be called `argumentArity`? It's more aligned with e.g. `argumentType`. 

Or should we go the other way and remove the term "argument" from the public interface of `Option`? Is it confusing?

src/System.CommandLine.DragonFruit.Tests/ConfigureFromMethodTests.cs
```diff
@@ -220,7 +220,7 @@ public void Options_are_not_built_for_infrastructure_types_exposed_by_method_par
             var options = handlerMethod.BuildOptions();
 
             options.Should()
-                   .NotContain(o => o.Argument.ArgumentType == type);
+                   .NotContain(o => o.ArgumentType == type);
```
jonsequitur, line 223: I wonder if `ValueType` would be a better name for this property. That would match the `IValueDescriptor` interface, for what it's worth.

jonsequitur (approved):

PR #1159, opened 2021-01-06
Allow CommandHandler to accept 16 arguments

Ssloto:
This PR modifies `System.CommandLine.Invocation.CommandHandler.Create` so that it can handle actions with up to 16 arguments. The current limit of 11 is still too few for many usecases. I would have preferred to set this to something like 32, but that would require deeper changes, as the current methods rely on System.Action, which can only handle up to 16 parameters at once.

While binding arguments to an arbitrary number of properties does work as a workaround, it increases the verbosity of users' code somewhat significantly. 11 is a relatively small number to have as an upper limit for standard functionality, and 16 is still relatively small. (For instance, the legacy code I recently modified to use System.CommandLine has 24 CLI arguments, and it is unlikely that I could eliminate 8 while maintaining functionality.)

 This change may increase utility, but I would suggest that it should be increased further if possible. Doing so is outside of my bandwidth currently.

### DIFF
--- a/src/System.CommandLine/Invocation/CommandHandler.cs
+++ b/src/System.CommandLine/Invocation/CommandHandler.cs
@@ -62,6 +62,26 @@ namespace System.CommandLine.Invocation
             Action<T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11> action) =>
             HandlerDescriptor.FromDelegate(action).GetCommandHandler();
 
+        public static ICommandHandler Create<T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12>(
+            Action<T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12> action) =>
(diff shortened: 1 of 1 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/command-line-api?pullRequest=1159) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/command-line-api?pullRequest=1159) before we can accept your contribution.<br/><br/>:x: Ssloto [sign now](https://cla.dotnetfoundation.org/dotnet/command-line-api?pullRequest=1159)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/command-line-api?pullRequest=1159) it.</sub>

jonsequitur (approved):

jonsequitur:
Thanks, @Ssloto!

Issue #1127, opened 2020-12-11
Consolidate redundant or ambiguous APIs


jonsequitur:
Remove the following APIs, as there are less ambiguous (all for the same reason outlined in #138) and more performant approaches are available (i.e. passing the `Symbol` for which you want results to overloads of these methods): (#1337)

- [ ] `ParseResultExtensions.HasOption(ParseResult, string)` *(Currently marked obsolete)*
- [ ] `ParseResult.ValueForOption(string)`  *(Currently marked obsolete)*
- [ ] `ParseResult.ValueForOption<T>(string)`  *(Currently marked obsolete)*
- [ ] `ParseResult.ValueForArgument(string)` *(Currently marked obsolete)*
- [ ] `ParseResult.ValueForArgument<T>(string)` *(Currently marked obsolete)*
- [x] `ParseResult[string]`
- [x] `CommandResult.OptionResult(string)`
- [x] `CommandResult[string]`
- [x] `ParseResult.ValueForOption<T>(Option)`
- [x] `ParseResult.ValueForArgument<T>(Argument)`

Consider renaming the following APIs to `GetValueOrDefault` for consistency with the `ArgumentResult.GetValueOrDefault` and `OptionResult.GetValueOrDefault` extension methods. (#1394)

- [ ] `ParseResult.ValueForArgument<T>(Argument<T>)` *(Currently marked obsolete)*
- [ ] `ParseResult.ValueForArgument(Argument)` *(Currently marked obsolete)*
- [ ] `ParseResult.ValueForOption<T>(Option<T>)` *(Currently marked obsolete)*
- [ ]  `ParseResult.ValueForOption(Option option)`  *(Currently marked obsolete)*
 
Consider creating new methods for recursively looking up `Symbol` and/or `SymbolResult` instances by alias, e.g. (naming needs discussion):

~- [ ] `IEnumerable<ArgumentResult> ParseResult.FindArgumentResultsByName(string name)`~
~- [ ] `IEnumerable<OptionResult> ParseResult.FindOptionResultsByAlias(string alias)`~
~- [ ] `IEnumerable<ISymbol> CommandLineConfiguration.FindSymbolsByAlias(string alias)`~

Related: #138.

PR #1161, opened 2021-01-08
More perf tuning

jonsequitur:
The major change in this PR is to redefine `ICommand.Options` and `ICommand.Arguments` to return `IReadOnlyCollection<T>` rather than `IEnumerable<T>`, and then remove some use of LINQ and replace `foreach` loops with `for` loops.

I've removed a test (`Global_options_appear_in_child_command_options_list`) and the behavior it's testing, but I think the change makes reasonable sense. In effect, `Command.Options` and `Command.GlobalOptions` are separate collections, and I think this change makes that clearer.

### DIFF
--- a/src/System.CommandLine.DragonFruit/CommandLine.cs
+++ b/src/System.CommandLine.DragonFruit/CommandLine.cs
@@ -223,12 +223,13 @@ namespace System.CommandLine.DragonFruit
                         }
                         else
                         {
-                            foreach (var argument in builder.Command.Arguments)
+                            for (var i = 0; i < builder.Command.Arguments.Count; i++)
                             {
+                                var argument = builder.Command.Arguments[i];
                                 if (string.Equals(
-                                        argument.Name,
-                                        kebabCasedParameterName, 
-                                        StringComparison.OrdinalIgnoreCase))
+                                    argument.Name,
+                                    kebabCasedParameterName,
+                                    StringComparison.OrdinalIgnoreCase))
                                 {
                                     argument.Description = parameterDescription.Value;
                                 }
--- a/src/System.CommandLine.Tests/Binding/TypeConversionTests.cs
+++ b/src/System.CommandLine.Tests/Binding/TypeConversionTests.cs
@@ -493,16 +493,18 @@ namespace System.CommandLine.Tests.Binding
         {
             var directoryInfo = new DirectoryInfo(Directory.GetCurrentDirectory());
 
+            var argument = new Argument<DirectoryInfo>("the-arg", () => directoryInfo);
+
             var command = new Command("something")
             {
-                new Argument<DirectoryInfo>("the-arg", () => directoryInfo)
+                argument
             };
 
             var result = command.Parse("something");
 
             result.Errors.Should().BeEmpty();
 
-            var value = result.CommandResult.GetArgumentValueOrDefault("the-arg");
+            var value = result.ValueForArgument("the-arg");
 
             value.Should().Be(directoryInfo);
         }
--- a/src/System.CommandLine/Collections/AliasedSet.cs
+++ b/src/System.CommandLine/Collections/AliasedSet.cs
@@ -12,9 +12,18 @@ namespace System.CommandLine.Collections
     {
         private protected readonly Dictionary<string, T> ItemsByAlias = new Dictionary<string, T>();
 
+        public int Count => Items.Count;
+
         private protected List<T> Items { get; } = new List<T>();
 
         private protected HashSet<T> DirtyItems { get; } = new HashSet<T>();
+
+        public bool Contains(string alias)
+        {
+            EnsureAliasIndexIsCurrent();
+
+            return ItemsByAlias.ContainsKey(alias);
+        }
 
         public T? GetByAlias(string alias)
         {
@@ -24,8 +33,6 @@ namespace System.CommandLine.Collections
 
             return value;
         }
-
-        public int Count => Items.Count;
 
         IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
 
@@ -53,17 +60,11 @@ namespace System.CommandLine.Collections
 
         protected abstract IReadOnlyCollection<string> GetAliases(T item);
 
-        public bool Contains(string alias)
-        {
-            EnsureAliasIndexIsCurrent();
-
-            return ItemsByAlias.ContainsKey(alias);
-        }
-
         public T this[int index] => Items[index];
 
         private protected void EnsureAliasIndexIsCurrent()
         {
+            // FIX: (EnsureAliasIndexIsCurrent) 
             if (DirtyItems.Count == 0)
             {
                 return;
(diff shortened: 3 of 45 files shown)


rvats (approved):

src/System.CommandLine.Tests/Binding/TypeConversionTests.cs
```diff
@@ -493,16 +493,18 @@ public void A_default_value_with_a_custom_constructor_can_be_specified_for_a_com
...
-                new Argument<DirectoryInfo>("the-arg", () => directoryInfo)
+                argument
             };
 
             var result = command.Parse("something");
 
             result.Errors.Should().BeEmpty();
 
-            var value = result.CommandResult.GetArgumentValueOrDefault("the-arg");
+            var value = result.ValueForArgument("the-arg");
```
paulomorgado, line 507: That's a strange method name. Method names should start with verbs_ `GetValueForArgument`

I guess it's like `Newtonsoft.Json`. A parameterized property.

src/System.CommandLine.DragonFruit/CommandLine.cs
```diff
@@ -223,12 +223,13 @@ public static class CommandLine
...
+                            for (var i = 0; i < builder.Command.Arguments.Count; i++)
                             {
+                                var argument = builder.Command.Arguments[i];
                                 if (string.Equals(
-                                        argument.Name,
-                                        kebabCasedParameterName, 
-                                        StringComparison.OrdinalIgnoreCase))
+                                    argument.Name,
+                                    kebabCasedParameterName,
+                                    StringComparison.OrdinalIgnoreCase))
```
paulomorgado, line 232: I'm just curious on how big of an improvement is this. Indexing the collection vs. boxing the enumerator.

src/System.CommandLine/Collections/SymbolSet.cs
```diff
@@ -91,5 +111,51 @@ item switch
...
+                        if (this[i] is Option Option)
+                        {
+                            options.Add(Option);
+                        }
+                    }
+
+                    return options;
+                }
+            }
+        }
```
paulomorgado, line 159: If these are read-only lists, consider storing them as `ReadOnlyCollection<T>` to avoid messing with the collection.

The empty collection could be a static reused.

src/System.CommandLine/SymbolExtensions.cs
```diff
@@ -2,14 +2,12 @@
...
 using System.Collections.Generic;
-using System.CommandLine.Parsing;
-using System.Linq;
 
 namespace System.CommandLine
 {
     public static class SymbolExtensions
     {
-        internal static IEnumerable<IArgument> Arguments(this ISymbol symbol)
+        internal static IReadOnlyList<IArgument> Arguments(this ISymbol symbol)
```
paulomorgado, line 10: `GetArguments()` ???

src/System.CommandLine/Collections/AliasedSet.cs
```diff
@@ -69,34 +69,29 @@ private protected void EnsureAliasIndexIsCurrent()
...
                     if (pair.Value.Equals(dirtyItem))
                     {
                         ItemsByAlias.Remove(pair.Key);
                     }
                 }
 
-                var wasRemoved = !Items.Contains(dirtyItem);
-
-                if (!wasRemoved)
+                if (Items.Contains(dirtyItem))
```
paulomorgado, line 84: Are repeated items allowed?

I wonder if a `SortedList<T>` would perform better.

paulomorgado: I was considering `SortedList<T>` instead of `List<T>` if `HashSet<T>` is not an option.

`List<T>.Contains(T)` is O(N).

jonsequitur: Ah, got it. I'll take a look.

src/System.CommandLine.Tests/Binding/TypeConversionTests.cs
jonsequitur: We'll likely rename these: #1127.

src/System.CommandLine.DragonFruit/CommandLine.cs
jonsequitur: I don't love the code change, but cumulatively, it seems to be helping. This is one of the benchmarks I've been checking:

image

jonsequitur: The code being benchmarked is: 

```csharp
            var subcommandOne = new Command("subcommand-one")
            {
                new Option<int>(new[] { "-i", "--int" }),
                new Argument<string[]>("arg")
            };
            subcommandOne.Handler = CommandHandler.Create<int, string[]>((@int, arg) => { });

            var subcommandTwo = new Command("subcommand-two")
            {
                new Option<int>(new[] { "-i", "--int" }),
                new Argument<string[]>("arg")
            };
            subcommandTwo.Handler = CommandHandler.Create<int, string[]>((@int, arg) => { });
            
            var root = new RootCommand
            {
                subcommandOne,
                subcommandTwo
            };

            var parser = new CommandLineBuilder(root)
                         .UseDefaults()
                         .Build();
                
            await parser.InvokeAsync(new []{"subcommand-two","-i","123","one","two","three"});
```

paulomorgado: Is the benchmark code id the repo?

jonsequitur: Not currently. The existing benchmarks project has a large number of benchmarks but didn't seem suited to a more exploratory perf tuning approach with prior baselines. I've been tinkering with a notebook for this kind of thing, and am still really learning my way around BenchmarkDotNet.

paulomorgado:
Consider changing this:

```C#
namespace System.CommandLine.Suggestions
{
    internal static class SuggestionExtensions
    {
        public static IEnumerable<string?> Containing(
            this IEnumerable<string?> candidates,
            string textToMatch) =>
            candidates.Where(c => c?.ContainsCaseInsensitive(textToMatch) == true);
    }
}
```

To this:

```C#
namespace System.CommandLine.Suggestions
{
    internal static class SuggestionExtensions
    {
        public static IEnumerable<string?> Containing(
            this IEnumerable<string?> candidates,
            string textToMatch)
        {
            foreach (var candidate in candidates)
            {
                if ((candidate is not null) && candidate.ContainsCaseInsensitive(textToMatch))
                {
                    yield return candidate;
                }
            }
        }
    }
}
```

Avoids:

* Delegate invocation
* Closure over `textToMatch`
* `bool?`

src/System.CommandLine/Collections/SymbolSet.cs
jonsequitur: There's no clear way to initialize the `ReadOnlyCollection<T>` without introducing an intermediate enumerator, so it won't save us anything.

src/System.CommandLine.DragonFruit/CommandLine.cs
jonsequitur: image

Issue #495, opened 2019-04-15
help for bool types


tmds:
```
  --overwrite <true|false>  Overwrite the specified region. (If not, scroll.)
```
If overwrite is false by default (which is common for bool arguments), the <true|false> can be omitted:
```
  --overwrite               Overwrite the specified region. (If not, scroll.)
```

tmds:
Thanks!

Issue #987, opened 2020-07-24
Help is not printed properly for Option<List<T>>. Error message when arity is wrong is misleading.


chhh:
I have an option that needs to take exactly 3 integers:
```csharp
var optFilterSize = new Option<List<int>>(
        aliases: new[] {"--filter-size"},
        getDefaultValue: () => new List<int>() {2, 2, 0},
        description: "distances from center") {
        IsRequired = false,
        Name = "FilterSize"
      };
optFilterSize.Argument.Arity = new ArgumentArity(3,3);
```
Note that I provide a default value.

It gets mapped to a property:
```csharp
public List<int> FilterSize { get; set; }
```
**There are two problems:**
1) When help is printed, defaults are not printed properly. Type is printed instead of defaults being listed, as follows:
```
--filter-size <filter-size>          distances from center [default: System.Collections.Generic.List`1[System.Int32]]
```
2) If wrong number of integers is provided as arguments by user, the error message is misleading:
```
Required argument missing for option: --filter-size
Required argument missing for option: --filter-size
```
This error message is printed twice, and then help is displayed. Instead it should say that this option requires 3 values to be specified.

chhh:
Related: https://github.com/dotnet/command-line-api/issues/984

jonsequitur:
Formatting lists can get complicated fairly quickly. I quite like the approach @mristin took here: https://github.com/dotnet/command-line-api/issues/984#issuecomment-663471185 

There's a second issue here though that we should clearly address, which is that the error message could be more helpful.

chhh:
Why formatting a list can get complicated quickly?
The only difference from formatting a single value is the delimiting character (or sequence), all the other complications are the same - formatting a single value.
There are rules for parsing arguments, the rules specify the allowed delimiters. The first one of the allowed delimiters (in the current config) can be used.

What about formatting floating point values? That is also complicated. How many decimal places to display?

It looks like one solution would be to allow setting a delegate for formatting default values `string displayValue(T val)`

mristin:
> Why formatting a list can get complicated quickly?

You need to fix the delimiter, the quote symbols as well as escaping sequences which differ across consoles (*e.g.* poweshell *versus* bash). Think regular expressions and the general solution becomes quite messy.

Issue #984, opened 2020-07-22
Default ToString used in help is unhelpful for many types


mristin:
Hi,
(Please apologize if this is a duplicate -- I couldn't find a similar issue in the archive nor could I find anything relevant in the documentation.)

When you provide a default value as an array of strings:

```cs
var rootCommand = new RootCommand(
	"...")
{
	new Option<string[]>(
		new[]{"--prefixes"},
		()=>new[]{"^TODO|BUG|HACK"},
		"Prefix regular expressions marking the TODOs"
	)
};
```

you get a cryptic message in the help:

```
Options:
  --prefixes <prefixes>                   Prefix regular expressions marking the TODOs [**default: System.String[]**]
```

Is there a way to provide the string representation of the default value? Or at least handle the default case for `string[]` (*e.g.*, join the array with commas)?

monoman:
I think you should not say that your option is typed as <string[]>, but as
<string> and set the Arity for it, which says that multiple values can be
provided by the user in the command line.

new Option<string>(
		"--prefixes",
		"^(TODO|BUG|HACK)",
		"Prefix regular expressions marking the TODOs"
	){ Argument = new Argument { Arity = ArgumentArity.OneOrMore } };


Rafael Teixeira
O..:.)oooo

mristin:
@monoman thanks for the hint; that escaped my attention while reading the documentation. 

Could you maybe write down how to define the default values in the case where you set the arity?

chhh:
@monoman That doesn't sound right. There either is no way to access values after the first one, or no way to set the defauts.
Related issue: https://github.com/dotnet/command-line-api/issues/987

mristin:
Actually, it is not obvious how to serialize an array to a string that would be informative in help messages for all the cases (*e.g.*, should you concatenate with a comma or whitespace or ...?).

Here is what I did to temporarily fix the issue in my code.

Define somewhere the default value:

```csharp
        private static readonly string[] DefaultPrefixes =
        {
            "^TODO", "^BUG", "^HACK"
        };
```

Do not provide the default value in the option and stringify the array myself:
```csharp
                new Option<string[]?>(
                    new[]{"--prefixes"},
                    "Prefix regular expressions marking the TODOs. " +
                    $"[Default: {string.Join(" ", DefaultPrefixes)}]"
                ),
```

Use null coalescing operator to select the default value in the code:
```csharp
var resolvedPrefixes = prefixes ?? DefaultPrefixes;
```

[...]

PR #1168, opened 2021-01-19
Suggest output for boolean arguments with --help

### LINKED ISSUE #1157: Bool arguments missing name in help page
Consider this application:

```c#
static int Main(string[] args)
{
    {
        new Argument<bool>("value1", "Some value"),
        new Argument<int>("value2", "Another value"),
    };

    rootCommand.Handler = CommandHandler.Create<bool>(value1 =>
    {
        Console.WriteLine($"The value for 'value1' is: {value1}");
    });

}
```

When calling this application with the `--help` parameter, we get an output like this:

```
Usage:
  CommandLineReproducer [options] [<value1> <value2>]

Arguments:
              Some value
  <value2>    Another value

Options:
  --version         Show version information
  -?, -h, --help    Show help and usage information
```

Notice how in the **Arguments** sections we see `<value2>` but not `<value1>`. It seems it's not printed here because the type parameter is `bool`.

adamhewitt627:
Resolves #1157. While it is extra noise to show `<False|True>` on boolean options, it's good to see it in the help text when a command has a boolean argument.

### DIFF
--- a/src/System.CommandLine.Tests/Help/HelpBuilderTests.cs
+++ b/src/System.CommandLine.Tests/Help/HelpBuilderTests.cs
@@ -809,7 +809,7 @@ Arguments:
         [Theory]
         [InlineData(typeof(bool))]
         [InlineData(typeof(bool?))]
-        public void Command_argument_descriptor_is_empty_for_boolean_values(Type type)
+        public void Command_argument_descriptor_includes_suggestions_for_boolean_values(Type type)
         {
             var description = "This is the argument description";
 
@@ -828,7 +828,7 @@ Arguments:
 
             var expected =
                 $"Arguments:{NewLine}" +
-                $"{_indentation}{_columnPadding}{description}";
+                $"{_indentation}<False|True>{_columnPadding}{description}";
 
             _console.Out.ToString().Should().Contain(expected);
         }
--- a/src/System.CommandLine.Tests/Help/HelpBuilderTests.cs
+++ b/src/System.CommandLine.Tests/Help/HelpBuilderTests.cs
@@ -828,7 +828,7 @@ Arguments:
 
             var expected =
                 $"Arguments:{NewLine}" +
-                $"{_indentation}<False|True>{_columnPadding}{description}";
+                $"{_indentation}<{bool.FalseString}|{bool.TrueString}>{_columnPadding}{description}";
 
             _console.Out.ToString().Should().Contain(expected);
         }
--- a/src/System.CommandLine/Help/HelpBuilder.cs
+++ b/src/System.CommandLine/Help/HelpBuilder.cs
@@ -526,8 +526,8 @@ namespace System.CommandLine.Help
 
         protected virtual string ArgumentDescriptor(IArgument argument)
         {
-            if (argument.ValueType == typeof(bool) ||
-                argument.ValueType == typeof(bool?))
+            if (argument.Parents.OfType<IOption>().Any()
+                && (argument.ValueType == typeof(bool) || argument.ValueType == typeof(bool?)))
             {
                 return "";
             }


src/System.CommandLine.Tests/Help/HelpBuilderTests.cs
```diff
@@ -828,7 +828,7 @@ public void Command_argument_descriptor_is_empty_for_boolean_values(Type type)
 
             var expected =
                 $"Arguments:{NewLine}" +
-                $"{_indentation}{_columnPadding}{description}";
+                $"{_indentation}<False|True>{_columnPadding}{description}";
```
Keboo, line 831:
```suggestion
                $"{_indentation}<{bool.FalseString}|{bool.TrueString}>{_columnPadding}{description}";
```

src/System.CommandLine/Help/HelpBuilder.cs
```diff
@@ -526,8 +526,8 @@ private IEnumerable<HelpItem> GetArgumentHelpItems(ISymbol symbol)
 
         protected virtual string ArgumentDescriptor(IArgument argument)
         {
-            if (argument.ValueType == typeof(bool) ||
-                argument.ValueType == typeof(bool?))
+            if (argument.Parents.OfType<IOption>().Any()
```
jonsequitur, line 529: Is the parent check needed here?

adamhewitt627: That's the part I'm least clear on. From my looking through history on this line, tracking back to #495 and #500, it appears that the intention is to not show the suggestions on boolean _options_, but that means that an `Argument` on a command only outputs the description instead of suggesting an input.

I'm not 100% sure this was the correct check, but I also didn't see a better one. From:
```cs
 {
    new Argument<bool>("value1", "Some value"),
    new Argument<int>("value2", "Another value"),
};
```

### Previous
```
Usage:
  CommandLineReproducer [options] [<value1> <value2>]

Arguments:
              Some value
  <value2>    Another value

Options:
  --version         Show version information
  -?, -h, --help    Show help and usage information
```

### Proposed
```
Usage:
  CommandLineReproducer [options] [<value1> <value2>]

Arguments:
  <False|True>   Some value
  <value2>       Another value

Options:
  --version         Show version information
  -?, -h, --help    Show help and usage information
```

jonsequitur: Looking at this now, I'm not sure special-casing `bool`/`bool?` is correct at all.

So what should we show?

The proposal here isn't consistent with how we show allowed values elsewhere, though how we do that is also not ideal. 

For comparison, here's comparable output from System.CommandLine (`t-rex`) and `git`, with allowed values circled:

image

I wonder if we should do something more like the `t-rex` example for `bool` as well, e.g.:

```console
Usage:
  CommandLineReproducer [options] [<value1> <value2>]

Arguments:
  value1 <False|True>   Some value
  value2                         Another value

Options:
  --version         Show version information
  -?, -h, --help    Show help and usage information
```

adamhewitt627: I agree that is better, although to some extent it's reverting #495. I'd picture this landing as:

1. Suppress the `<true|false>` on an `Option` when the default is false. (This is #495)
2. Show `<true|false>` on Options when the default is true. (Not sure it does this today)
3. Always show `<true|false>` on arguments that are separate from an option. (Resolving #1157)

jonsequitur: At the time we fixed #495, the display of default values hadn't been implemented. I do think it's awkward that we have the allowable values in the left column and the default value in the right column. 

If we combined the allowed and default values into a single piece of the help, could it be made clearer overall?

adamhewitt627: Maybe, but I'm not entirely sure what you mean that to end up being.

jonsequitur: I'm not sure. Open to suggestions.

skrysmanski: Ah - I'm not sure what the state of this PR is but having the documentation like this is wrong:

```
Arguments:
  <False|True>   Some value
  <value2>       Another value
```

It should be:

```
Arguments:
  <value1>       Some value
  <value2>       Another value
```

adamhewitt627: @skrysmanski The state at the moment is to take the time to evaluate help output as a whole in light of your issue and a couple others, #984 and #987. I agree that the current implementation is wrong, and I'm open to suggestions. What do you think about this:
```
Arguments:
  <value1>                 Some value
    allowed: true|false
  <value2>                 Another value
  <value3>                 An enum value
    allowed: Read|ReadWrite|Write
```

skrysmanski: @adamhewitt627 I have no personal preference on either option. My only point was that the name of the argument was missing. I know how complicated it can be to have proper help pages. This is why I'm using `System.CommandLine` so that I don't need to think about this ;)

jonsequitur: Since this discussion seems to be larger than the current issue, I opened #1174 to discuss.

PR #1170, opened 2021-01-22
Making HelpOption localizable

### LINKED ISSUE #1067: Allow help option description to be localized
The help option description is being taken directly from a string instead of the ValidationMessages class, which means it can't be localized, see here: https://github.com/dotnet/command-line-api/blob/main/src/System.CommandLine/Help/HelpOption.cs#L15

cc @jonsequitur

Keboo:
Renamed ValidationMessages to Resources

Fixes #1067

### DIFF
--- a/src/System.CommandLine/Resources.cs
+++ b/src/System.CommandLine/Resources.cs
@@ -8,11 +8,11 @@ using System.Linq;
 
 namespace System.CommandLine
 {
-    public class ValidationMessages
+    public class Resources
     {
-        public static ValidationMessages Instance { get; } = new ValidationMessages();
+        public static Resources Instance { get; } = new Resources();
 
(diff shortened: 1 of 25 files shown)


src/System.CommandLine/Resources.cs
```diff
@@ -8,11 +8,11 @@
 
 namespace System.CommandLine
 {
-    public class ValidationMessages
+    public class Resources
```
jonsequitur, line 11: I think the naming might be pointed in the wrong direction here. It's reflecting the implementation detail (that the underlying strings are resources) and not the use case (localizable messages). This is especially noticeable because it exposes parameterized methods, not strings.

jonsequitur (approved):

PR #1171, opened 2021-01-24
Default empty collections when argument not specified

### LINKED ISSUE #1140: Change default value for enumerable types to empty rather than null
Currently, `Argument<T>` and `Option<T>` when `T` is `IEnumerable` or `string` will bind to `null` when no corresponding input is provided on the command line. These should instead default to return `Enumerable.Empty`, `Array.Empty`, and empty `List<T>` or `string.Empty` as appropriate.

Keboo:
Fixes #1140 
Sets up default "empty" values for a variety of known types.
Types handles are:
- array => `Array.Empty`
- string => `String.Empty`
- IEnumerable<T>/IEnumerable => `Enumerable.Emtpy`
- ICollection<T>/ICollection => `Array.Empty`
- IList => `Array.Empty`
- IList<T> => `new List<T>()`;

### DIFF
--- a/src/System.CommandLine/Binding/ArgumentConverter.cs
+++ b/src/System.CommandLine/Binding/ArgumentConverter.cs
@@ -195,7 +195,11 @@ namespace System.CommandLine.Binding
                       .GetInterfaces()
                       .FirstOrDefault(IsEnumerable);
 
-            return enumerableInterface?.GenericTypeArguments.FirstOrDefault();
+            return enumerableInterface?.GenericTypeArguments switch
+            {
+                { Length: 1 } genericTypeArguments => genericTypeArguments[0],
+                _ => null
+            };
         }
 
         internal static bool IsEnumerable(this Type type)
--- a/src/System.CommandLine/Binding/ArgumentConverter.cs
+++ b/src/System.CommandLine/Binding/ArgumentConverter.cs
@@ -330,17 +330,32 @@ namespace System.CommandLine.Binding
         [return: MaybeNull]
         internal static T GetValueOrDefault<T>(this ArgumentConversionResult result)
         {
-            switch (result)
+            return result switch
             {
-                case SuccessfulArgumentConversionResult successful:
-                    return (T)successful.Value!;
-                case FailedArgumentConversionResult failed:
-                    throw new InvalidOperationException(failed.ErrorMessage);
-                case NoArgumentConversionResult _:
-                    return default!;
-                default:
-                    return default!;
+                SuccessfulArgumentConversionResult successful => (T)successful.Value!,
+                FailedArgumentConversionResult failed => throw new InvalidOperationException(failed.ErrorMessage),
+                NoArgumentConversionResult _ => default!,
+                _ => default!,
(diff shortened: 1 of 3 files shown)


src/System.CommandLine/Binding/ArgumentConverter.cs
```diff
@@ -195,7 +195,7 @@ static IList CreateArray(Type itemType, int capacity)
                       .GetInterfaces()
                       .FirstOrDefault(IsEnumerable);
 
-            return enumerableInterface?.GenericTypeArguments[0];
+            return enumerableInterface?.GenericTypeArguments.FirstOrDefault();
```
jonsequitur, line 198: Though I like the change, I'd revert this. This is one of these little perf optimizations that do add up.

src/System.CommandLine/Binding/ArgumentConverter.cs
```diff
@@ -330,17 +330,68 @@ public static bool CanBeBoundFromScalarValue(this Type type)
...
+            };
+            
+            static object GetEmptyList(Type itemType)
+            {
+                return Activator.CreateInstance(typeof(List<>).MakeGenericType(itemType));
+            }
+
+            static IEnumerable GetEmptyEnumerable(Type itemType)
+            {
+                var genericMethod = EnumerableEmptyMethod.MakeGenericMethod(itemType);
```
jonsequitur, line 389: I wonder if these are worth caching.

src/System.CommandLine/Binding/ArgumentConverter.cs
Keboo: I assume you mean replace `.FirstOrDefault()` with a bounds check? Because the `[0]` throws an out of range exception when working with non-generic. `IsEnumerable` returns true for non generic `IEnumerable` so there may not be any generic arguments.

src/System.CommandLine/Binding/ArgumentConverter.cs
Keboo: I could certainly put that in. Figured it was better to wait for perf benchmarks. But my suspicion is that caching may not offer much.

src/System.CommandLine/Binding/ArgumentConverter.cs
jonsequitur: Did you hit that exception? At least before, that wasn't happening.

Agreed, this change is safer, but we should attempt to keep the optimization.

src/System.CommandLine/Binding/ArgumentConverter.cs
jonsequitur: Probably true.

src/System.CommandLine/Binding/ArgumentConverter.cs
Keboo: Yea the exception occurs on the tests I added for non-generic IEnumerable, ICollection, and IList

jonsequitur (approved):

jonsequitur (approved):

Issue #1175, opened 2021-01-26
Exact match on longer identifier symbols should take precedence over POSIX unbundling


jonsequitur:
Currently, POSIX unbundling will take precedence over an exact match on a longer symbol name, and this behavior should be the other way around.

For example, given this setup:

```csharp
var cmd = new RootCommand
{
    new Option<string>("-c"),
    new Option<string>("-clp"),
};
```

Currently the input `-clp` is parsed to be the equivalent of `-c lp`, and `lp` will be considered to be an argument to option `-c`. 

This should change so that it considers this to match option `-clp`, since it's an exact match. 

This should address #1169 as well, e.g. matching additional symbols, and not just the argument case in the example above.

PR #1177, opened 2021-01-28
Ensure longer option is parsed when aliases are prefixed

sfoslund:
Fixes https://github.com/dotnet/command-line-api/issues/1175

### DIFF
--- a/src/System.CommandLine/Parsing/StringExtensions.cs
+++ b/src/System.CommandLine/Parsing/StringExtensions.cs
@@ -205,9 +205,7 @@ namespace System.CommandLine.Parsing
                 }
 
                 // don't unbundle if this is a known option
-                if (knownTokens.Values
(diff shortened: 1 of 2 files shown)


src/System.CommandLine/Parsing/StringExtensions.cs
```diff
@@ -204,6 +204,14 @@ bool CanBeUnbundled(string arg, out IReadOnlyList<string>? replacement)
                     return false;
                 }
 
+                // don't unbundle if this is a known option
+                if (knownTokens.Values
```
jonsequitur, line 208: Does `knownTokens.ContainsKey(arg)` also work? It will be more performant.

sfoslund: Sure, just made that change.

jonsequitur (approved):

PR #1179, opened 2021-01-29
Ensure longer option is parsed with argument delimiters

sfoslund:
Follow up from https://github.com/dotnet/command-line-api/pull/1177, which did not apply the exact matching behavior to options specifying arguments after delimiters.

### DIFF
--- a/src/System.CommandLine.Tests/OptionTests.cs
+++ b/src/System.CommandLine.Tests/OptionTests.cs
@@ -371,15 +371,14 @@ namespace System.CommandLine.Tests
             option.Aliases.Should().Contain("name");
         }
 
-        [Fact]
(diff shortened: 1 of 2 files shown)


src/System.CommandLine.Tests/OptionTests.cs
```diff
@@ -378,6 +378,9 @@ public void When_aliases_overlap_the_longer_alias_is_chosen()
 
             var parseResult = option.Parse("-option value");
             parseResult.ValueForOption(option).Should().Be("value");
+
+            parseResult = option.Parse("-option:value");
```
jonsequitur, line 382: Please split this into two test cases.

jonsequitur (approved):

PR #1181, opened 2021-02-01
XML documentation (Argument, Option, Command)

jonsequitur:
This updates the XML doc comments for `Argument`, `Option`, `Command`, and a few related types.


src/System.CommandLine/Argument.cs
```diff
@@ -10,7 +10,7 @@
 namespace System.CommandLine
 {
     /// <summary>
-    /// Represents a value passed to an <see cref="Option"/> or <see cref="Command"/>.
+    /// A symbol defining a value that can be passed to a <see cref="ICommand">command</see> or <see cref="IOption">option</see>.
```
Keboo, line 13: nit: Is "passed" really the right word when used in relation to the the `ICommand` and `IOption`? I think passed is a little ambiguous since it is not clear if that is passed in code or passed on the command line. Perhaps something more like: "A symbol defining a value that can be passed on the command line to a command or option."

src/System.CommandLine/Argument.cs
```diff
@@ -108,7 +108,7 @@ public List<ISuggestionSource> Suggestions
         }
 
         /// <summary>
-        /// Gets or sets the <see cref="System.Type" /> that the argument will be converted to.
+        /// Gets or sets the <see cref="Type" /> that the argument will be converted to.
```
Keboo, line 111:
```suggestion
        /// Gets or sets the <see cref="Type" /> that the argument value will be converted to.
```

src/System.CommandLine/IArgument.cs
```diff
@@ -6,7 +6,7 @@
 namespace System.CommandLine
 {
     /// <summary>
-    /// Defines a symbol with an arity.
+    /// A symbol defining a value that can be passed to a <see cref="ICommand">command</see> or <see cref="IOption">option</see>.
```
Keboo, line 9: Same suggestion as on Argument

src/System.CommandLine/IIdentifierSymbol.cs
```diff
@@ -6,20 +6,20 @@
 namespace System.CommandLine
 {
     /// <summary>
-    /// Defines a symbol with aliases.
+    /// Defines a symbol which is an identifier in the command line, such as a <see cref="ICommand">command</see> or <see cref="IOption">option</see>.
```
Keboo, line 9:
```suggestion
    /// Defines a symbol which is an identifier on the command line, such as a <see cref="ICommand">command</see> or <see cref="IOption">option</see>.
```

Keboo (approved):

src/System.CommandLine/Argument.cs
jonsequitur: Added "on the command line". 

Do you think that the `see` tags make this more ambiguous?

src/System.CommandLine/Argument.cs
jonsequitur: Instead of "value" how about "token(s)"?

I feel like "value" is also ambiguous, per your comment above. Does it refer to a value on the command line or a typed, converted value after parsing is done?

src/System.CommandLine/Argument.cs
Keboo: No i think those are good

src/System.CommandLine/Argument.cs
Keboo: Yea token would be a better word. I keep looking at the terminology docs you wrote up and comparing with those to make sure things are consistent.

Keboo (approved):

PR #1196, opened 2021-02-08
Allow abstract command handler invocation

ycrumeyrolle:
Use `ReflectedType` instead of `DeclaringType` for finding the correct invocation method. 
This allow to support abstract command handler without overriding the `InvokeAsync` method.


ycrumeyrolle:
Git misuse... Do you prefer a clean branch?

PR #1202, opened 2021-02-13
Expose ITerminal.Width property

twsouthwick:


adamsitnik:
Hello!

We have decided to deprecate the System.CommandLine.Rendering package: https://github.com/dotnet/command-line-api/issues/2576#Deprecating-experimental-projects

We recommend using https://www.nuget.org/packages/Spectre.Console which became the golden standard in the .NET Ecosystem

PR #1215, opened 2021-02-26
Honor user defined order of commands, subcommands and options

gabrielmaldi:
Currently, `HelpBuilder` sorts help items, which results in the following behavior:

*Example 1*

```cs
new CommandEx(new[] { "migrations", "mig" })
```

```
Commands:
  mig, migrations    Commands to manage migrations.
```

*Example 2*

```cs
new CommandEx(new[] { "health-checks", "hc" })
new CommandEx(new[] { "mini-profiler", "mp" })
```

```
mini-profiler, mp                Generates the MiniProfiler script
hc, health-checks <from> <to>    Generates the HealthChecks script [from: '0' (the initial database), to: the last migration]
```

---

User defined order should be respected; otherwise, the aliases (`hc`) can be shown before the main command (`health-checks`). Also, multiple commands can result in odd ordering: `mini-profiler, mp` and `hc, health-checks` (main/alias, alias/main; see Example 2).

---

For completeness:
```cs
public CommandEx(string[] aliases, string description)
    : base(aliases.First(), description)
{
    foreach (var alias in aliases.Skip(1))
    {
        AddAlias(alias);
    }
}
```

### DIFF
--- a/src/System.CommandLine.Tests/Help/HelpBuilderTests.cs
+++ b/src/System.CommandLine.Tests/Help/HelpBuilderTests.cs
@@ -1275,14 +1275,14 @@ Arguments:
         {
             var command = new RootCommand
             {
-                new Option(new[] { "-z", "-a", "--zzz", "--aaa" })
+                new Option(new[] { "--aaa", "--zzz", "-z", "-a" })
             };
 
             _helpBuilder.Write(command);
 
             _console
                 .Out
-                .ToString().Should().Contain("-a, -z, --aaa, --zzz");
+                .ToString().Should().Contain("-z, -a, --aaa, --zzz");
         }
 
         [Fact]
@@ -1324,7 +1324,7 @@ Arguments:
 
             output
                 .Should()
-                .Be("-a, -z, --aaa, --zzz    from a to z");
+                .Be("-z, -a, --zzz, --aaa    from a to z");
         }
 
         #endregion Options
--- a/src/System.CommandLine/Help/HelpBuilder.cs
+++ b/src/System.CommandLine/Help/HelpBuilder.cs
@@ -555,7 +555,6 @@ namespace System.CommandLine.Help
                              .Aliases
                              .Select(r => r.SplitPrefix())
                              .OrderBy(r => r.prefix, StringComparer.OrdinalIgnoreCase)
-                             .ThenBy(r => r.alias, StringComparer.OrdinalIgnoreCase)
                              .GroupBy(t => t.alias)
                              .Select(t => t.First())
                              .Select(t => $"{t.prefix}{t.alias}");
--- a/src/System.CommandLine/Help/HelpOption.cs
+++ b/src/System.CommandLine/Help/HelpOption.cs
@@ -7,11 +7,11 @@ namespace System.CommandLine.Help
     {
         public HelpOption() : base(new[]
         {
+            "-?",
+            "/?",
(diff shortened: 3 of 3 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/command-line-api?pullRequest=1215) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/command-line-api?pullRequest=1215) before we can accept your contribution.<br/><br/>:x: gabrielmaldi [sign now](https://cla.dotnetfoundation.org/dotnet/command-line-api?pullRequest=1215)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/command-line-api?pullRequest=1215) it.</sub>

jonsequitur:
The convention in most command line help (at least for options) is for the short-form aliases to precede the long-form aliases, so the current ordering is deliberate.

There's been discussion  though about making help easier to customize, in a way that would accommodate the desire to reorder the aliases.

@Keboo @adamhewitt627 ...thoughts?

elgonzo:
> The convention in most command line help (at least for options) is for the short-form aliases to precede the long-form aliases, so the current ordering is deliberate.

What you describe is the ordering of option aliases, but this convention does not seem to apply to command aliases (and the issue here is about command aliases). Note that in the generated help text "mp" is not preceding "mini-profiler" in OP's example, whereas "hc" is preceding "health-checks".

Keboo:
I agree, I think this is another case that needs to be handled with making configuring the help output

gabrielmaldi:
This is how I'm handling it right now:

```cs
new CommandLineBuilder(rootCommand)
    .UseHelpBuilder(context => new HelpBuilderEx(context.Console))
    .UseDefaults()
    .Build()
    .Invoke(args);
```
```cs
public class HelpBuilderEx : HelpBuilder
{
    public HelpBuilderEx(
        IConsole console,
        int? columnGutter = null,
        int? indentationSize = null,
        int? maxWidth = null)
        : base(console, columnGutter, indentationSize, maxWidth)
    {
    }

    protected override string DefaultValueHint(IArgument argument, bool isSingleArgument)
    {
        if (argument is IArgumentEx argumentEx && argumentEx.DefaultValueHintFactory != null)
        {
            var surrogate = new Argument(argument.Name);
            surrogate.SetDefaultValueFactory(argumentEx.DefaultValueHintFactory);
            argument = surrogate;
        }

        return base.DefaultValueHint(argument, isSingleArgument);
    }
}
```
Which of course is less than ideal. I agree that it would be better to make `HelpBuilder` more configurable instead of just removing the call to `ThenBy`. In that regard, I would change this method to `protected`:
```cs
private IEnumerable<HelpItem> GetOptionHelpItems(IIdentifierSymbol symbol)
{
    var rawAliases = symbol
                     .Aliases
                     .Select(r => r.SplitPrefix())
                     .OrderBy(r => r.prefix, StringComparer.OrdinalIgnoreCase)
                     .GroupBy(t => t.alias)
                     .Select(t => t.First())
                     .Select(t => $"{t.prefix}{t.alias}");
```
And extract this bit (`rawAliases`) into a new protected method so it can be overriden. I would be happy to subit a new PR and close this one.

Issue #614, opened 2019-08-12
TableView/GridView should move the cursor down using a line break to better support console resizing


christianrondeau:
When you run the TableView samples:

```
dotnet run --project .\samples\RenderingPlayground\RenderingPlayground.csproj -- [enable-vt] [output:ansi] --sample TableView
```

Try resizing your terminal window horizontally (reproduced using the default Windows console in PowerShell as well as in ConEmu) and you'll see the table layout breaking down.

I can pretty much solve the problem by sending a `\n` character to the console:

https://github.com/dotnet/command-line-api/blob/732c0e685349a0e917e19fbe62bba4e26e430a3a/src/System.CommandLine.Rendering/Views/GridView.cs#L114

(I added a `renderer.LineBreak()` method here that did `_terminal.Out.Write("\n");` but I didn't spend much time testing if it had side effects)

BobSilent:
The problem with the "missing linebreak" also occurs on rendering a GridView in `Region.Scrolling` Mode. when the whole grid is smaller than `Console.WindowWidth`.
In this case the Cursor remains at the row end, when writting the next row, it is simply appended, as `Region.Scrolling.Left`  returns `Console.CursorLeft`.
PR #1216  would resolve the issue as well, currently I am using a workaround (for my case) by simply setting last column to `ColumnDefinition.Star(1)` which in fact does not help in case of resizing, but PR #1214 also not.

adamsitnik:
Hello!

We have decided to deprecate the System.CommandLine.Rendering package: https://github.com/dotnet/command-line-api/issues/2576#Deprecating-experimental-projects

We recommend using https://www.nuget.org/packages/Spectre.Console which became the golden standard in the .NET Ecosystem

PR #1216, opened 2021-02-27
Fix GridView when called with Region.Scrolling

albyrock87:
Hi,

when rendering `GridView` with `Region.Scrolling` the grid isn't displayed correctly because `region.Left` returns the current console `Left` coordinate.

We need to store the initial left value of the region passed in.

Thanks


BobSilent:
I do not think that this is the real fix for the issue.

The grid does not render correctly if the total grid width is less than the possible console window width, because then the cursor remains at the grid end and (as no line break is written as grid end) the next row will be immediately printed to where the cursor stops.

So I think the issue describe here is more or less related to #614

PR #1222, opened 2021-03-11
Fixing confusing variable names in powershell suggest shim

powercode:
The shim uses incorrect variable names that may confuse readers.

`wordToComplete` is called `commandName`, and `commandAst` is called `wordToComplete`.


jonsequitur:
Thanks for spotting this! Please also update the `DOTNET_SUGGEST_SCRIPT_VERSION` variable in the script to version `1.0.1`.

jonsequitur (approved):

Issue #1190, opened 2021-02-05
Missed options in help section


Artur2:
Missed options in help section when calling with only `-h` argument

Way to reproduce: 
```C#
private static async Task Main(string[] args)
        {
            var rootCommand = new RootCommand("Migration tool for SQL Database")
            {
                new Option<string>(new[] { "--source-connection-string",  "-scs" }, "Full connection string to source of database.") { IsRequired = true },
                new Option<string>(new[] { "--destination-connection-string", "-dcs" }, "Full connection string to source sql server, may contain master database name.") { IsRequired = true },
                new Option<string>(new[] { "--created-database-connection-string", "-cdcs" }, "Full connection string to destination database with specified name.") { IsRequired = true },
                new Option<string>(new[] { "--database-name", "-dn" }, "Name of destination database") { IsRequired = true },
            };


            var helpBuilder = (HelpBuilder)default;
            rootCommand.TreatUnmatchedTokensAsErrors = true;
            rootCommand.Handler = CommandHandler.Create<string, string, string, string>(Migrate);

            var commandLineBuilder = new CommandLineBuilder(rootCommand)
                .UseTypoCorrections()
                .UseParseErrorReporting()
                .CancelOnProcessTermination()
                .UseExceptionHandler()
                .UseHelpBuilder(c =>
                {
                    helpBuilder = new CustomHelpBuilder(c.Console);
                    return helpBuilder;
                })
                .Build();

            await commandLineBuilder.InvokeAsync(args);
        }
```

And then call executable with -h argument
Result of call 

![image](https://user-images.githubusercontent.com/5252769/106996020-ec7de000-67a1-11eb-85c6-97bc54857e03.png)

When calling without any arguments

![image](https://user-images.githubusercontent.com/5252769/106996360-a70de280-67a2-11eb-8c8b-52b2c07010c7.png)

So how to display all options, when calling with `-h` or `--help` or `-?`

jonsequitur:
It looks like the issue is in your `CustomHelpBuilder`.

Aethon:
OK, here is an even simpler example that does not involve the `CustomHelpBuilder`:

```csharp
        public static async Task Main(string[] args)
        {
            var rootCommand = new RootCommand
            {
                new Argument<string>(
                    name: "inputDir",
                    description: "directory containing the domain definition (YAML) files"
                ),
                new Option<string?>(
                    "--modelbook-dir",
                    () => null,
[...]

Issue #1184, opened 2021-02-02
Help text aligning not working for custom IConsole


skrysmanski:
When providing one's own `IConsole` implementation, help text aligning no longer works.

Consider this code snippet:

```c#
var rootCommand = new RootCommand
{
    new Option<bool>("--force", "Suppresses prompting to confirm you want to overwrite an existing destination file. With some very long text that make it necessary to break this line even though it does not provide any additional information."),
    new Option<int>("--value", "Another value"),
};

rootCommand.Invoke(new [] { "--help" });
```

Running this code results in something like this:

![regular](https://user-images.githubusercontent.com/914747/106623555-07b7d800-6575-11eb-99a3-8fcbc33ae7bf.png)

However, if an `IConsole` object is passed to `Invoke`:

```c#
rootCommand.Invoke(new [] { "--help" }, myConsole);
```

the output changes to this:

![broken](https://user-images.githubusercontent.com/914747/106624229-b6f4af00-6575-11eb-96b6-2669927d1165.png)

Note how the wrapping and aligning of the help text of `--force` no longer works.

As far as I can tell, the problem comes from this code:

https://github.com/dotnet/command-line-api/blob/98a466e60db38e00bd747e3a27c21ae54b1154ec/src/System.CommandLine/Help/HelpBuilder.cs#L831-L841

The problem is that `GetConsoleWindowWidth()` is not part of `IConsole` - thus users with custom `IConsole` implementations/wrappers have no way of getting proper help text aligning/wrapping.

The most straight way forward to solve this would be to simply add `GetConsoleWindowWidth()` to `IConsole` - but this would be a breaking change.

skrysmanski:
I've been thinking a little bit about this and I think the new member should be `int?` (not just `int`) - so implementers can return `null` for consoles that don't have a width (like logs or similar):

```c#
int? ConsoleWidth { get; }
```

I also think the member should be a property rather than a method.

jonsequitur:
The `IConsole` interface deliberately doesn't have properties such as width and cursor position, in order to more accurately represent the core console (as opposed to terminal) concept. By the definition we're using (and there's admittedly a lot of confusion around this):

_Console_: stdin, stdout, stderr
_Terminal_: A stateful app for communicating over a console interface.

[...]

Issue #1125, opened 2020-12-11
Localizability 


KathleenDollard:
Complete the basic story for allowing messages, etc to be localizable. This will also make it easier to customize the messages.

- [ ] Make sure all user-visible strings are localizable
- [ ] Rename `ValidationMessages` to be able to include help and other user strings.

Issue #994, opened 2020-07-26
Can I control what suggestions are show for my arguments in --help


michaelgwelch:
The number of suggestions available to a user of my app is quite large unless they've typed some letters. Showing all possible suggestions during a call to help would be overwhelming. (During an actual auto-complete at least bash will ask you "Are you sure you want to show all 200 lines?" or whatever it turns out to be.

I know I could hook into Environment.GetCommandLineArgs to see if --help or -h were part of the command line and then adjust my output.

Is there something built-in to an argument or option to control this behavior?

UPDATE: Based on some of the discussion it's apparent that I didn't state that in my case I'm talking about arguments not options. As noted in a comment by me below it would be like if you typed `git help branch` and the documentation that came back listed all of your branches.

jonsequitur:
Let me see if I understand correctly. Are you saying the available values list (example circled in the screenshot) is too large to be helpful?

![image](https://user-images.githubusercontent.com/547415/88492710-982bb500-cf61-11ea-8b9e-749f6eff2f8a.png)

If so, this isn't controllable currently, although some sort of limit on how many we print might be useful. Another approach would be a custom `HelpBuilder`, though I'm not sure we currently expose functionality to make that very easy.

What would be your ideal output?

michaelgwelch:
> Are you saying the available values list (example circled in the screenshot) is too large to be helpful?

That is correct. In my case one of my arguments has 2200 possible values. 

What I'm doing now is in the delegate for GetSuggestions for this argument, I'm using Environment.GetCommandLineArgs() to do a simple scan to see if -h, --help, help, -? etc. are in the list and if so just returning an empty array.

> What would be your ideal output?

So what would be acceptable (and what happens with my solution) is no arguments are shown. But in the description for the argument I list some examples. The user of the app will no what they are looking for. In this case suggestions are used to auto-complete, it's not so important that they be there to list all the possibilities.

Another idea would be to show a very small subset of possibilities rather than put the examples in the Description.

So my output might look like


```text
Arguments:
  <set> <opt1|opt2|opt3|...>             The name or id of the enumeration set you are looking for.
```

KathleenDollard:
I'm running into this modeling the CLI as well, in that it pushes the description farther to the right than I think is pretty. I could see these options: 

* Check the length of the option name and option list. If it is more than <n> characters, use ellipses
[...]

PR #1229, opened 2021-03-18
Improve help API

Keboo:
This PR aims to address a number of issues around help output.

#984 - The exact case present in the issue is the same as #987 so the default `.ToString()` implementation will not be used. More importantly it is much simpler to provide default values in the cases where the default `.ToString()` should not be used.
#994 - The descriptor for a an option/argument's suggestions can be overwritten to any arbitrary value allowing for customization of suggestions.
#987 - Enumerable default values now show pipe delimited by default.
#1125 - Strings used in help have now been localized.
#1184 - HelpBuilder no longer relies on casting to internal types to get the width of the console. Instead it takes in the width as a parameter that can be controlled.
#1190 - Options section will always be show if there is at least one non-hidden option.

There are a few keys changes in here.
- The API of the `HelpBuilder` has been greatly simplified. This should help (pun intended) clear up much of the confusion for people wanting to provide their own `IHelpBuilder` implementation. 
- The default `HelpBuilder` has been reworked with many methods being made `protected`/`virtual` to allow for easier extending of the help build rather than needing to start from stratch.
- The HelpBuilder has several new `.Customize(...)` extensions that can be used to adjust how individual items display in help. These are easily accessed via the `CommandLineBuilderExtensions.UseHelp` generic overloads that are added here.

As a proof-of-concept I did a [small example](https://github.com/Keboo/command-line-api/tree/fix1174Spectre/samples/Spectre.Help/Spectre.Help) using [Spectre.Console](https://spectresystems.github.io/spectre.console/) to display help output.


src/System.CommandLine.Tests/Help/Approvals/HelpBuilderTests.Help_describes_default_values_for_complex_root_command_scenario.approved.txt
```diff
@@ -1,24 +1,24 @@
-﻿the-root-command:
+﻿the-root-command
   Test description
 
 Usage:
-  the-root-command [options] <the-root-arg-no-description-no-default> [<the-root-arg-no-description-default> <the-root-arg-no-default> [<the-root-arg> [<the-root-arg-enum-default>]]]
+  the-root-command <the-root-arg-no-description-no-default> [<the-root-arg-no-description-default> <the-root-arg-no-default> [<the-root-arg> [<the-root-arg-enum-default>]]] [options]
```
jonsequitur, line 5: In most CLI synopses, options precede arguments. In many cases the options are actually fully expanded. Something to consider?

Keboo: That is a good point, I will fix it.

jonsequitur (approved):

PR #1243, opened 2021-04-01
RFC: Add runAsDaemon option to UseHost

joemphilips:
In my own project I wanted to start Asp.Net server from command line.
I also wanted to bind parsed CLI arguments into my option by `BindCommandLine()`,  So I had to use `UseHost()`, which does not support running `IHost` forever and instead stops right after the startup.

I ended up re-writing my own commandline middleware which is mostly the same with `UseHost` ,but with one difference that it invokes `host.RunAsync()` instead of `host.StartAsync()`.
So I thought it would be better to have this feature built in to this library.

Please don't merge yet, this PR has no test. I'm not even sure this is the right way to achieve my goal (I'm not super familiar to dotnet itself either).
Any feedback is welcome.

### DIFF
--- a/src/System.CommandLine.Hosting/HostingExtensions.cs
+++ b/src/System.CommandLine.Hosting/HostingExtensions.cs
@@ -16,7 +16,8 @@ namespace System.CommandLine.Hosting
 
         public static CommandLineBuilder UseHost(this CommandLineBuilder builder,
             Func<string[], IHostBuilder> hostBuilderFactory,
-            Action<IHostBuilder> configureHost = null) =>
+            Action<IHostBuilder> configureHost = null,
+            bool runAsDaemon = false) =>
             builder.UseMiddleware(async (invocation, next) =>
             {
                 var argsRemaining = invocation.ParseResult.UnparsedTokens.ToArray();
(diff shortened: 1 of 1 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/command-line-api?pullRequest=1243) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/command-line-api?pullRequest=1243) before we can accept your contribution.<br/><br/>:x: joemphilips [sign now](https://cla.dotnetfoundation.org/dotnet/command-line-api?pullRequest=1243)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/command-line-api?pullRequest=1243) it.</sub>

src/System.CommandLine.Hosting/HostingExtensions.cs
```diff
@@ -43,16 +44,23 @@ public static class HostingExtensions
...
 
-                await host.StartAsync();
-
-                await next(invocation);
-
-                await host.StopAsync();
+                if (runAsDaemon)
+                {
+                    await next(invocation);
+                    await host.RunAsync();
```
WeihanLi, line 50: Maybe better using the cancellation token from the invocationContext so that we could graceful stop

Issue #1265, opened 2021-04-28
dotnet-format broken from removal of Option.Argument setter


kzu:
I just noticed that the [FormatCommand](https://github.com/dotnet/format/blob/main/src/FormatCommand.cs#L48-L84) customizes options by setting the argument directly.

What's the migration path for scenarios like those?

[edit]

At least these two seem to be (mostly) equivalent:

```csharp
new Option(new[] { "--fix-style", "-s" })
{
    Argument = new Argument<string?>("severity") { Arity = ArgumentArity.ZeroOrOne }.FromAmong(SeverityLevels)
},
```

and

```csharp
new Option<string?>(new[] { "--fix-style", "-s" }) { Name = "severity" }.FromAmong(SeverityLevels),
```

The only issue is that the second option (compatible with latest package) actually adds "severity" as an alias to the option, which the former doesn't.

jonsequitur:
What's your goal in setting `Name = "severity"`? Are you doing this in order to customize help or for model binding or value lookup from the parse result?

kzu:
This is not my CLI app, it's dotnet/format. No idea what or why they are setting that :). But asking on their side of the issue might help answer that: https://github.com/dotnet/format/issues/1111

JoeRobich:
> What's your goal in setting `Name = "severity"`?

@jonsequitur It is to customize the help message.

jonsequitur:
Got it. For that, you can now do this:

```cs
new Option<string?>(new[] { "--fix-style", "-s" }) { ArgumentHelpName = "severity" }.FromAmong(SeverityLevels)
```

PR #1266, opened 2021-04-28
Add missing overload on Option<T> with multiple aliases and arity

kzu:
Seems like an oversight that this wasn't there already, and it makes it impossible to have a typed option with a given arity and multiple aliases, unless you declare it as a variable and invoke `AddAlias` explicilty.

Discovered while trying to figure out how to fix https://github.com/dotnet/command-line-api/issues/1265

### DIFF
--- a/src/System.CommandLine/Option{T}.cs
+++ b/src/System.CommandLine/Option{T}.cs
@@ -16,8 +16,9 @@ namespace System.CommandLine
 
         public Option(
             string[] aliases,
-            string? description = null) 
-            : base(aliases, description, new Argument<T>())
+            string? description = null,
+            IArgumentArity? arity = null)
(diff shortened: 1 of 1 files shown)


elgonzo:
Nice find. However, i would tend going as far as replacing all six `Option<T>` constructors with just two constructors that are equivalent to the two public `Option` constructors (without the _argumentType_ argument that would be unnecessary for `Option<T>`, of course).

kzu:
Yep, simplifying is good. Wasn't sure if such a drastic change was still possible, but I'm all for it 👍

jonsequitur:
As of #1295, you can now set the arity on an option. The approach taken was a little different, removing rather than adding constructors on `Option<T>`. I'm going to close this but please open an issue if this doesn't work for you.

PR #1270, opened 2021-05-06
Fix version option validation

JoeRobich:
As reported at https://github.com/dotnet/format/issues/1136

The version option validation (introduced in https://github.com/dotnet/command-line-api/pull/1124) treats default options and arguments as if they were passed by the user. This PR adds an IsImplicit property to ArgumentResult and filters implicit argument and option results when validating the version option.

### DIFF
--- a/src/System.CommandLine/Builder/CommandLineBuilderExtensions.cs
+++ b/src/System.CommandLine/Builder/CommandLineBuilderExtensions.cs
@@ -527,14 +527,15 @@ namespace System.CommandLine.Builder
                 parseArgument: result =>
                 {
                     var commandChildren = result.FindResultFor(command)?.Children;
-                    if (commandChildren == null)
+                    if (commandChildren is null)
                     {
                         return true;
                     }
 
                     var versionOptionResult = result.Parent;
-                    foreach (var symbolResult in commandChildren)
+                    for (int i = 0; i < commandChildren.Count; i++)
                     {
+                        var symbolResult = commandChildren[i];
                         if (symbolResult == versionOptionResult)
                         {
                             continue;
--- a/src/System.CommandLine/Builder/CommandLineBuilderExtensions.cs
+++ b/src/System.CommandLine/Builder/CommandLineBuilderExtensions.cs
@@ -380,7 +380,7 @@ namespace System.CommandLine.Builder
             return builder.UseHelp(helpOption);
         }
 
-        public static TBuilder UseHelpBuilder<TBuilder>(this TBuilder builder, 
+        public static TBuilder UseHelpBuilder<TBuilder>(this TBuilder builder,
             Func<BindingContext, IHelpBuilder> getHelpBuilder)
             where TBuilder : CommandLineBuilder
         {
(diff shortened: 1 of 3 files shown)


src/System.CommandLine/Parsing/ArgumentResult.cs
```diff
@@ -20,9 +20,11 @@ public class ArgumentResult : SymbolResult
 
         public IArgument Argument { get; }
 
+        public bool IsImplicit => PassedOnTokens == null;
```
jonsequitur, line 23: I'm not sure this is correct. `IsImplicit` means something a bit different in the existing `OptionResult.IsImplicit`, which is that the token was generated by the parser as a sort of placeholder for a default value. 

At the least, the `PassedOnTokens == null` case looks like it might only accidentally correlate with the presence of a default value. This collection is _intended_ to capture the overflow when a custom parser doesn't want to consume all of the available tokens.

src/System.CommandLine/Builder/CommandLineBuilderExtensions.cs
```diff
@@ -526,7 +526,7 @@ public static CommandLineBuilder UseHelp(this CommandLineBuilder builder)
                 description: Resources.Instance.VersionOptionDescription(),
                 parseArgument: result =>
                 {
-                    if (result.FindResultFor(command)?.Children.Count > 1)
+                    if (result.FindResultFor(command)?.Children.Where(IsNotImplicit).Count() > 1)
```
jonsequitur, line 529: Can we do this without the LINQ? A bunch of LINQ removal was done for performance reasons.

I wonder if a simpler implementation could work based on just the token count?

src/System.CommandLine/Builder/CommandLineBuilderExtensions.cs
```diff
@@ -526,8 +526,14 @@ public static CommandLineBuilder UseHelp(this CommandLineBuilder builder)
...
                 {
+                    var commandChildren = result.FindResultFor(command)?.Children;
+                    if (commandChildren == null)
+                    {
+                        return true;
+                    }
+
                     var versionOptionResult = result.Parent;
-                    foreach (var symbolResult in result.FindResultFor(command)?.Children)
+                    foreach (var symbolResult in commandChildren)
```
jonsequitur, line 536: I think you can use a `for` loop here.

src/System.CommandLine/Builder/CommandLineBuilderExtensions.cs
```diff
@@ -526,8 +526,14 @@ public static CommandLineBuilder UseHelp(this CommandLineBuilder builder)
                 description: Resources.Instance.VersionOptionDescription(),
                 parseArgument: result =>
                 {
+                    var commandChildren = result.FindResultFor(command)?.Children;
+                    if (commandChildren == null)
```
jonsequitur, line 530:
```suggestion
                    if (commandChildren is null)
```

src/System.CommandLine/Parsing/ArgumentResult.cs
```diff
@@ -20,9 +20,11 @@ public class ArgumentResult : SymbolResult
 
         public IArgument Argument { get; }
 
+        public bool IsImplicit => Argument.HasDefaultValue && Tokens.Count == 0;
```
jonsequitur, line 23: Let's keep this internal. To make it public I'd want to make sure we understand its use case better and test it more directly.

JoeRobich:
```suggestion
        internal bool IsImplicit => Argument.HasDefaultValue && Tokens.Count == 0;
```

jonsequitur (approved):

PR #1278, opened 2021-05-14
ConsoleRenderer throw NotSupportedException in case of console is no ITerminal

### LINKED ISSUE #1249: Rendering: Improvement suggestion for ConsoleRenderer
just a stupid thing, I am not using DragonFruit and building a Hosting commandline app, but I simply forgot to add `.UseAnsiTerminalWhenAvailable()` to my `CommandLineBuilder`.

This took me a whole day until I understood the issue, as I only got a sense less NullRef Exception from anywhere, after calling `screen.Render()`

I would like to have an earlier and much clearer Exception:
e.g. here 

https://github.com/dotnet/command-line-api/blob/722ada8a237d940eb6234792bfa054e1c351129d/src/System.CommandLine.Rendering/ConsoleRenderer.cs#L18-L24

something like
```csharp
            _terminal = console as ITerminal ??
                        throw new NotSupportedException($"Provided {nameof(console)} is no Terminal.");
```

BobSilent:
Fixes #1249

### DIFF
--- a/src/System.CommandLine.Rendering.Tests/Views/GridViewTests.cs
+++ b/src/System.CommandLine.Rendering.Tests/Views/GridViewTests.cs
@@ -19,8 +19,8 @@ namespace System.CommandLine.Rendering.Tests.Views
             grid.SetColumns(ColumnDefinition.Star(1), ColumnDefinition.Star(1));
             grid.SetRows(RowDefinition.Star(1), RowDefinition.Star(1));
 
-            var console = new TestConsole();
-            var renderer = new ConsoleRenderer(console);
+            var terminal = new TestTerminal();
+            var renderer = new ConsoleRenderer(terminal);
             Size measuredSize = grid.Measure(renderer, new Size(10, 10));
 
             measuredSize.Width.Should().Be(10);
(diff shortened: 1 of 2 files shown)


elgonzo:
I don't know for sure, but it seems a console not being a `ITerminal` is meant to be possible. Looking at the source of ConsoleRenderer, it seems that the requirement of a console being or not having to be an ITerminal depends on the OutputMode argument of the ConsoleRenderer constructor.

Your pull request changes this, making the requirement of the console being an ITerminal a requirement regardless of the chosen OutputMode. While it makes sense to add the ITerminal check here, it should however be gated by the actual output modes that require a terminal. (Unfortunately, due to lacking documentation, i can't tell with certainty which output modes would require a console to be an ITerminal.) Probably best if @jonsequitur would chime in here...

BobSilent:
@elgonzo ok I see the point, that's why I originally wanted the discussion here: #1249
but as I did not get any feedback, I created this PR to discuss it.

As I described in #1249 for me it was simply a wrong usage, but it was quite hard to find the root of the error, I can live with cancelling this PR

jonsequitur:
This is correct. The idea is that the renderers should be able to act on a console or on a terminal. The intended distinction here is that when output is redirected, there is no terminal. `System.Console` cursor positioning and other APIs will throw in this case. The renderer should still be able to output structures such as tables, but it needs to do so via a more limited set of APIs, e.g. by explicitly writing out whitespace.

BobSilent:
Ok i see, a first quick test shows me, that a newer release does not throw a null ref exception, so i will close this PR and the issue.

Thanks for your answers

PR #1280, opened 2021-05-14
GridView preserve column width definition after sorting for size calculation

BobSilent:
Fix column width definition, after calculation of width in a mixed size mode, the column width was not according to definition but according to sorting (first all fixed column sizes, size to content, star size)


BobSilent:
@jonsequitur can you please review this PR, as it is quite annoying that currently I cannot mix the column width definitions Star, Fix, SizeToContent in a Grid.

If I define 3 Columns: 
- Fix(10)
- Star(1)
- Fix(10)

I would assume that middle Column would be dynamic in width depending on console width, where as first and last are always sized to 10. 
But due to a Bug in Size Calculation, the first and second column are set to fixed size 10 and the last column will be the dynamic one.
This gets even more worth if you have more columns and mix all three size definitions.

jonsequitur (approved):

jonsequitur:
Thanks, @BobSilent!

PR #1288, opened 2021-05-25
Add Arity to Option<T> ctor with multiple aliases

mahardan-oup:
Suggest adding IArgumentArity? arity argument to Option<T>(string[], string?) ctor.

This maintain parity with Option<T>(string, string?, IArgumentArity?) ctor and prevents much frustration when wanting to add a second alias to an Option requires rewriting the ctor altogether in order to keep the ArgumentArity.

### DIFF
--- a/src/System.CommandLine/Option{T}.cs
+++ b/src/System.CommandLine/Option{T}.cs
@@ -16,8 +16,9 @@ namespace System.CommandLine
 
         public Option(
             string[] aliases,
-            string? description = null) 
-            : base(aliases, description, new Argument<T>())
+            string? description = null,
+            IArgumentArity? arity = null) 
+            : base(aliases, description, new Argument<T> { Arity = arity })
         { }
 
         public Option(


jonsequitur:
These constructors are already complex, and arity is a useful property to be able to set after construction.

I'd suggest instead that we add a property (`Option.Arity`?) that delegates to the internal `Argument.Arity` property.

mahardan-oup:
If what you say is true, why have an Arity parameter in `Option(string, string?, IArgumentArity?)` ?
That's not to say there shouldn't be an `Option.Arity` property too, but adding the argument to the `Option(string[], string?)` ctor reduces perceived complexity from the developer's (my) point of view.

jonsequitur:
I'm realizing that there are too many `Option` constructors overall. Some of these extra constructors are holdovers from when these classes were enforcing immutability, but arity is a less commonly used property and probably doesn't belong on any of the constructors.

I'll do a cleanup PR for this.

mahardan-oup:
@jonsequitur Apologies - this is my first PR to a project like this. Do I close this PR when #1295 is closed \ merged or will it do it automatically?

jonsequitur:
#1295  is merged and I believe it accomplishes the goal you were looking for by slightly different means. It exposes custom arity for `Option<T>` via a property instead of a constructor. It was a little more complex than I expected so as I investigated it, rather than asking you to do it, I just created a separate PR.

Feel free to close this and if you think there's still work to be done, please open an issue.

mahardan-oup:
Closing.

PR #1289, opened 2021-05-25
fix #1137

### LINKED ISSUE #1137: Unexpected exception for classes having a ReadOnlySpan<> constructor
Dear,

Since I upgraded to the latest version of the nuget package (2.0.0-beta1.20574.7), coming from previous version (2.0.0-beta1.20371.2).
I'm facing an exception when I'm not specifying some options on the argument line. The previous package didn't cause that exception to be thrown. 
Apparently this exception occurs as I've options that use `IPAddress` type for which I use a `parseArgument` function but `IPAddress` has a constructor with a `ReadOnlySpan<byte>` argument that apparently causes the exception when the option isn't specified in the command line.

Here is a small .NET 5 program that reproduces the issue:
```
using System;
using System.Net;
using System.CommandLine;
using System.CommandLine.Binding;

#nullable enable

RootCommand root = new RootCommand()
{
    new Option<IPAddress?>("--host", parseArgument: ConvertIPAddress, description: "IP address of the server"),
    new Option<ushort>("--port"),
};
root.Handler = CommandHandler.Create<ConnectionOptions>(options=> {Console.WriteLine($"Args: {options.Host}:{options.Port}");} );

root.Invoke("--host 2.3.4.5 --port 5678 ");
root.Invoke("--host 2.3.4.5");
// throw exception System.InvalidOperationException: The type System.ReadOnlySpan`1[System.Byte] cannot be bound
root.Invoke("--port 5678");

static IPAddress? ConvertIPAddress(ArgumentResult result)
{
    if (result.Tokens.Count == 1)
    {
        if (IPAddress.TryParse(result.Tokens[0].Value, out IPAddress? outAddress))
            return outAddress;
        result.ErrorMessage=$"Invalid IPAddress : {result.Tokens[0].Value}";
        return null;
    }
    return IPAddress.Any;
}

public class ConnectionOptions
{
    public IPAddress? Host { get; set; } = null;
    public ushort Port { get; set; } = 1234;
}

```

jonsequitur:


oblaise:
Fix seems to apply for the decimal issue but not for the more generic ReadOnlySpan<> issue that is described in #1137

PR #1293, opened 2021-05-29
Fix 1292

### LINKED ISSUE #1292: Attribute Validator not called when option with value exists
Given the following Code:

```csharp
{
   var option = new Option<string>("option", "Demo Option") { IsRequired = true };
   option.AddValidator(val =>
   {
      Console.WriteLine("Option validator called");
      return null;
   });

   var argument = new Argument<string>("argument");
   argument.AddValidator(val =>
   {
     Console.WriteLine("Argument validator called");
     return null;
   });

   var rootCommand = new RootCommand("My Tool")
   {
      option,
      argument
    };

    rootCommand.AddValidator(val =>
    {
      Console.WriteLine("Root command validator called");
      return null;
     });

    rootCommand.Invoke(args);
}
```

and executing
`.\CommandLineParser.exe option=bla asd`
i would expect an output of

```
Root command validator called
Option validator called
Argument validator called
```

however the output is:
```
Root command validator called
Option validator called
```

this only happens if the option expect a value, if i change
`var option = new Option<string>("option", "Demo Option") { IsRequired = true };`
to
`var option = new Option("option", "Demo Option") { IsRequired = true };`

the Argument validator is called.

Not sure if this intended behaviour, but i think you should be able to validate an argument regardless of precending options requiring a value..

jonsequitur:
Fixes #1292

### DIFF
--- a/src/System.CommandLine.Tests/ArgumentTests.cs
+++ b/src/System.CommandLine.Tests/ArgumentTests.cs
@@ -9,6 +9,7 @@ using FluentAssertions;
 using System.Linq;
 using System.Threading.Tasks;
 using Xunit;
+using Xunit.Abstractions;
 
 namespace System.CommandLine.Tests
 {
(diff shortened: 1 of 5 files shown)


Spacefish (commented):
I think the LINQ version was more readable TBH.. But yeah, doesnt matter in the end.

jonsequitur:
@Spacefish Agreed, it was more readable. We've been removing LINQ usage in many places though because it does have an effect on performance and memory footprint.

PR #1295, opened 2021-05-30
remove arity parameter from Option<T> constructor, expose it as a property

jonsequitur:


jonsequitur:
Now, that said, there are also constructors on the base `Option` (non-generic) class that accept arity. In that case though, there's no generic type that can be used to infer the arity like there is on `Option<T>`. 

Does it make more sense there? Should they be consistent? Instead of this, should we go ahead with #1288?

Keboo (approved):

Issue #1322, opened 2021-06-23
Allow customizing apartment model when using System.CommandLine.DragonFruit


andrewhickman:
Hi,

I tried creating a small command line app which launches a browser using the [WebBrowser](https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.webbrowser?view=net-5.0) class. 

The following code

<details>
<summary>
Program.cs
</summary>

```
using System;
using System.Windows.Forms;

namespace MyApp
{
    public class Program
    {
        [STAThread]
        public static void Main()
        {
            new WebBrowser();
        }
    }
}
```
</details>

<details>
<summary>
MyApp.csproj
</summary>

```
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <DisableWinExeOutputInference>true</DisableWinExeOutputInference>
    <TargetFramework>net5.0-windows</TargetFramework>
    <UseWindowsForms>true</UseWindowsForms>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="System.CommandLine.DragonFruit" Version="0.3.0-alpha.21216.1" />
  </ItemGroup>

</Project>
```

</details>

complains it needs to run in a single-threaded apartment even though I marked `Main` with `[STAThread]`:

<details>
<summary>
Stacktrace
</summary>

```
Unhandled exception: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.Threading.ThreadStateException: ActiveX control '8856f961-340a-11d0-a96b-00c04fd705a2' cannot be instantiated because the current thread is not in a single-threaded apartment.
   at System.Windows.Forms.WebBrowserBase..ctor(String clsidString)
   at System.Windows.Forms.WebBrowser..ctor()
   at MyApp.Program.Main() in D:\<redacted>\Program.cs:line 11
   --- End of inner exception stack trace ---
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor, Boolean wrapExceptions)
   at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at System.Reflection.MethodBase.Invoke(Object obj, Object[] parameters)
   at System.CommandLine.Invocation.ModelBindingCommandHandler.InvokeAsync(InvocationContext context)
[...]

PR #1324, opened 2021-06-25
Run synchronous command handlers on the same thread

andrewhickman:
Currently, when not using any asynchronous handlers, `InvocationPipeline.InvokeAsync` will run them on the current thread whereas `InvocationPipeline.Invoke` runs them on a different thread. This PR fixes that inconsistency. This makes scenarios like #1322 easier to handle because we can assume the handlers run on the main thread

### DIFF
--- a/src/System.CommandLine.Tests/Invocation/InvocationPipelineTests.cs
+++ b/src/System.CommandLine.Tests/Invocation/InvocationPipelineTests.cs
@@ -9,6 +9,7 @@ using System.CommandLine.IO;
 using System.CommandLine.Parsing;
 using System.Linq;
 using System.Reflection;
+using System.Threading;
 using System.Threading.Tasks;
 using FluentAssertions;
 using Xunit;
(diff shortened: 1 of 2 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/command-line-api?pullRequest=1324) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/command-line-api?pullRequest=1324) before we can accept your contribution.<br/><br/>:x: andrewhickman [sign now](https://cla.dotnetfoundation.org/dotnet/command-line-api?pullRequest=1324)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/command-line-api?pullRequest=1324) it.</sub>

Keboo:
The reason for wrapping the call in a Task.Run is because this library doesn't know what the command handlers may do. Wrapping an async method in a synchronous one is actually quite difficult to do well and avoid dead locks (especially when you don't know what the method may need to do). Stephan Toub has an [excellent article](https://devblogs.microsoft.com/pfxteam/should-i-expose-synchronous-wrappers-for-asynchronous-methods/) on this. Specifically the section on "What if I really do need “sync over async”?" the implementation with Task.Run is intentional to offload the call to the threadpool where we won't have to worry about any SynchronizationContext.

PR #1335, opened 2021-07-01
Switch the top level help item to be Description: rather than the name of the command which was inconsistent

### LINKED ISSUE #1276: Add "Description" to help
6.0.100-preview.4.21209.4 

We've added the description to help (yeah!!!). However, it currently appears under the subcommand name. To better mirror the other headers like "Usage", I suggest we use "Description:" instead. 

image

marcpopMSFT:
Fixes #1276


marcpopMSFT:
@jonsequitur when I built, it correctly updated all of the xlf files but it didn't generate the resources.cs and resources.designer.cs files so I hand edited those. Let me know if there was a step I was supposed to do to update those.

src/System.CommandLine/Properties/xlf/Resources.fr.xlf
```diff
@@ -126,6 +126,11 @@ ERR:
         <target state="new">Commands:</target>
         <note />
       </trans-unit>
+      <trans-unit id="HelpDescriptionTile">
```
jonsequitur, line 129: I think this and a few others should be `"*Title"` rather than `"*Tile"`.

jonsequitur: I fixed the existing ones here: #1347

jonsequitur (approved):

jonsequitur:
@marcpopMSFT Please rebase this onto `main`.

marcpopMSFT:
@jonsequitur To clarify your checkin process, do you want me to click "rebase and merge" or do a rebase locally or something else? My other repos don't follow this process (msbuild squashes and sdk just merges)

jonsequitur:
Ah, nevermind. I'll do a merge commit.

PR #1337, opened 2021-07-02
obsoleting and removing methods in pursuit of #1127

jonsequitur:
Working on #1127.


src/System.CommandLine/Parsing/ParseResult.cs
```diff
@@ -84,12 +84,14 @@ public class ParseResult
 
         public IReadOnlyList<string> UnparsedTokens => _unparsedTokens.Select(t => t.Value).ToArray();
 
+        [Obsolete("This method is obsolete and will be removed in a future version. Please use ParseResult.ValueForOption<T>(Option<T>) instead. For details see https://github.com/dotnet/command-line-api/issues/1127")]
```
Keboo, line 87: We should log an item to remove these prior to the milestone release

PR #1340, opened 2021-07-05
Fix user-specific resources not used

angelobreuer:
Currently, `System.CommandLine` does only make partial use of the `System.CommandLine.Resources` passed by the user. Instead, it used the default instance (`System.CommandLine.Resources.Instance`).
This commit fixes the occurrences where access is made to the shared instance instead of the user-passed resources.

When using the `CommandLineBuilder` it is possible to set a custom resources instance that can override the localization strings provided by `System.CommandLine`:

```csharp
await new CommandLineBuilder(...)
    .UseResources(new CommandLineResources())
    .UseDefaults()
    .Build()
    .InvokeAsync(args);
```

For example, the following output is expected:

Image demonstration the expected output of the application when specifying a custom resources instance that overrides the `HelpUsageTile` member

But the current implementation ignores the resources instance passed by the user:

Image demonstration the actual output of the application when specifying a custom resources instance that overrides the `HelpUsageTile` member

Note that the screenshots above only cover the `string HelpUsageTile()` member but *most* of the strings are not fetched from the user-passed `Resources` instance.

A workaround is possible using the following snippet. The snippet replaces the `Resources` instance (`System.CommandLine.Resources`) with an own implementation (Note that this should be never used as a real solution).

```csharp
var staticInstanceBackingField = typeof(System.CommandLine.Resources).GetField(
    $"<{nameof(Instance)}>k__BackingField",
    BindingFlags.Static | BindingFlags.NonPublic);

var method = new DynamicMethod("Do", null, new[] { typeof(System.CommandLine.Resources), });
var ilGenerator = method.GetILGenerator();

ilGenerator.Emit(OpCodes.Ldarg_0);
ilGenerator.Emit(OpCodes.Stsfld, staticInstanceBackingField!);
ilGenerator.Emit(OpCodes.Ret);

method.CreateDelegate<Action<System.CommandLine.Resources>>()(/* your resources instance */);
```

There should be "only" one public API change in the constructor of the `HelpBuilder` class:

https://github.com/angelobreuer/command-line-api/commit/4330f8f8175da409dd8e859b56edfd2efc8323e8#diff-ed9641de0c4af7c7a4bd4c48a6f1e66d495f8f0887ebbc6e945cb0f2a601b237L96-R96

```diff
- public HelpBuilder(IConsole console, int maxWidth = int.MaxValue)
+ public HelpBuilder(IConsole console, Resources? resources = null, int maxWidth = int.MaxValue)
```

In the case of an implementor wanting to inherit from HelpBuilder and change use a custom factory, it would be probably useful to provide the `Resources` instance more easily in the BindingContext because it seems to be a bit odd to retrieve the instance from `context.ParseResult.CommandResult.Resources`.

I've also provided a repro which can be found under the following link: https://github.com/angelobreuer/SCLResourcesRepro

### DIFF
--- a/src/System.CommandLine/Builder/CommandLineBuilderExtensions.cs
+++ b/src/System.CommandLine/Builder/CommandLineBuilderExtensions.cs
@@ -230,7 +230,7 @@ ERR:
                     string debuggableProcessNames = GetEnvironmentVariable(environmentVariableName);
                     if (string.IsNullOrWhiteSpace(debuggableProcessNames))
                     {
-                        context.Console.Error.WriteLine(Resources.Instance.DebugDirectiveExecutableNotSpecified(environmentVariableName, process.ProcessName));
+                        context.Console.Error.WriteLine(context.Parser.Configuration.Resources.DebugDirectiveExecutableNotSpecified(environmentVariableName, process.ProcessName));
                         context.ExitCode = errorExitCode ?? 1;
                         return;
                     }
@@ -240,7 +240,7 @@ ERR:
                         if (processNames.Contains(process.ProcessName, StringComparer.Ordinal))
                         {
                             var processId = process.Id;
-                            context.Console.Out.WriteLine(Resources.Instance.DebugDirectiveAttachToProcess(processId, process.ProcessName));
+                            context.Console.Out.WriteLine(context.Parser.Configuration.Resources.DebugDirectiveAttachToProcess(processId, process.ProcessName));
                             while (!Debugger.IsAttached)
                             {
                                 await Task.Delay(500);
@@ -248,7 +248,7 @@ ERR:
                         }
                         else
                         {
-                            context.Console.Error.WriteLine(Resources.Instance.DebugDirectiveProcessNotIncludedInEnvironmentVariable(process.ProcessName, environmentVariableName, debuggableProcessNames));
+                            context.Console.Error.WriteLine(context.Parser.Configuration.Resources.DebugDirectiveProcessNotIncludedInEnvironmentVariable(process.ProcessName, environmentVariableName, debuggableProcessNames));
                             context.ExitCode = errorExitCode ?? 1;
                             return;
                         }
@@ -329,7 +329,7 @@ ERR:
                     context.Console.ResetTerminalForegroundColor();
                     context.Console.SetTerminalForegroundRed();
 
-                    context.Console.Error.Write(Resources.Instance.ExceptionHandlerHeader());
+                    context.Console.Error.Write(context.Parser.Configuration.Resources.ExceptionHandlerHeader());
                     context.Console.Error.WriteLine(exception.ToString());
 
                     context.Console.ResetTerminalForegroundColor();
@@ -340,7 +340,7 @@ ERR:
 
         public static CommandLineBuilder UseHelp(this CommandLineBuilder builder)
         {
-            return builder.UseHelp(new HelpOption());
+            return builder.UseHelp(new HelpOption(builder.Resources ?? Resources.Instance));
         }
 
         internal static CommandLineBuilder UseHelp(
@@ -368,7 +368,7 @@ ERR:
             Action<THelpBuilder>? configureHelp)
             where THelpBuilder : IHelpBuilder
         {
-            return builder.UseHelp(new HelpOption(), configureHelp);
+            return builder.UseHelp(new HelpOption(builder.Resources ?? Resources.Instance), configureHelp);
         }
 
         internal static CommandLineBuilder UseHelp<THelpBuilder>(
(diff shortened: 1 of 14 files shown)


src/System.CommandLine/Builder/CommandLineBuilderExtensions.cs
```diff
@@ -222,7 +222,7 @@ public static CommandLineBuilder CancelOnProcessTermination(this CommandLineBuil
                     string debuggableProcessNames = GetEnvironmentVariable(environmentVariableName);
                     if (string.IsNullOrWhiteSpace(debuggableProcessNames))
                     {
-                        context.Console.Error.WriteLine(Resources.Instance.DebugDirectiveExecutableNotSpecified(environmentVariableName, process.ProcessName));
+                        context.Console.Error.WriteLine((builder.Resources ?? Resources.Instance).DebugDirectiveExecutableNotSpecified(environmentVariableName, process.ProcessName));
```
Keboo, line 225: There is a subtle change here, that now causes this lambda to capture the builder parameter. Though I don't think it is likely to cause an issue here, it might be cleaner to retrieve the resource by going through the already passed in `context` (as you noted in the description: `context.ParseResult.CommandResult.Resources`). 

The same would apply to the two other usages further down in this method.

I do also agree that the Resources should probably be exposed on `InvocationContext` directly rather than burring them down so deep.

src/System.CommandLine/Builder/CommandLineBuilderExtensions.cs
```diff
@@ -332,7 +332,7 @@ void Default(Exception exception, InvocationContext context)
 
         public static CommandLineBuilder UseHelp(this CommandLineBuilder builder)
         {
-            return builder.UseHelp(new HelpOption());
+            return builder.UseHelp(new HelpOption(builder.Resources ?? Resources.Instance));
```
Keboo, line 335: This does raise an interesting question since the value provided here could be different depending on the order that the caller invokes the `.UseResources` and `.UseHelp` extension methods. The same applies to the changes to the Version option below as well.
@jonsequitur feedback?

It seems like we should either explicitly document this behavior, or adjust the internal `.UseHelp` method lazily create the HelpOption.

src/System.CommandLine/Help/HelpBuilder.cs
```diff
@@ -17,11 +18,13 @@ public class HelpBuilder : IHelpBuilder
...
 
         protected IConsole Console { get; }
+        protected Resources Resources { get; }
         public int MaxWidth { get; }
 
-        public HelpBuilder(IConsole console, int maxWidth = int.MaxValue)
+        public HelpBuilder(IConsole console, Resources? resources = null, int maxWidth = int.MaxValue)
         {
             Console = console ?? throw new ArgumentNullException(nameof(console));
+            Resources = resources ?? Resources.Instance;
```
Keboo, line 27: IMO rather than this being an option parameter, I would make it required, non-null, and throw on null. The fact that `maxWidth` is still an optional parameter is a bit of a hold-over from a previous version of the API.
```suggestion
        public HelpBuilder(IConsole console, Resources resources, int maxWidth = int.MaxValue)
        {
            Console = console ?? throw new ArgumentNullException(nameof(console));
            Resources = resources ?? throw new ArgumentNullException(nameof(resources));
```

Keboo (commented):
Great work! Thank you for doing this.

src/System.CommandLine/Help/HelpBuilder.cs
angelobreuer: Resolved (7da17af1fbd8e26c12d652ac100ad383c4391e34).

src/System.CommandLine/Builder/CommandLineBuilderExtensions.cs
angelobreuer: Resolved (7da17af1fbd8e26c12d652ac100ad383c4391e34).

src/System.CommandLine/Builder/CommandLineBuilderExtensions.cs
jonsequitur: It's important for these methods to be callable in any order and preferably idempotent. I think it's better to avoid passing the `Resources` via `HelpOption`.

angelobreuer: @Keboo @jonsequitur I made a commit that should resolve the issue. Now the descriptions of the `HelpOption` and version option are set when building the command line parser.

jonsequitur (approved):

jonsequitur:
@angelobreuer Could you rebase this?

jonsequitur:
Thanks, @angelobreuer!

PR #1342, opened 2021-07-06
fix #1199

### LINKED ISSUE #1199: Change default for Option.AllowMultipleArgumentsPerToken to false
The behavior of allowing multiple tokens to parse to a single option is uncommon in practice. This is currently enabled by default but should probably be toggled to disabled by default. 

Related discussion: #1162.

jonsequitur:
Fixes #1199.

### DIFF
--- a/src/System.CommandLine.Tests/Binding/BindingTestCase.cs
+++ b/src/System.CommandLine.Tests/Binding/BindingTestCase.cs
@@ -11,12 +11,12 @@ namespace System.CommandLine.Tests.Binding
             string commandLineToken,
             Type parameterType,
             Action<object> assertBoundValue,
-            string variationName)
+            string variationName) : this(
+            new[] { commandLineToken },
+            parameterType,
+            assertBoundValue,
+            variationName)
         {
-            _assertBoundValue = assertBoundValue;
-            VariationName = variationName;
-            CommandLineTokens = new[] { commandLineToken };
-            ParameterType = parameterType;
         }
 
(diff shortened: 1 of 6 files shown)


sfoslund (approved):
This is great for the CLI, we're explicitly setting AllowMultipleArgumentsPerToken to false right now across all options that take multiple args. Thanks!

src/System.CommandLine.Tests/Binding/BindingTestCase.cs
```diff
@@ -6,20 +6,32 @@ namespace System.CommandLine.Tests.Binding
...
         {
             _assertBoundValue = assertBoundValue;
             VariationName = variationName;
-            CommandLine = commandLine;
+            CommandLineTokens = new[] { commandLineToken };
             ParameterType = parameterType;
         }
 
-        public string CommandLine { get; }
+        private BindingTestCase(
```
Keboo, line 22: Rather than duplicating the assignments, should the public ctor just invoke this one?

Keboo (approved):

src/System.CommandLine.Tests/Binding/BindingTestCase.cs
jonsequitur: I changed it so that one constructor calls the other. Neither is public, forcing the use of the factory method which is generic and allows for strongly-typed assertions.

Keboo (approved):

PR #1354, opened 2021-07-17
Delete unused file

danielchalmers:
I might be wrong but I think this file was added by accident in 3c96fe591ea32272a184703c213e0229b1bc9301


jonsequitur (approved):

jonsequitur:
Thanks, @danielchalmers! Nice catch.

PR #1356, opened 2021-07-19
Hosting di inject commandhandler

### LINKED ISSUE #1344: When using UseCommandHandler with subcomand, Help overrides RootCommand
Hi all,

I love this project. We are using it to build [dotnet-affected](https://github.com/leonardochaia/dotnet-affected)!

I'm not sure if this is a known issue, but I'm migrating to use the Hosting setup and I'm seeing different behavior between declaring the RootCommand with a Handler, and binding a handler through the HostBuilder.

It only happens when using a subcommand. If I remove the subcommad, it works.

For example:

```csharp
public class PublicApiTests
{
    private class SubCommand : Command
    {
        public SubCommand() 
            : base("thing")
        {
        }
    }
    private class HelloRootCommand : RootCommand
    {
        public HelloRootCommand()
        {
            this.AddCommand(new SubCommand());
        }
    }
    
    private class HelloHandler : ICommandHandler
    {
        {
            context.Console.Out.WriteLine("Hello World");
            return Task.FromResult(0);
        }
    }

    [Fact]
    public void Should_Print_HelloWorld()
    {
        var console = new TestConsole();
        new CommandLineBuilder(new HelloRootCommand() {Handler = new HelloHandler()})
            .UseDefaults()
            .UseHost(host =>
            {
                // .. configure DI
            })
            .Build()
            .Invoke("", console);
        
        Assert.Contains("Hello World", console.Out.ToString());
    }
    
   // this test fails! It prints the help
    [Fact]
    public void With_DI_Handler_Should_Print_HelloWorld()
    {
        var console = new TestConsole();
        new CommandLineBuilder(new HelloRootCommand())
            .UseHost(host =>
            {
                // .. configure DI
                host.UseCommandHandler<HelloRootCommand, HelloHandler>();
            })
            .UseDefaults()
            .Build()
            .Invoke("", console);
        
        Assert.Contains("Hello World", console.Out.ToString());
    }
}
```

fredrikhr:
Following the discussion in #1344 we have detected flaws in the implementation of the `UseCommandHandler` extension methods.

This PR adds a new helper class `HostedCommandHandler` which works similarly to `CommandHandler` by creating instances of `ICommandHandler` that forward the call to `InvokeAsync(InvocationContext)` to a handler instance constructed by the .NET Generic Host DI-system. 

`UseCommandHandler` is removed by this PR.

Test have also been updated to use the `HostedCommandHandler` instead of `UseCommandHandler`.

The test have also been adjusted to better reflect design principles using the .NET Generic Host. In particular this involves having options-classes that hold the values bound from command-line. The handlers get an injected `IOptions<T>` instance from which they read bound command-line argument values.

**/cc** @leonardochaia, @Gronex

A minimalistic example of using DI-bound handlers would look like this:

``` C#
public static class Program
{
    public static Task<int> Main(string[] args)
    {
        {
            Handler = HostedCommandHandler.CreateFromHost<MyCommandHandler>()
        };
        var parser = new CommandLineBuilder(rootCommand)
            .UseHost(CreateHostBuilder)
            .Build();
        return parser.InvokeAsync(args);
    }

    public static IHostBuilder CreateHostBuilder(string[] args) =>
        Host.CreateDefaultBuilder(args)
            .ConfigureServices(services =>
            {
                services.AddTransient<MyCommandHandler>();
                services.AddOptions<MyAppOptions>()
                    .BindCommandLine()
                    ;
            })
            ;
}

class MyCommandHandler : ICommandHandler
{
    private readonly MyAppOptions options;

    public MyCommandHandler(IOptions<MyAppOptions> options) : base()
    {
        this.options = options.Value;
    }

    {
        /* Do something here */
    }
}

class MyAppOptions { /* Add properties here which match Options/Arguments from command line */ }
```

closes #1344

### DIFF
--- a/src/System.CommandLine.Hosting.Tests/HostingHandlerTest.cs
+++ b/src/System.CommandLine.Hosting.Tests/HostingHandlerTest.cs
@@ -113,6 +113,27 @@ namespace System.CommandLine.Hosting.Tests
             service.StringValue.Should().Be("TEST");
         }
 
+        [Fact]
+        public static void Throws_When_Injected_HandlerType_is_not_ICommandHandler()
+        {
+            new object().Invoking(_ =>
+            {
+                var handlerWrapper = HostedCommandHandler.CreateFromHost(
+                    typeof(MyNonCommandHandler));
+            }).Should().ThrowExactly<ArgumentException>(
+                because: $"{typeof(MyNonCommandHandler)} does not implement {typeof(ICommandHandler)}"
+            );
+        }
+
+        [Fact]
+        public static void Throws_When_Injected_HandlerType_is_null()
+        {
+            new object().Invoking(_ =>
+            {
+                var handlerWrapper = HostedCommandHandler.CreateFromHost(null);
+            }).Should().ThrowExactly<ArgumentNullException>();
+        }
+
         public class MyCommand : Command
         {
             public MyCommand() : base(name: "mycommand")
@@ -178,6 +199,11 @@ namespace System.CommandLine.Hosting.Tests
             public int Value { get; set; }
 
             public string StringValue { get; set; }
+        }
+
+        public class MyNonCommandHandler
+        {
+            public static int DoSomething() => 0;
         }
     }
 }
--- a/src/System.CommandLine.Hosting.Tests/HostingHandlerTest.cs
+++ b/src/System.CommandLine.Hosting.Tests/HostingHandlerTest.cs
@@ -1,16 +1,15 @@
 ﻿using System.CommandLine.Binding;
 using System.CommandLine.Builder;
 using System.CommandLine.Invocation;
-using System.CommandLine.IO;
 using System.CommandLine.Parsing;
-using System.Linq;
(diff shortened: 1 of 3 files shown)


fredrikhr:
Currently this PR only constructs the Handler from .NET Generic Host DI.

We could extend the logic to do the construction in the following order:

1. Attempt `host.Services.GetService<THandler>()`
    1.1. If successful, get or create a `ModelBinder<THandler>` and use `UpdateInstance` on the constructed handler instance.
    1.2. Return instance
2. Get or create `ModelBinder<THandler>`
    2.1. Call `CreateInstance` to create `THandler` from `BindingContext`
    2.2. Return instance

This would allow the handler to be constructed **both** from the hosted DI-system as well as from Added services from the command-line binding context. The additional call to `UpdateInstance` would also allow the `ModelBinder` to bind command-line arguments directly to the handler instance instead of requiring an `Options` type, this would simply be a shortcut to the normal .NET Generic Host way of doing things.

PR #1358, opened 2021-07-20
Simplified usage of BindingContext resolved command handlers

fredrikhr:
This PR adds convenience helper methods to simplify the usage of services resolved by the BindingContext.

* Add static helper `FromBindingContext` to `CommandHandler`.
  Creates an `ICommandHandler` where the specified command handler type is filled in at runtime by the BindingContext. This requires that the BindingContext knows how to instantiate the handler type (requires service registration).
* Add a `ConfigureBindingContext` middleware.
  Middleware simply takes an `Action<BindingContext>` parameter. In this callback the developer can call `AddModelBinder` and `AddService`. This simplifies service-registration on the BindingContext.
* Add no-parameter `AddService` instance method on `BindingContext`.
  Simplified service registration for trivial types. Uses a `ModelBinder` to auto-construct the service.

inspired by discussion in #1344.

**/cc** @leonardochaia

### DIFF
--- a/src/System.CommandLine.Tests/Invocation/CommandHandlerTests.cs
+++ b/src/System.CommandLine.Tests/Invocation/CommandHandlerTests.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System.CommandLine.Binding;
+using System.CommandLine.Builder;
 using System.CommandLine.Invocation;
 using System.CommandLine.IO;
 using System.CommandLine.Parsing;
@@ -322,7 +323,6 @@ namespace System.CommandLine.Tests.Invocation
             boundContext.ParseResult.ValueForOption(option).Should().Be(123);
         }
 
-
         private class ExecuteTestClass
         {
             public string boundName = default;
@@ -506,6 +506,39 @@ namespace System.CommandLine.Tests.Invocation
         {
             public override Task<int> InvokeAsync(InvocationContext context)
                 => Task.FromResult(41);
+        }
+
+        [Fact]
+        public static void FromBindingContext_forwards_invocation_to_bound_handler_type()
+        {
+            var command = new RootCommand
+            {
+                Handler = CommandHandler.FromBindingContext<BindingContextResolvedCommandHandler>()
+            };
+            command.Handler.Should().NotBeOfType<BindingContextResolvedCommandHandler>();
+            var parser = new CommandLineBuilder(command)
+                .ConfigureBindingContext(context => context.AddService<BindingContextResolvedCommandHandler>())
+                .Build();
+
+            var console = new TestConsole();
+            parser.Invoke(Array.Empty<string>(), console);
+            console.Out.ToString().Should().Be(typeof(BindingContextResolvedCommandHandler).FullName);
+        }
+
+        public class BindingContextResolvedCommandHandler : ICommandHandler
+        {
+            public BindingContextResolvedCommandHandler(IConsole console)
+            {
+                Console = console;
+            }
+
+            public IConsole Console { get; }
+
+            public Task<int> InvokeAsync(InvocationContext context)
+            {
+                Console.Out.Write(GetType().FullName);
+                return Task.FromResult(0);
+            }
         }
     }
 }
--- a/src/System.CommandLine/Binding/BindingContext.cs
+++ b/src/System.CommandLine/Binding/BindingContext.cs
@@ -51,7 +51,7 @@ namespace System.CommandLine.Binding
 
         internal ServiceProvider ServiceProvider { get; }
 
-        public void AddModelBinder(ModelBinder binder) => 
+        public void AddModelBinder(ModelBinder binder) =>
             _modelBindersByValueDescriptor.Add(binder.ValueDescriptor.ValueType, binder);
 
         public ModelBinder GetModelBinder(IValueDescriptor valueDescriptor)
@@ -67,7 +67,7 @@ namespace System.CommandLine.Binding
         {
             ServiceProvider.AddService(serviceType, factory);
         }
-        
+
         public void AddService<T>(Func<IServiceProvider, T> factory)
         {
             if (factory is null)
@@ -77,6 +77,22 @@ namespace System.CommandLine.Binding
 
             ServiceProvider.AddService(typeof(T), s => factory(s));
         }
+
+        public void AddService(Type serviceType)
+        {
+            object factory(IServiceProvider serviceProvider)
+            {
+                var bindingContext =
+                    serviceProvider.GetService(typeof(BindingContext)) as BindingContext
+                    ?? this;
+                var valueDescriptor = new ModelBinder.AnonymousValueDescriptor(serviceType);
+                var modelBinder = bindingContext.GetModelBinder(valueDescriptor);
+                return modelBinder.CreateInstance(bindingContext)!;
+            }
+            AddService(serviceType, factory);
+        }
+
+        public void AddService<T>() => AddService(typeof(T));
 
         internal bool TryGetValueSource(
             IValueDescriptor valueDescriptor,
@@ -108,8 +124,8 @@ namespace System.CommandLine.Binding
                 else
                 {
                     var parsed = ArgumentConverter.ConvertObject(
-                        valueDescriptor as IArgument ?? new Argument(valueDescriptor.ValueName), 
-                        valueDescriptor.ValueType, 
+                        valueDescriptor as IArgument ?? new Argument(valueDescriptor.ValueName),
+                        valueDescriptor.ValueType,
                         value,
                         resources);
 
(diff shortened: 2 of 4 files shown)


fredrikhr:
I added https://github.com/dotnet/command-line-api/pull/1358/commits/2f2796b13d15b3965bfd6e4a71530cb280ec6a05 following the [suggestion](https://github.com/dotnet/command-line-api/issues/1344#issuecomment-883463843) from @leonardochaia

leonardochaia:
Awesome! It looks like I would be able to remove my DI workaround with this changes.

This allows to override injected dependencies for testing purposes!

Can we please add this last test case? :)
```csharp
var parser = new CommandLineBuilder(command)
   .ConfigureBindingContext(context => context.AddService<IGreeter, GreeterImpl>())
   .ConfigureBindingContext(context => context.AddService<IGreeter>(sp=> new GreeterMock())) // would use Moq or something here
   .Build();
```

`IGreeter` should be of type `GreeterMock` 
I think this should work since the IServiceProvider impl has a Dictionary with a Type for the key as far as I remember.

fredrikhr:
@leonardochaia I don't understand what you mean? The first one will be done by this PR, the second one you could always do already?

leonardochaia:
Sorry, should have been clearer.

I'm just trying to be sure that the second `ConfigureBindingContext` will override the first one (and don't throw, or silently ignore it)

That's why I would assert that
> IGreeter should be of type GreeterMock

fredrikhr:
The second one thows??? Really? What exception do you get? 🤔

leonardochaia:
I'm sorry (again). Didn't mean to say it is throwing right now. My point is that, if the expected API is that multiple calls to `ConfigureBindingContext` should be supported, and the latter call should be able to override things configured on previous calls, we should have a test that makes sure that that behavior is not gonna be broken in the future by someone i.e changing the inner code to throw when a service is already registered or something like that.

So that's why I propose we add a test that represents the actual use case of overriding an already registered service.

Something like `Registered_services_can_be_overriden_by_subsequent_calls_to_ConfigurreBindingContext`

leonardochaia:
This looks awesome! :tada:

src/System.CommandLine/Binding/BindingContext.cs
```diff
@@ -65,19 +65,38 @@ public ModelBinder GetModelBinder(IValueDescriptor valueDescriptor)
 
         public void AddService(Type serviceType, Func<IServiceProvider, object> factory)
         {
+            _ = serviceType ?? throw new ArgumentNullException(nameof(serviceType));
+            _ = factory ?? throw new ArgumentNullException(nameof(factory));
```
jonsequitur, line 69: It's preferable to use `if` blocks for these.

jonsequitur:
The new `AddService` overloads increase the scope of the DI features in ways that might be misleading and will likely lead to additional issues being opened when people misunderstand the goal. The pre-existing `AddService` method accepted a `Func` so that you can know when it's called and control things like lifetime management, singleton versus transient instances, and recursive resolution, none of which System.CommandLine's service provider currently does for external types.

src/System.CommandLine.Tests/Invocation/CommandHandlerTests.cs
```diff
@@ -507,5 +507,101 @@ public class OverridenVirtualTestCommandHandler : VirtualTestCommandHandler
...
+            parser.Invoke(Array.Empty<string>(), console);
+            console.Out.ToString().Should().Be(typeof(BindingContextResolvedCommandHandler).FullName);
+        }
+
+        [Fact]
+        public static void Subsequent_call_to_configure_overrides_service_registration()
+        {
+            BindingContextCommandHandlerAction action = (handler, Console) =>
+            {
+                handler.Should().BeOfType<BindingContextCommandHandler2>();
```
jonsequitur, line 533: If the handler is not called, then this assertion won't be called. It's safer to capture a local from the closure and assert on its state at the end of the test. This is also more readable as people generally expect the assertion at the end of the test.

src/System.CommandLine/Invocation/CommandHandler.cs
```diff
@@ -282,6 +282,10 @@ public static class CommandHandler
             Func<T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12, T13, T14, T15, T16, Task<int>> action) =>
             HandlerDescriptor.FromDelegate(action).GetCommandHandler();
 
+        public static ICommandHandler FromBindingContext<THandler>()
```
jonsequitur, line 285: It might be helpful for the naming to reflect that the actual handler will be created during binding. One not great idea to open up discussion: `DuringBindingCreateHandlerOfType<T>`

src/System.CommandLine.Tests/Invocation/CommandHandlerTests.cs
```diff
@@ -507,5 +507,101 @@ public class OverridenVirtualTestCommandHandler : VirtualTestCommandHandler
...
         }
+
+        [Fact]
+        public static void FromBindingContext_forwards_invocation_to_bound_handler_type()
+        {
+            var command = new RootCommand
+            {
+                Handler = CommandHandler.FromBindingContext<BindingContextResolvedCommandHandler>()
+            };
+            command.Handler.Should().NotBeOfType<BindingContextResolvedCommandHandler>();
```
jonsequitur, line 518: This looks like it deserves to be a separate test to surface this detail.

src/System.CommandLine.Tests/Invocation/CommandHandlerTests.cs
```diff
@@ -507,5 +507,101 @@ public class OverridenVirtualTestCommandHandler : VirtualTestCommandHandler
...
+            {
+                handler.Should().BeOfType<BindingContextCommandHandler2>();
+            };
+            var parser = new CommandLineBuilder(new RootCommand
+            {
+                Handler = CommandHandler.FromBindingContext<IBindingContextCommandHandlerInterface>()
+            })
+                .ConfigureBindingContext(context => context.AddService(_ => action))
+                .ConfigureBindingContext(context => context.AddService<IBindingContextCommandHandlerInterface, BindingContextCommandHandler1>())
+                .ConfigureBindingContext(context => context.AddService<IBindingContextCommandHandlerInterface, BindingContextCommandHandler2>())
```
jonsequitur, line 541: One concern I have with this approach is that all handler registrations for all subcommands will happen regardless of which subcommand was used, when only one subcommand is ever used. This can be a bit wasteful and is part of why we've generally steered away from the common pattern for longer-lived apps where DI is configured up front but the cost is amortized over a longer lifetime. This is very often not the case for command-line apps.

src/System.CommandLine/Invocation/CommandHandler.cs
fredrikhr: `CreateLazyBoundHandler` (to match with the existing `Create`)?

src/System.CommandLine.Tests/Invocation/CommandHandlerTests.cs
fredrikhr: resolved in https://github.com/dotnet/command-line-api/pull/1358/commits/765d3a57f17a5eb4b2235e026d05a3ad3ce12609

src/System.CommandLine.Tests/Invocation/CommandHandlerTests.cs
fredrikhr: resolved in https://github.com/dotnet/command-line-api/pull/1358/commits/5db4df8aa57b46a17b3a60c57c0a92a847fad59b

src/System.CommandLine.Tests/Invocation/CommandHandlerTests.cs
fredrikhr: @jonsequitur In this case (using the lazily bound registration) on the registration is added, no instance of `BindingContextCommandHandler1` nor `BindingContextCommandHandler2` is actually created at this point. And neither does the call to `CommandHandler.FromBindingContext`. The actual instance is first created when the command handler for the choses command is invoked.

fredrikhr:
@jonsequitur 

> The new AddService overloads increase the scope of the DI features in ways that might be misleading and will likely lead to additional issues being opened when people misunderstand the goal.

Yeah, I can see that. Basically, all service registrations are technically Transient services (they all have a factory registration that is called each time the service is resolved), but of course the factory for the basic services returns singleton instances. However, the no-func overloads that I added really are Transient. Maybe add the `Transient` suffix? Would that help?

src/System.CommandLine.Tests/Invocation/CommandHandlerTests.cs
jonsequitur: Understood. At invocation time though, all of these registrations happen on the `BindingContext`, and coupled with the addition of non-lazy `AddService` calls, this encourages the pattern I mentioned above.

jonsequitur:
I'd prefer to avoid increasing the scope of the existing DI implementation at all, which originally lacked a public `AddService` method of any kind.

A safer approach might be to allow people to bring their own `IServiceProvider` implementation and therefore be able to completely control object lifetimes.

fredrikhr:
@jonsequitur Hmm, okay, so we could just remove all the overloads to `AddService`. So, what about `CommandHandler.CreateLazyFromModelBinder`? If I get a ModelBinder that would be able to create an instance regardless of whether it was added to the BindingContext, no? So we don't actually need to add Handlers as BindingContext services? (Provided paramless-ctor)

jonsequitur:
The way we create command handlers is undergoing some rethinking so that we can eventually remove all of the name-based matching and a lot of model binding complexity. If you're interested in the source generator approach we're looking at, take a look at this PR: #1362.

PR #1361, opened 2021-07-27
Fix 1351 and 1365

### LINKED ISSUE #1351: When option argument matches an option token, it's not parsed as an argument
This test fails:

```cs
[Fact]
public void Option_arguments_can_match_option_tokens()
{
        var option = new Option<string>("-x");

        var result = option.Parse("-x -x");

        result.ValueForOption(option).Should().Be("-x");
}
```

jonsequitur:
Fixes #1351 and #1365.

### DIFF
--- a/src/System.CommandLine.Tests/OptionTests.MultipleArgumentsPerToken.cs
+++ b/src/System.CommandLine.Tests/OptionTests.MultipleArgumentsPerToken.cs
@@ -98,9 +98,9 @@ namespace System.CommandLine.Tests
 
                     var result = command.Parse("--option 1 2");
 
-                    var value = result.ValueForOption(option);
+                    var optionResult = result.ValueForOption(option);
 
-                    value.Should().BeEquivalentTo(new[] { "1" });
+                    optionResult.Should().BeEquivalentTo(new[] { "1" });
                     result.UnmatchedTokens.Should().BeEquivalentTo(new[] { "2" });
                 }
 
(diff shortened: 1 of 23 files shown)


src/System.CommandLine.Tests/OptionTests.MultipleArgumentsPerToken.cs
```diff
@@ -86,6 +96,76 @@ public void When_option_is_not_respecified_and_limit_is_reached_then_the_followi
                         .Should()
                         .BeEquivalentTo("some-arg");
                 }
+
+                [Theory]
+                [InlineData("--option 1 --option 2")]
+                [InlineData("xyz --option 1 --option 2")]
+                [InlineData("--option 1 xyz --option 2")]
```
Keboo, line 103: Should there be one more option with xyz at the end?
```suggestion
                [InlineData("--option 1 xyz --option 2")]
                [InlineData("--option 1 --option 2 xyz")]
```

src/System.CommandLine/Parsing/StringExtensions.cs
```diff
@@ -330,6 +358,11 @@ bool TryUnbundle(out IReadOnlyList<string>? replacement)
                 }
             }
 
+            bool PreviousTokenIsAnOptionExpectingAnArgument() =>
+                tokenList.Count > 1 &&
+                tokenList[tokenList.Count - 1] is { Type: TokenType.Option } optToken &&
+                currentCommand!.Children.GetByAlias(optToken.Value) is Option { Arity: { MaximumNumberOfValues: > 0 } };
```
Keboo, line 364: Are you sure that `currentCommand` cannot be null here? From looking at the two invocations I am not sure that is always guaranteed. Perhaps consider putting the null propagation back on it, as returning false in the null case appears to be correct anyway.

Keboo (approved):

src/System.CommandLine/Parsing/StringExtensions.cs
jonsequitur: Good call.

Keboo (approved):

database64128:
This PR broke the parsing of certain commands:

```
Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
   at System.CommandLine.Parsing.StringExtensions.ValidTokens(ICommand command)
   at System.CommandLine.Parsing.StringExtensions.Tokenize(IReadOnlyList`1 args, CommandLineConfiguration configuration)
   at System.CommandLine.Parsing.Parser.Parse(IReadOnlyList`1 arguments, String rawInput)
   at System.CommandLine.CommandExtensions.GetInvocationPipeline(Command command, String[] args)
   at System.CommandLine.CommandExtensions.InvokeAsync(Command command, String[] args, IConsole console)
   at ShadowsocksUriGenerator.CLI.Program.<Main>(String[] args)
```

Options also won't accept more than 9 values. For example, passing `--values 1 2 3 4 5 6 7 8 9 10` would leave 10 out of the array.

PR #1362, opened 2021-07-28
Source generators for command handlers

Keboo:
This is a proof-of-concept just to solicit feedback. It is not a complete solution (yet?).

This leverages source generators to automatically build up command handlers. 
The most interesting part of the solution is the GeneratedCommandHandlerTests.cs. This contains the tests to show how this works. Currently this adds in a new Generator property on the CommandHandlers class. This class is meaningless on its own, and only used as an anchor point for the source generator to look for. There are a couple "Generate" extension methods on this class (it is key that they are extension methods). When the generator sees a call to those Generate methods its generates a new extension method with better parameter matching to ensure that it will always be the one resolved. This is how it is able to build up the handlers at compile time.

There are lots of known failure cases, lots of things that have not been handled, and the source generator code is a mess. 

Looking to know if this is an idea that is interesting enough to pursue and feedback on the approach.


KathleenDollard (commented):
Wow! Can we meet to go over this. I love the idea but got lost in so much code and would like to work back from the code we are aiming for. 

I'd like to ensure this plays nice with models as I think you'll like the most recent model I'm working on.

Keboo:
@KathleenDollard yes I would would love to get together to go over this. My next steps are going to be to handle some of the error cases, and replace the C# string building. 
Jon had suggested getting this decoupled from the main core lib so it could be released separately to let people try it out and report issues, so I will also work on getting that moved.

jonsequitur (approved):

PR #1367, opened 2021-08-02
Show ArgumentHelpName in help if explicitly set by user

sfoslund:


src/System.CommandLine/Help/HelpBuilder.cs
```diff
@@ -522,7 +522,7 @@ protected string GetArgumentDescriptor(IArgument argument)
 
             string descriptor;
             var suggestions = argument.GetSuggestions().ToArray();
-            if (suggestions.Length > 0)
+            if (suggestions.Length > 0 && string.Join("|", suggestions).Length < 50)
```
jonsequitur, line 525: Any thoughts on how to do this without the magic number 50? Should it be based on the terminal width? And is the calculated table width thrown off by this?

@keboo?

Keboo: My thought would be to not have it automatically turn off at a fixed threshold. It is hard to say what that threshold should be (even with accounting for terminal width). I can imagine some program wanting lots of suggestions output, while others do not. I think the question is what should be the right API for including/excluding them. 
Perhaps a better solution would be to expand the `HelpBuilderExtension.Customize` method on options/arguments to provide a means of controlling this?

src/System.CommandLine.Tests/Help/HelpBuilderTests.cs
```diff
@@ -469,6 +469,38 @@ public void Arguments_section_includes_configured_argument_aliases()
...
+        }
+
+        [Fact]
+        public void Arguments_section_uses_name_if_suggestions_are_long()
+        {
+            var command = new Command("the-command")
+            {
+                new Option<VerbosityOptions>(new[] { "-v", "--verbosity" })
+                {
+                    ArgumentHelpName = "LEVEL"
```
jonsequitur, line 494: I wonder if just setting the `ArgumentHelpName` should disable the use of suggestions in help output.

sfoslund: That makes sense to me.

sfoslund: @jonsequitur should I switch to implementing your suggestion instead?

jonsequitur: That would be great.

src/System.CommandLine/Help/HelpBuilder.cs
sfoslund: Switched to a different approach, see new commit.

src/System.CommandLine/Argument.cs
```diff
@@ -65,6 +66,16 @@ public Argument(string name)
             set => _arity = value;
         }
 
+
+        /// <summary>
+        /// Gets or sets if the ArgumentHelpName is specified by the user.
+        /// </summary>
+        public bool IsHelpNameSpecified
```
jonsequitur, line 73: This should probably be `internal`.

jonsequitur: Actually, this feels like an implementation concern of the `HelpBuilder` rather than something intrinsic to `Argument`. Maybe just move the check there e.g. `string.IsNullOrEmpty(argument.Name)`?

sfoslund: I can make it internal. Although I don't think we can just check `string.IsNullOrEmpty(argument.Name)` since the argument name gets a default value when not set explicitly: https://github.com/dotnet/command-line-api/blob/e0edf1801e2ba9b4b5f733e3484aa7dbe8bc8aca/src/System.CommandLine/Argument.cs#L124

jonsequitur: Sorry, I was thinking of `ArgumentHelpName` and realized that that's only present on `Option`. It might be useful to decouple this from `Argument.Name`.

sfoslund: How so, add a separate help name field on the argument class?

jonsequitur: Yeah, I think that's nice and simple. @Keboo?

Keboo: LGTM

sfoslund: Just pushed this change.

src/System.CommandLine/Help/HelpBuilder.cs
```diff
@@ -522,7 +522,12 @@ protected string GetArgumentDescriptor(IArgument argument)
 
             string descriptor;
             var suggestions = argument.GetSuggestions().ToArray();
-            if (suggestions.Length > 0)
+            var helpName = GetArgumentHelpName(argument);
+            if (helpName != null)
```
jonsequitur, line 526:
```suggestion
            if (!string.IsNullOrEmpty(helpName))
```

src/System.CommandLine/Help/HelpBuilder.cs
```diff
@@ -538,6 +543,12 @@ protected string GetArgumentDescriptor(IArgument argument)
             return descriptor;
         }
 
+        private string? GetArgumentHelpName(IArgument argument)
+        {
+            var arg = argument as Argument;
+            return arg == null ? null : arg.HelpName;
```
jonsequitur, line 549:
```suggestion
            return arg?.HelpName;
```

src/System.CommandLine/Option.cs
```diff
@@ -133,8 +133,11 @@ internal virtual Argument Argument
         /// </value>
         public string ArgumentHelpName
         {
-            get => Argument.Name;
-            set => Argument.Name = value;
+            get => Argument.HelpName ?? Argument.Name;
```
jonsequitur, line 136: I wonder if we should skip the fallback value and move the fallback logic to the `HelpBuilder`.

sfoslund: Yeah, that makes sense to me. I was going to make the type here `string?` and not have this default but that breaks this test: https://github.com/dotnet/command-line-api/blob/d6d49b068ffa199ca60a2aa4a2d31e748c4dbe73/src/System.CommandLine.Tests/OptionTests.cs#L239, are you okay with that behavior change?

jonsequitur: If that's the only test that breaks, that seems fine.

sfoslund: Sounds good, pushed that change along with other suggestions.

src/System.CommandLine.Tests/OptionTests.cs
```diff
@@ -240,7 +240,8 @@ public void Argument_takes_option_alias_as_its_name_when_it_is_not_provided()
         {
             var command = new Option("--alias", arity: ArgumentArity.ZeroOrOne);
 
-            command.ArgumentHelpName.Should().Be("alias");
+            command.Name.Should().Be("alias");
+            command.ArgumentHelpName.Should().Be(null);
```
jonsequitur, line 244: We can leave off the `ArgumentHelpName` assertion as it's unrelated to name of the test.

jonsequitur (approved):

PR #1411, opened 2021-09-16
rename source generator anchor to SetHandler

jonsequitur:
image
image


Keboo (approved):

ScottArbeit:
If I'm using System.CommandLine from F#, and it starts using Source Generators, which requires Roslyn... I think that's a problem? I don't invoke Roslyn as part of my compilation.

If this change precludes using System.CommandLine from F#, or somehow forks the syntax so there's one way to use it with C#, and another way to use it with F#, I do not approve this PR. 😜👎🏼

ScottArbeit:
Straight from Don Syme, in the F# Language issue [Support Source Generators](https://github.com/fsharp/fslang-suggestions/issues/864#issuecomment-903267638):

> For the foreseeable future that story will be "add a C# project and generate the source there".

So... I'd argue to postpone this change until and unless there's a way to make it work smoothly in F#.

Keboo:
@ScottArbeit to be clear this is merely one means of creating a command handler in this library. All of the previous APIs are still in place and will continue to function. This is simply being added as an experiment in leveraging source generators to make creating handlers in C# easier.

jonsequitur:
@ScottArbeit I had a conversation with @TIHan about this exact thing today and it's definitely something we'd like to solve. Currently the anchor method is an inviting dead end for non-C# users so clearly not ready for a non-beta release.

ScottArbeit:
@Keboo @jonsequitur thank you for the confirmation.

And, yeah, sigh. I don't want to get in the way of taking advantage of Source Generators, it's an awesome new feature in C# and worth exploring here. And I also shift uncomfortably in my seat bifurcating things between C# and F#, but I understand leaning towards C# usage.

Would an F#-focused documentation/usage page be useful? I'm already using it from F#, so I have code to draw examples from... that could be a home for calling out this difference.

jonsequitur:
We'd love to see some F# examples, and yeah, we'll have F# documentation.

PR #1423, opened 2021-09-27
Remove 0-value comment

JayBazuzi:


jonsequitur:
There are a lot more like this one... 

A quick check seems to indicate that this change won't emit a compiler warning.

JayBazuzi:
> compiler warning

And you have warn-as-error enabled, right?

Right?

jonsequitur:
Correct. But for this particular warning, it's suppressed for the moment while we fill in the missing doc comments.

PR #1428, opened 2021-09-28
Allow help descriptions to be based on parse result

sfoslund:


src/System.CommandLine/Help/HelpBuilder.cs
```diff
@@ -17,11 +17,13 @@ public class HelpBuilder : IHelpBuilder
         private Dictionary<ISymbol, Customization> Customizations { get; } = new();
 
         protected Resources Resources { get; }
+        protected ParseResult? ParseResult { get; }
         public int MaxWidth { get; }
 
-        public HelpBuilder(Resources resources, int maxWidth = int.MaxValue)
+        public HelpBuilder(Resources resources, int maxWidth = int.MaxValue, ParseResult? parseResult = null)
```
jonsequitur, line 23: It will make more sense to add the `ParseResult` parameter to the `IHelpBuilder.Write` method. Otherwise, this won't work with custom help builders.

src/System.CommandLine/ISymbol.cs
```diff
@@ -21,6 +22,11 @@ public interface ISymbol : ISuggestionSource
         /// </summary>
         string? Description { get; }
 
+        /// <summary>
+        /// Gets the symbol description.
+        /// </summary>
+        Func<ParseResult, string>? DescriptionCallback { get; }
```
jonsequitur, line 28: We shouldn't add this to the interface.

src/System.CommandLine/Symbol.cs
```diff
@@ -24,6 +24,12 @@ private protected Symbol()
         /// </summary>
         public string? Description { get; set; }
 
+
+        /// <summary>
+        /// Gets the symbol description.
+        /// </summary>
+        public Func<ParseResult, string>? DescriptionCallback { get; set; }
```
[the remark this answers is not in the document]
sfoslund: Sounds good, I've switched to using the customize methods in the latest implementation.

sfoslund:
@jonsequitur does this new approach look okay to you? If so, I'll resolve merge conflicts

src/System.CommandLine.Tests/Help/HelpBuilderTests.cs
```diff
@@ -1337,10 +1337,45 @@ public void Option_can_customize_descriptor()
...
+            helpBuilder.Write(parseResultB.CommandResult.Command, _console, parseResultB);
+            help = _console.ToString();
+            help.Should().Contain(optionBHelpText);
+        }
+
         [Fact]
         public void Null_option_on_customize_throws()
         {
-            Action action = () => _helpBuilder.Customize((IOption)null);
+            Action action = () => _helpBuilder.Customize((IOption)null, (string)null);
```
jonsequitur, line 1378: This can be an optional parameter and avoid some of these test changes.

src/System.CommandLine.Tests/Help/HelpBuilderTests.cs
```diff
@@ -1338,10 +1338,45 @@ public void Option_can_customize_descriptor()
...
+            var optionBHelpText = "option b help";
+            Func<ParseResult, string> descriptionCallback = (ParseResult parseResult) =>
+                parseResult.CommandResult.Command.Equals(commandA) ? optionAHelpText : optionBHelpText;
+
+            HelpBuilder helpBuilder = new HelpBuilder(Resources.Instance, LargeMaxWidth);
+            helpBuilder.Customize(option, descriptor: descriptionCallback);
+
+            var parseResultA = command.Parse("root a --option");
+            helpBuilder.Write(parseResultA.CommandResult.Command, _console, parseResultA);
+            var help = _console.ToString();
```
jonsequitur, line 1367: I would consider using two different console instances for the two different invocations since the content won't be cleared in between.

src/System.CommandLine/Help/HelpBuilder.cs
```diff
@@ -6,6 +6,9 @@
 using System.CommandLine.Parsing;
 using System.IO;
 using System.Linq;
+//using System.Runtime.CompilerServices;
+
+//[assembly: InternalsVisibleTo("System.CommandLine.Tests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100b5fc90e7027f67871e773a8fde8938c81dd402ba65b9201d60593e96c492651e889cc13f1415ebb53fac1131ae0bd333c5ee6021672d9718ea31a8aebd0da0072f25d87dba6fc90ffd598ed4da35e44c398c454307e8e33b8426143daec9f596836f97c8f74750e5975c64e2189f45def46b2a2b1247adc3652bf5c308055da9")]
```
jonsequitur, line 11: Please delete the commented out code.

src/System.CommandLine/Command.cs
```diff
@@ -25,6 +25,8 @@ public class Command :
     {
         private readonly SymbolSet _globalOptions = new();
 
+        public static readonly Command Empty = new Command("command");
```
jonsequitur, line 28: I would inline this. It's an implementation detail of `ParseResult.Empty` and so no change should be needed here.

src/System.CommandLine/Parsing/ParseResult.cs
```diff
@@ -18,6 +18,15 @@ public class ParseResult
         private readonly IReadOnlyList<Token> _unparsedTokens;
         private readonly IReadOnlyList<Token> _unmatchedTokens;
 
+        public static readonly ParseResult Empty = new ParseResult(
```
jonsequitur, line 21: For now let's keep this internal.

jonsequitur: Since this won't be needed unless help is requested, it's better to make this a method and return the instance only when needed rather than allocating it up front all the time.

src/System.CommandLine/Parsing/RootCommandResult.cs
```diff
@@ -11,6 +11,13 @@ internal class RootCommandResult : CommandResult
         private readonly Dictionary<ICommand, CommandResult> _allCommandResults;
         private readonly Dictionary<IOption, OptionResult> _allOptionResults;
 
+        internal new static readonly RootCommandResult Empty = new RootCommandResult(
```
jonsequitur, line 14: I would inline this. It's an implementation detail of `ParseResult.Empty` and so no change should be needed here.

src/System.CommandLine/Parsing/ParseResult.cs
```diff
@@ -18,6 +18,15 @@ public class ParseResult
...
+            new Parser(),
+            Parsing.RootCommandResult.Empty,
+            CommandResult.Empty,
+            new DirectiveCollection(),
+            new TokenizeResult(new List<Token>(), new List<TokenizeError>()),
+            new List<Token>(),
+            new List<Token>());
```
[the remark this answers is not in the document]
jonsequitur: Actually, the simplest thing to do might be to do this:

```cs
internal static ParseResult Empty() => new RootCommand().Parse(Array.Empty<string>());
```

src/System.CommandLine/Help/HelpBuilder.cs
```diff
@@ -35,8 +35,8 @@ public HelpBuilder(LocalizationResources localizationResources, int maxWidth = i
         /// The maximum width for which to format help output.
         /// </summary>
         public int MaxWidth { get; }
-
-        public virtual void Write(ICommand command, TextWriter writer)
+        
+        public virtual void Write(ICommand command, TextWriter writer, ParseResult parseResult)
```
jonsequitur, line 39: We should validate that this isn't null.

sfoslund:
@jonsequitur is this good to go in?

sfoslund:
Thanks Jon!

PR #1443, opened 2021-10-11
Reset private option list when a symbol is added without collision checks

### LINKED ISSUE #1437: Referencing the Command.Options property before adding a global option causes Command.Options to remain empty
I noticed some of our unit tests would fail when walking through them with a debugger. Specifically, a command's `Options` property would sometimes contain options added with `.AddGlobalOption(...)` but sometimes it wouldn't.

A coworker and I investigated further and concluded it had to do with how `SymbolSet` caches `_options` from its parent class `AliasedSet<ISymbol>.Items`

https://github.com/dotnet/command-line-api/blob/74b012dfa05aca5952f1f9200ba6e4348a52fee2/src/System.CommandLine/Collections/SymbolSet.cs#L130-L151

When `SymbolSet.Options` is referenced before adding the global option, `_options` caches the symbols as they were before they were added by `.AddGlobalOption(...)`. 

Here's a couple of xUnit tests that replicate the behavior.

```csharp
[Fact]
public void ReferencePropertyBeforeAdding_Fails()
{
    // this fails when you reference command.Options before adding a global option
    var command = new Command("test");
    var option = new Option<string>("--option");
    Assert.Empty(command.Options);
    command.AddGlobalOption(option);
    Assert.NotEmpty(command.Options);
}

[Fact]
public void ReferencePropertyAfterAdding()
{
    // this succeeds when you don't reference command.Options before adding a global option
    var command = new Command("test");
    var option = new Option<string>("--option");
    //Assert.Empty(command.Options);
    command.AddGlobalOption(option);
    Assert.NotEmpty(command.Options);
}
```

The `ReferencePropertyBeforeAdding_Fails` test always fails. The `ReferencePropertyAfterAdding` test always suceeds. The only difference is the `Assert.Empty(command.Options);` line, which executes the `SymbolSet.Options` property getter. Inspecting the non-public fields of `command` with the debugger shows that `command.Children.Items` contains `option` but `command.Children._options` does not.

image

Because global options are added through [`SymbolSet.AddWithoutAliasCollisionCheck`](https://github.com/dotnet/command-line-api/blob/b751c2e3cc5aeee1916be9892a2fee5ddbef69fb/src/System.CommandLine/Command.cs#L80), but that [implementation](https://github.com/dotnet/command-line-api/blob/74b012dfa05aca5952f1f9200ba6e4348a52fee2/src/System.CommandLine/Collections/SymbolSet.cs#L50) does not call `SymbolSet.ResetIndex(...)`, the `_options` field is never reset [like it is in `SymbolSet.Add(...)`](https://github.com/dotnet/command-line-api/blob/74b012dfa05aca5952f1f9200ba6e4348a52fee2/src/System.CommandLine/Collections/SymbolSet.cs#L24).

asasine:
Fixes #1437

### DIFF
--- a/src/System.CommandLine.Tests/CommandTests.cs
+++ b/src/System.CommandLine.Tests/CommandTests.cs
@@ -478,7 +478,7 @@ namespace System.CommandLine.Tests
 
         // https://github.com/dotnet/command-line-api/issues/1437
         [Fact]
-        public void AddGlobalOption_referencing_Options_before_AddGlobalOption()
+        public void When_Options_is_referenced_before_a_global_option_is_added_then_adding_a_global_option_updates_the_Options_collection()
         {
             var option = new Option<string>("-x");
             var command = new Command("mycommand");


asasine:
No problem, appreciate the speedy review :)

PR #1461, opened 2021-11-08
fix #1238

### LINKED ISSUE #1238: Arguments after double dash are not parsed as arguments
I'm upgrading a project that was using a 0.1 version of System.CommandLine to a 2.0 version, but found in 2.0 arguments after a double dash are not parsed as arguments. You even have tests for it that treat them as unparsed text, despite having [commented on a past issue](https://github.com/dotnet/command-line-api/issues/959#issuecomment-653192286):

> From https://unix.stackexchange.com/questions/11376/what-does-double-dash-mean
>> More precisely, a double dash (--) is used in most bash built-in commands and many other commands to signify the end of command options, after which only positional parameters are accepted.

Whether someone actually uses this or not we can't know, but it does create a possible compatibility issue. If a goal is unix-like behavior, the passing positional arguments after a double dash should at least be an option.

jonsequitur:
This fixes #1238. The old behavior can be opted into using `EnableLegacyDoubleDashBehavior`.

### DIFF
--- a/src/System.CommandLine.Hosting.Tests/HostingTests.cs
+++ b/src/System.CommandLine.Hosting.Tests/HostingTests.cs
@@ -110,6 +110,7 @@ namespace System.CommandLine.Hosting.Tests
                 {
                     Handler = CommandHandler.Create<IHost>(Execute),
                 })
+                .EnableLegacyDoubleDashBehavior()
                 .UseHost(host =>
                 {
                     var invocation = (InvocationContext)host.Properties[typeof(InvocationContext)];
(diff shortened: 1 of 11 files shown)


Keboo (approved):

sfoslund (approved):
Looks okay to me, the SDK will want to opt into `EnableLegacyDoubleDashBehavior` to preserve existing behavior then?

Keboo (approved):

jonsequitur:
@sfoslund Yes.

PR #1472, opened 2021-11-18
remove dependency on Microsoft.CSharp

jonsequitur:


Keboo (approved):

adamsitnik:
cc @jkotas

jonsequitur:
Related to this, we're moving the name-based binding and associated reflection code out to a new package so people can opt into it: #1488.

PR #1489, opened 2021-11-28
use .slnf instead of patch files for source build

jonsequitur:
Remove patch files and use .slnf to specify projects for source build.


sourcebuild.slnf
```diff
@@ -0,0 +1,12 @@
+{
+  "solution": {
+    "path": "System.CommandLine.sln",
+    "projects": [
+      "src\\System.CommandLine\\System.CommandLine.csproj",
+      "src\\System.CommandLine.DragonFruit\\System.CommandLine.DragonFruit.csproj",
+      "src\\System.CommandLine.DragonFruit\\System.CommandLine.NamingConventionBinder.csproj",
```
MichaelSimons, line 7:
```suggestion
      "src\\System.CommandLine.NamingConventionBinder\System.CommandLine.NamingConventionBinder.csproj",
```

src/System.CommandLine.Hosting/System.CommandLine.Hosting.csproj
```diff
@@ -11,6 +11,10 @@
     <DebugType>portable</DebugType>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <ExcludeFromSourceBuild>true</ExcludeFromSourceBuild>
```
MichaelSimons, line 15: Why not just exclude this project from the slnf?

eng/source-build-patches/0002-Use-package-version-properties-for-reference-version.patch
```diff
@@ -1,29 +0,0 @@
...
-     <UsingToolSymbolUploader>true</UsingToolSymbolUploader>
-     <PublishWindowsPdb>true</PublishWindowsPdb>
-   </PropertyGroup>
--  <!--   <PropertyGroup>
-+  <PropertyGroup>
-     <FluentAssertionsVersion>4.18.0</FluentAssertionsVersion>
-     <NewtonsoftJsonVersion>9.0.1</NewtonsoftJsonVersion>
-     <SystemNetHttpVersion>4.3.1</SystemNetHttpVersion>
--  </PropertyGroup>  -->
-+    <MicrosoftExtensionsHostingPackageVersion Condition="'$(MicrosoftExtensionsHostingPackageVersion)' == ''">2.2.0</MicrosoftExtensionsHostingPackageVersion>
```
MichaelSimons, line 25: Have you verified the removal of this patch and others aren't going to introduce prebuilts?  They were added for a reason.

src/System.CommandLine.Hosting/System.CommandLine.Hosting.csproj
jonsequitur: It is excluded from the .slnf. This is vestigial.

eng/source-build-patches/0002-Use-package-version-properties-for-reference-version.patch
jonsequitur: My mistake. My impression was that the .slnf approach was a replacement for the patch files.

To make this easier going forward, if patches are needed, I would prefer we add:

* a way to reproduce them easily so that contributors unfamiliar with the process don't get blocked or update them incorrectly, and
* an automated test that verifies that they're producing exactly what's expected.

I'm also not clear on the reasons why some packages are included and others aren't. 

@omajid, any insights here?

MichaelSimons: I went ahead and pulled this commit and performed a full source-build to verify no prebuilts were introduced.  Everything built cleanly although there was an api change affecting runtime.

The goal is to eliminate all source-build patches.  Any changes necessary for source-build should be part of the product.  We are also working to catch as many issues as possible in the source-build leg of the repo builds.  The last big remaining work is to enable prebuilt detection.  This was something we wanted to get in for 6.0 but couldn't fit it in so it will happen in the 7.0 timeframe.  Once this is complete the dev UX for source-build should be much better.  Thanks for your patience and thank you Jon for integrating the patches here and removed the unneeded ones.

PR #1491, opened 2021-11-30
Move command handler generator anchor into its own assembly

Keboo:
This separates out the generator's anchor into its own assembly so that we don't clutter up the main project with generator specific code.

### DIFF
--- a/.vsts-ci.yml
+++ b/.vsts-ci.yml
@@ -29,7 +29,7 @@ stages:
       enablePublishTestResults: true
       enablePublishBuildAssets: true
       enableTelemetry: true
-      enableSourceBuild: true
+      enableSourceBuild: false
       helixRepo: dotnet/command-line-api
       timeoutInMinutes: 180 # increase timeout since BAR publishing might wait a long time
       jobs:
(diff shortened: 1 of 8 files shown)


jonsequitur (approved):

jkotas:
Does this mean that you will get an extra tiny assembly in the output if one uses source generators? I think that it is unfortunate.

I think we should aim for the System.CommandLine + source generator to be the lean and mean path that we recommend for best performance. Adding a small extra assembly to enable source generators is working against that.

cc @adamsitnik

Keboo:
@jkotas that is correct, at least at present. The idea is we want to make sure this is in fact the right source generator before baking it into the main library. There are some other approaches that also leverage source generators in a similar fasion to make building out your Parser and handlers easier. Eventually we may land with something in the core library after these approaches have had a little time to be tested out.

.vsts-ci.yml
```diff
@@ -29,7 +29,7 @@ stages:
       enablePublishTestResults: true
       enablePublishBuildAssets: true
       enableTelemetry: true
-      enableSourceBuild: true
+      enableSourceBuild: false
```
MichaelSimons, line 32: Why was source-build disabled?  This has caused ripple effects - https://github.com/dotnet/installer/pull/12757#issuecomment-991068628

PR #1493, opened 2021-12-01
Help API adjustments

jonsequitur:
This removes `IHelpBuilder`. 

We've been pushing toward making help output easier to compose and reuse and since much of the functionality is on `HelpBuilder`, the interface was starting to seem superfluous.

An entirely custom implementation is still possible by inheriting `HelpBuilder` and overriting `Write`. Given this, there's no compelling reason to have both `HelpBuilder` and `IHelpBuilder`.


Keboo (approved):

sfoslund (approved):
The SDK already does not use IHelpBuilder so I expect this won't be much of  breaking change for us.

vlada-shubina (approved):

PR #1501, opened 2021-12-02
Perf improvements part 1

adamsitnik:
I've added a simple benchmark based on @jkotas sample (https://github.com/jkotas/commandline-perf/blob/master/command-line-api/Program.cs)

`String[0]` is test case when user has not passed any args
`String[4]` is for `--bool true -s test`

| Method |      Args | Mean before | Mean after |
|------- |---------- |------------:|-----------:|
|   Sync | String[0] |    88.04 us |   45.60 us |
|  Async | String[0] |    61.21 us |   46.05 us |
|   Sync | String[4] |    85.71 us |   50.65 us |
|  Async | String[4] |    66.33 us |   51.73 us |

Explanation:
* there is no need to start a new task, we can call `invocationChain` (it returns a task) and wait for it directly
* there is no need to check if directory exists if we know that the file that belongs to this directory exists

@jonsequitur PTAL

### DIFF
--- a/src/System.CommandLine.Benchmarks/CommandLine/Perf_Parser_Simple.cs
+++ b/src/System.CommandLine.Benchmarks/CommandLine/Perf_Parser_Simple.cs
@@ -1,0 +1,35 @@
+﻿using BenchmarkDotNet.Attributes;
+using System.CommandLine.Invocation;
+using System.Threading.Tasks;
+
+namespace System.CommandLine.Benchmarks.CommandLine
+{
+    [BenchmarkCategory(Categories.CommandLine)]
+    public class Perf_Parser_Simple
+    {
+        [Params(new string[0], new string[4] { "--bool", "true", "-s", "test" })]
+        public string[] Args { get; set; }
+
+        [Benchmark]
+        public int Sync() => BuildCommand().Invoke(Args);
+
+        [Benchmark]
+        public Task<int> Async() => BuildCommand().InvokeAsync(Args);
+
+        private static RootCommand BuildCommand()
+        {
+            Option<bool> boolOption = new Option<bool>(new[] { "--bool", "-b" }, "Bool option");
+            Option<string> stringOption = new Option<string>(new[] { "--string", "-s" }, "String option");
+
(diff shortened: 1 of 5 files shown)


jkotas:
This is good change. It would be also interesting to measure the total time to execute the program w/ the command line parser (ie include JITing and type loading). It is where the most severe System.CommandLine perf issues has been.

src/System.CommandLine.Benchmarks/CommandLine/Perf_Parser_Simple.cs
```diff
@@ -0,0 +1,35 @@
...
+    public class Perf_Parser_Simple
+    {
+        [Params(new string[0], new string[4] { "--bool", "true", "-s", "test" })]
+        public string[] Args { get; set; }
+
+        [Benchmark]
+        public int Sync() => BuildCommand().Invoke(Args);
+
+        [Benchmark]
+        public Task<int> Async() => BuildCommand().InvokeAsync(Args);
```
jonsequitur, line 17: A significant chunk of code that could be optimized is inside of the `Invoke` and `InvokeAsync` extension methods on `Command`. These use the default middleware pipeline. This is all opt-in. It represents many of System.CommandLine's features that are not found in many other command line libraries. An apples-to-apples performance comparison to other libraries would omit these until/unless we optimize the middleware pipeline. 

Here's what it's doing:

https://github.com/dotnet/command-line-api/blob/edee3cf6c15e9d3430d8cc4eec10b4928fcf63ce/src/System.CommandLine/Builder/CommandLineBuilderExtensions.cs#L345-L359

Leaving this out (or optimizing it) will knock about 50% off of the allocations. The pipeline is composed together here

https://github.com/dotnet/command-line-api/blob/843fdc7958ec919c70462635b18ac2152a907f67/src/System.CommandLine/Invocation/InvocationPipeline.cs#L42-L68

jonsequitur (approved):

jonsequitur:
@adamsitnik Please also feel free to prune the other benchmarks.  They're maybe more granular than necessary.

PR #1502, opened 2021-12-02
Adding in cancellation support for InvokeAsync

### LINKED ISSUE #1469: add `CancellationToken` parameter to command handler `InvokeAsync` methods

Keboo:
Fixes #1469 

There is a slight behavior change in this. Previously, if you did not request a `CancellationToken` in a command handler, it would not register up on the `Console.CancelKeyPress` and `AppDomain.CurrentDomain.ProcessExit` events.
With these changes it now always registers for those events if you call `CommandLineBuilder.CancelOnProcessTermination` (which is part of the defaults; so likely often).

Key areas to review are `CommandLineBuilderExtensions` and `InvocationContext` changes.

### DIFF
--- a/src/System.CommandLine.ApiCompatibility.Tests/ApiCompatibilityApprovalTests.System_CommandLine_api_is_not_changed.approved.txt
+++ b/src/System.CommandLine.ApiCompatibility.Tests/ApiCompatibilityApprovalTests.System_CommandLine_api_is_not_changed.approved.txt
@@ -66,8 +66,8 @@ System.CommandLine
   public static class CommandExtensions
     public static System.Int32 Invoke(this Command command, System.String[] args, IConsole console = null)
     public static System.Int32 Invoke(this Command command, System.String commandLine, IConsole console = null)
-    public static System.Threading.Tasks.Task<System.Int32> InvokeAsync(this Command command, System.String[] args, IConsole console = null)
-    public static System.Threading.Tasks.Task<System.Int32> InvokeAsync(this Command command, System.String commandLine, IConsole console = null)
+    public static System.Threading.Tasks.Task<System.Int32> InvokeAsync(this Command command, System.String[] args, IConsole console = null, System.Threading.CancellationToken cancellationToken = null)
+    public static System.Threading.Tasks.Task<System.Int32> InvokeAsync(this Command command, System.String commandLine, IConsole console = null, System.Threading.CancellationToken cancellationToken = null)
     public static ParseResult Parse(this Command command, System.String[] args)
     public static ParseResult Parse(this Command command, System.String commandLine)
   public class CommandLineBuilder
@@ -357,8 +357,8 @@ System.CommandLine.Help
 System.CommandLine.Invocation
   public interface IInvocationResult
     public System.Void Apply(InvocationContext context)
-  public class InvocationContext
-    .ctor(System.CommandLine.ParseResult parseResult, System.CommandLine.IConsole console = null)
+  public class InvocationContext, System.IDisposable
+    .ctor(System.CommandLine.ParseResult parseResult, System.CommandLine.IConsole console = null, System.Threading.CancellationToken cancellationToken = null)
     public System.CommandLine.Binding.BindingContext BindingContext { get; }
     public System.CommandLine.IConsole Console { get; set; }
     public System.Int32 ExitCode { get; set; }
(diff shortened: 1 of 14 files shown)


src/System.CommandLine/Invocation/InvocationContext.cs
```diff
@@ -73,39 +81,40 @@ public ParseResult ParseResult
...
+            _linkedTokensSource?.Dispose();
+            _linkedTokensSource = null;
             (Console as IDisposable)?.Dispose();
         }
+
+        public void AddLinkedCancellationToken(CancellationToken token)
+        {
+            if (_lazyCancellationToken.IsValueCreated)
+            {
+                throw new InvalidOperationException($"Cannot add additional linked cancellation tokens once {nameof(InvocationContext)}.{nameof(CancellationToken)} has been invoked");
```
jonsequitur, line 115:
```suggestion
                throw new InvalidOperationException($"Cannot add additional linked cancellation tokens once {nameof(InvocationContext)}.{nameof(CancellationToken)} has been cancelled");
```

Keboo: I don't think this is correct. The failure occurs if the Lazy has already been evaluated, meaning something requested the `CancellationToken` from the context. It may or may not be cancelled at this point.

tmds:
> Previously, if you did not request a CancellationToken in a command handler, it would not register up on the Console.CancelKeyPress and AppDomain.CurrentDomain.ProcessExit events. With these changes it now always registers for those events if you call CommandLineBuilder.CancelOnProcessTermination (which is part of the defaults; so likely often).

I haven't looked at the code yet, but it's important to know this:

Requesting the `CancellationToken` means you 'promise' to respect that token for terminating the application.
If it unconditionally enabled then applications that are not actually using the token will no longer terminate on SIGTERM or Ctrl+C.

src/System.CommandLine/Invocation/InvocationContext.cs
jonsequitur: This initially caught my eye  because I wasn't sure what it means to "invoke" a `CancellationToken`. If I saw this exception message, I'm not sure I'd understand what I should do differently. What can we do to make this error clearly actionable for developers?

tmds:
> Requesting the CancellationToken means you 'promise' to use that token for terminating the application.
If it is unconditionally enabled then applications that are not actually using the token will no longer terminate on SIGTERM or Ctrl+C.

@Keboo is this handled somehow?

Keboo:
Hi @tmds sorry for the long delay. I just got back to this and addressed the merge issues. After digging a bit more I did refactor the code so that it will not register for `Console.CancelKeyPress` or `AppDomain.CurrentDomain.ProcessExit` unless the command's handler requests a CancellationToken. I believe this addresses your concern above. Thank you very much for your feedback.

sakno:
I would like to propose a different approach to maintain linked token sources within `InvocationContext`:
```csharp
public sealed class InvocationContext : IDisposable
{
    private readonly CancellationToken _token; // cached token to avoid ObjectDisposedException due to concurrency between Dispose() and Cancel()
    private readonly LinkedList<CancellationTokenRegistration> _registrations = new();
    private volatile CancellationTokenSource? _source;
    private volatile int _canceled;

    public InvocationContext(...)
    {
        _source = new();
        _token = _source.Token;
    }

    public CancellationToken GetCancellationToken() => _token;

    // internal visible for ease access from program termination callbacks
    internal void Cancel()
    {
       using (var source = Interlocked.Exchange(ref _source, null))
       {
           source.Cancel();
       }
    }

    public void LinkToken(CancellationToken token)
    {
        _registrations.AddLast(token.Register(this.Cancel));
    }

    // explicit to prevent invocation from middleware or any other parts of user code
    void IDisposable.Dispose()
    {
        Interlocked.Exchange(ref _source, null)?.Dispose();
        foreach (var registration in _registrations)
           registration.Dispose();
    }
}
```

src/System.CommandLine.ApiCompatibility.Tests/ApiCompatibilityApprovalTests.System_CommandLine_api_is_not_changed.approved.txt
```diff
@@ -357,8 +357,8 @@ System.CommandLine.Help
 System.CommandLine.Invocation
   public interface IInvocationResult
     public System.Void Apply(InvocationContext context)
-  public class InvocationContext
-    .ctor(System.CommandLine.ParseResult parseResult, System.CommandLine.IConsole console = null)
+  public class InvocationContext, System.IDisposable
```
Keboo, line 360: @jonsequitur not sure if this is important to you or not, but explicit interface members are not shown.

jonsequitur (approved):

Issue #1402, opened 2021-09-09
Consider removing ICommand, IOption, and IArgument


jonsequitur:
These interfaces were originally introduced when the parser components were made mutable. They were intended provide a "read-only" API for parse results that would discourage modifying an already-configured `Parser`, because tracking state changes is additional overhead and complexity. But additional implementations of these interfaces aren't intended to be supported and might not work as expected. Removing them could simplify some code. 

We would want to address issues that might come up from e.g. adding `Option` aliases after the construction of a `Parser`.

adamsitnik:
Can this issue be closed now since #1538 got merged?

PR #1510, opened 2021-12-10
Perf improvements part 3

adamsitnik:
@jonsequitur I am about to start my weekend now, so I am just opening the PR and providing benchmark results. On Monday I am going to describe why I've changed all of these things.

|        Method |      Args | Mean before | Mean after | Allocated before | Allocated  after|
|-------------- |---------- |------------:|-----------:|-----------------:|----------------:|
|  DefaultsSync | String[0] |   38.892 us |  30.004 us |         25,611 B |        14,010 B |
| DefaultsAsync | String[0] |   38.594 us |  29.371 us |         25,611 B |        14,010 B |
|   MinimalSync | String[0] |    6.905 us |   3.366 us |         12,201 B |         6,753 B |
|  MinimalAsync | String[0] |    7.084 us |   3.453 us |         12,201 B |         6,753 B |
|  DefaultsSync | String[4] |   44.543 us |  33.569 us |         30,016 B |        17,432 B |
| DefaultsAsync | String[4] |   45.152 us |  33.727 us |         30,016 B |        17,432 B |
|   MinimalSync | String[4] |   10.513 us |   5.739 us |         16,608 B |        10,176 B |
|  MinimalAsync | String[4] |   10.398 us |   5.888 us |         16,608 B |        10,176 B |


jonsequitur:
This is a great improvement! Some of these changes were already on my to-do list so thank you for doing this.

One change that I'm wondering about is the addition of the `IMatchable` interface to support name-based matching. I'd like to investigate an alternate approach.

We've recently moved the name-based binding out into a separate package (System.CommandLine.NamingConventionBinder). The name matching functionality is partly there to support that but also to support validation of the parser, e.g. detect naming collisions that would make a command line unparseable. 

With that in mind, can you think of an approach that avoids introducing an interface?

On a possibly related note, we've also discussed [removing `ICommand`, `IArgument`, and `IOption`](https://github.com/dotnet/command-line-api/issues/1402). They were introduced when we made the parser mutable in order to avoid writing code to gently discourage mutating symbols once the parser is constructed. I don't believe they add value, though, and we don't expect people to implement them.

jonsequitur (approved):

adamsitnik:
> can you think of an approach that avoids introducing an interface?

Sure! I'll see what I can do and get back to you.

> On a possibly related note, we've also discussed removing ICommand, IArgument, and IOption

I am in favor of removing every abstraction that is implemented by a single type and|or is never implemented by the users. 👍

src/System.CommandLine/Command.cs
```diff
@@ -170,6 +117,56 @@ private protected override void AddSymbol(Symbol symbol)
...
+                            {
+                                throw new ArgumentException($"Alias '{alias}' is already in use.");
+                            }
+                        }
+
+                        if (identifierSymbol is Command command)
+                        {
+                            if (ReferenceEquals(this, command))
+                            {
+                                throw new ArgumentException("Parent can't be it's own child.");
```
jonsequitur, line 145: I would suggest a different exception type.

Also, minor correction: 

```suggestion
                                throw new ArgumentException("Parent can't be its own child.");
```

src/System.CommandLine/Command.cs
```diff
@@ -170,6 +117,56 @@ private protected override void AddSymbol(Symbol symbol)
         /// </summary>
         public IEnumerator<Symbol> GetEnumerator() => Children.OfType<Symbol>().GetEnumerator();
 
+        /// <summary>
+        /// Validates all symbols including the child hierarchy.
+        /// </summary>
+        /// <remarks>Due to the performance impact of this method, it's recommended to create
+        /// a Unit Test that calls this method to verify the RootCommand of every application.</remarks>
+        public void Validate()
```
jonsequitur, line 125: Let's discuss naming to make it clearer this is not about input validation.

src/System.CommandLine/Option.cs
```diff
@@ -170,11 +171,17 @@ public virtual IArgumentArity Arity
...
+
                 Argument.Arity = value;
             }
         }
 
+        /// <summary>
+        /// Global options are applied to the command and recursively to subcommands.
+        /// They do not apply to parent commands.
+        /// </summary>
+        public bool IsGlobal { get; internal set; }
```
jonsequitur, line 183: Can we make this internal?

src/System.CommandLine/Symbol.cs
```diff
@@ -37,25 +36,39 @@ public virtual string Name
         private protected abstract string DefaultName { get; }
 
         /// <summary>
-        /// Represents the parent symbols.
+        /// Represents the first parent node.
         /// </summary>
-        public ISymbolSet Parents => _parents;
+        internal ParentNode? FirstParent => _firstParent;
+
+        public virtual bool Matches(string name) => string.Equals(name, _name, StringComparison.Ordinal);
```
jonsequitur, line 43: Can this be removed or made internal?

src/System.CommandLine/Option.cs
lennybacon: What was the reason for this change? Does this improve performance and if yes how/why? There is no way (from outside) to differentiate between global and local options anymore.

adamsitnik: @lennybacon they were duplicated (added to two collections). This commit includes all the changes related to it: https://github.com/dotnet/command-line-api/pull/1510/commits/d38bae9846bdf87039cb0b846a55ff15685510d4

> There is no way (from outside) to differentiate between global and local options anymore.

Ech option exposed by `.Options` is not providing `IsGlobal` property

lennybacon: @adamsitnik I got that. 

> Ech option exposed by .Options is not providing IsGlobal property

Not publicly, but the property is still there - it is internal.

I used this and now can only use reflection to access this information.

**My question was to understand weather a property being public vs. internal improves performance.**

And if performance was not the reason for the change... what was ist?

adamsitnik: @lennybacon please excuse me, I missed the fact that the getter is not public anymore.

> My question was to understand weather a property being public vs. internal improves performance.

it does not affect performance

> And if performance was not the reason for the change... what was ist?

most likely encapsulation, @jonsequitur do you remember why we have made `IsGlobal` an `internal` property?

lennybacon: Thank you @adamsitnik 

To explain what I was doing: I created an *interactive* global option. That when provided reads the command line and jumps in at any stage to allow interactive selection of command/argument/option which results in a newly created command line that is then displayed and in the end also executed. I really enjoyed the *explorable* style of System.CommandLine and would love to continue using it without dirty hacks like reflection.

lennybacon: @jonsequitur would you like me to create an issue (feature request) to make `IsGlobal` public again?

jonsequitur: > @jonsequitur would you like me to create an issue (feature request) to make IsGlobal public again?

That would be great. I'd recommend a more problem-oriented approach though. What's the thing you're trying to accomplish? Given this bug, making `IsGlobal` public might not be the only way to accomplish it, and we can use the issue to discuss approaches.

lennybacon: I created an issue here: https://github.com/dotnet/command-line-api/issues/1736

PR #1524, opened 2021-12-14
change CommandHandler.Create to Command.SetHandler extension method

jonsequitur:


src/System.CommandLine/Invocation/CommandHandler.Func.cs
```diff
@@ -14,384 +14,401 @@ public static partial class CommandHandler
...
         /// <summary>
         /// Creates a command handler based on a <see cref="Func{T}"/>.
         /// </summary>
-        public static ICommandHandler Create<T>(
+        public static ICommandHandler SetHandler<T>(
+            this Command command,
             Func<T, Task> handle,
             params IValueDescriptor[] symbols) =>
-            new AnonymousCommandHandler(
-                async context =>
```
baronfel, line 28: all of these overloads lost their `async` syntax here, how have the init semantics of these members changed as a result?  to me this looks like there would be cleaner errors during binding as a result, because those are happening outside of an async state machine wrapper, but I wanted to raise that explicitly here.

jonsequitur: Making them async had introduced a bug where the return values were getting lost. I added verification of the fix to the `Theory` that tests all of these methods.

I don't think there should be an interesting semantic change? Everything this code is doing was already synchronous. It's really only the user-defined handler that might be asynchronous.

PR #1534, opened 2021-12-16
don't throw on unexpected handler return types

jonsequitur:


src/System.CommandLine/Invocation/AnonymousCommandHandler.cs
```diff
@@ -48,11 +48,9 @@ public async Task<int> InvokeAsync(InvocationContext context)
                 case int exitCode:
                     ret = exitCode;
                     break;
-                case null:
+                default:
```
baronfel, line 51: nit: This InvokeAsync uses ret/break, but the one in the Naming Handler uses straight returns. Do you prefer one way or the other?

baronfel (approved):

src/System.CommandLine/Invocation/AnonymousCommandHandler.cs
jonsequitur: Not particularly. I think this one is more constrained because there's no way to get any old `Delegate` in here. The one in `NamingConventionHandler` will accept basically anything. I couldn't repro the bug here but relaxing this seemed like the right thing to do.

PR #1548, opened 2021-12-23
Perf improvements part 5

adamsitnik:
@jonsequitur this time the perf gains are smaller (mostly allocations), I think I've reached the point where the only part where I can find some potential perf bottlenecks is argument conversion (not included in this PR, I've created another branch with large experiment).

|        Method |      Args | Mean before | Mean after | Allocated before | Allocated  after|
|-------------- |---------- |------------:|-----------:|-----------------:|----------------:|
|  DefaultsSync | String[0] |   27.654 us |  27.682 us |         11,850 B |        11,521 B |
| DefaultsAsync | String[0] |   27.608 us |  27.775 us |         11,849 B |        11,521 B |
|   MinimalSync | String[0] |    2.646 us |   2.592 us |          4,961 B |         4,665 B |
|  MinimalAsync | String[0] |    2.766 us |   2.719 us |          4,961 B |         4,664 B |
|  DefaultsSync | String[4] |   31.348 us |  30.990 us |         15,152 B |        14,624 B |
| DefaultsAsync | String[4] |   31.433 us |  31.284 us |         15,152 B |        14,624 B |
|   MinimalSync | String[4] |    4.805 us |   4.606 us |          8,264 B |         7,768 B |
|  MinimalAsync | String[4] |    4.981 us |   4.628 us |          8,264 B |         7,768 B |


jonsequitur:
re: argument conversion, there's some reflection code left there that we'd ideally like to move out of the core library to remove trimming warnings. Right now, any class with a constructor taking a single string (found via reflection) can be bound. We can move this open-ended implementation to `NamingConventionBinder` and keep support for a few key types if needed by adding converters for them.

src/System.CommandLine/Command.cs
```diff
@@ -33,17 +36,35 @@ public Command(string name, string? description = null) : base(name, description
...
         /// <summary>
         /// Represents all of the options for the command, including global options.
         /// </summary>
-        public IReadOnlyList<Option> Options => Children.Options;
+        public IReadOnlyList<Option> Options => _options is not null ? _options : Array.Empty<Option>();
+
+        /// <summary>
+        /// Represents all of the subcommands for the command.
+        /// </summary>
+        public IReadOnlyList<Command> SubCommands => _subCommands is not null ? _subCommands : Array.Empty<Command>();
```
jonsequitur, line 67:
```suggestion
        public IReadOnlyList<Command> Subcommands => _subcommands is not null ? _subcommands : Array.Empty<Command>();
```

src/System.CommandLine/Command.cs
```diff
@@ -129,59 +166,49 @@ public void AddGlobalOption(Option option)
...
-            symbol.AddParent(this);
-        }
-
         /// <inheritdoc />
         public override IEnumerable<CompletionItem> GetCompletions(CompletionContext context)
         {
-            if (Children.Count == 0)
+            List<CompletionItem> completions = new();
+
+            foreach (Command command in SubCommands)
```
jonsequitur, line 174: I made some decent performance improvements earlier this year by using `for` instead of `foreach`. Should we be avoiding `IEnumerable<T>` and `foreach` when `IReadOnlyList<T>` is available?

src/System.CommandLine/Command.cs
```diff
@@ -129,59 +166,49 @@ public void AddGlobalOption(Option option)
...
 
             return completions
                    .OrderBy(item => item.SortText.IndexOfCaseInsensitive(context.WordToComplete))
                    .ThenBy(symbol => symbol.Label, StringComparer.OrdinalIgnoreCase);
+
+            static void AddCompletions(string wordToComplete, IdentifierSymbol symbol, List<CompletionItem> completions)
+            {
+                if (!symbol.IsHidden)
+                {
+                    foreach (string alias in symbol.Aliases)
```
jonsequitur, line 203: `Aliases` is currently backed by a `HashSet<string>` which was to avoid collisions but if we're moving this sort of validation to testing, this might be a good candidate for a `List<string>`, faster iteration, and no up-front hash set cost.

src/System.CommandLine.NamingConventionBinder/CommandResultExtensions.cs
```diff
@@ -46,21 +46,20 @@ internal static class CommandResultExtensions
 
         for (var i = 0; i < options.Count; i++)
         {
-            if (options[i] is Option option)
+            Option option = options[i];
```
jonsequitur, line 49: The `TryGetValueForOption` method looks like it isn't used by `System.CommandLine` at this point, only via  `System.CommandLine.NamingConventionBinder`, if you wanted to move it.

adamsitnik:
Closing in favor of https://github.com/dotnet/command-line-api/pull/1601

PR #1558, opened 2022-01-04
make error on missing symbol in SetHandler more helpful

jonsequitur:
This change improves the error message that occurs when a `SetHandler`-created command handler can't bind a parameter. Previously, the error assumed there was a missing registration in the service provider, but in most cases this happens because a symbol wasn't passed to bind to. 

The new error message suggests symbols that can be bound to the parameter.

### DIFF
--- a/src/System.CommandLine.Tests/Binding/SetHandlerTests.cs
+++ b/src/System.CommandLine.Tests/Binding/SetHandlerTests.cs
@@ -9,6 +9,7 @@ using System.Linq;
 using System.Threading.Tasks;
 using FluentAssertions;
 using Xunit;
+using static System.Environment;
 
 namespace System.CommandLine.Tests.Binding
 {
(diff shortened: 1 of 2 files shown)


src/System.CommandLine.Tests/Binding/SetHandlerTests.cs
```diff
@@ -160,6 +161,62 @@ public void If_service_is_not_found_then_an_error_results()
...
+
+            subcommand.SetHandler((bool boolValue, string stringValue) => { });
+
+            var console = new TestConsole();
+
+            command.Invoke("TheCommand -o hi", console);
+
+            console.Error.ToString().Should()
+                   .Contain(
+                       $"No binding target was provided to the handler for command 'TheCommand' for the parameter at position 0. Did you mean to pass one of these?{NewLine}Option<Boolean> -o");
```
KathleenDollard, line 189: Brilliant that we are offering a suggestion. And it looks like pretty cool magic making that happen. But a few things:

* There are two options/args available and only one is listed in the message. Can we do both.
* I'd like different language. Can we do: 

$"SetHandler for command 'TheCommand' is missing options or arguments. Did you mean to pass one of these?{NewLine}Option<Boolean> -o");

src/System.CommandLine.Tests/Binding/SetHandlerTests.cs
```diff
@@ -160,6 +161,62 @@ public void If_service_is_not_found_then_an_error_results()
...
+
+            subcommand.SetHandler((bool boolValue, string stringValue) => { }, boolOption);
+
+            var console = new TestConsole();
+
+            command.Invoke("TheCommand -o hi", console);
+
+            console.Error.ToString().Should()
+                   .Contain(
+                       $"No binding target was provided to the handler for command 'TheCommand' for the parameter at position 1. Did you mean to pass one of these?{NewLine}Argument<String> value");
```
KathleenDollard, line 217: I'm missing the difference between this and the test above

src/System.CommandLine/Handler.cs
```diff
@@ -34,6 +37,33 @@ public static partial class Handler
...
+                        {
+                            Argument<T> argument => $"{nameof(Argument)}<{argument.ValueType.Name}> {argument.Name}",
+                            Argument argument => $"{nameof(Argument)} {argument.Name}",
+                            Option<T> option => $"{nameof(Option)}<{option.ValueType.Name}> {option.Aliases.First()}",
+                            Option option => $"{nameof(Option)} {option.Aliases.First()}",
+                            _ => throw new ArgumentOutOfRangeException(nameof(c))
+                        }));
+
+                throw new ArgumentException(
+                    $"No binding target was provided to the handler for command '{context.ParseResult.CommandResult.Command.Name}' for the parameter at position {index}. Did you mean to pass one of these?{NewLine}{candidatesDescription}");
```
KathleenDollard, line 64:
```suggestion
                   var candidateString = String.Join({NewLine}, candidates;
                    $"The SetHandler() method for command '{context.ParseResult.CommandResult.Command.Name}' is missing an argument or option, which was expected to be of type 'T.ToString()'. Did you mean to pass one of these?{NewLine}{candidatesDescription}");
```

That is air code.

KathleenDollard (commented):
Brilliant! I have some suggestions to further refine. 

Also, if candidates are empty, is there a reasonable message if they are expected?

src/System.CommandLine.Tests/Binding/SetHandlerTests.cs
jonsequitur: > There are two options/args available and only one is listed in the message. Can we do both.

They're filtered by type.

src/System.CommandLine.Tests/Binding/SetHandlerTests.cs
jonsequitur: They're specifying an issue for a different parameter.

PR #1568, opened 2022-01-07
allow for injecting host services with sethandler

brporter:


src/System.CommandLine.Hosting/HostingExtensions.cs
```diff
@@ -56,6 +56,14 @@ public static class HostingExtensions
             Action<IHostBuilder> configureHost = null
             ) => UseHost(builder, null, configureHost);
 
+        public static CommandLineBuilder UseHostServices(this CommandLineBuilder builder)
+            => builder.AddMiddleware(async (invocation, next) => {
+                var host = invocation.GetHost();
+                invocation.BindingContext.AddServiceProvider(host?.Services);
```
WeihanLi, line 62: Maybe it should not be registered when the `host` is `null`

jonsequitur: Should host even be allowed to be null? That seems like it should be a disallowed state.

PR #1591, opened 2022-01-19
Fix threading issue with StandardStreamWriter.WriteLine

### LINKED ISSUE #1590: IConsole is not thread safe
I just found that the IConsole (and related TextWriter wrappers) are not threadsafe:

A console tool (net6.0) that runs multiple operations in parallel using Task.Run(..) and write progress using IConsole.WriteLine

>result from process1\r\n
>result from process2\r\n
>result from process3result from process4\r\n
>\r\n
>result from process5\r\n

Initially, I thought it was an issue with System.Console (see https://github.com/dotnet/runtime/issues/53751).

After trying to repro this, I found this in System.Commandline StandardStreamWriter: https://github.com/dotnet/command-line-api/blob/main/src/System.CommandLine/IO/StandardStreamWriter.cs#L59

System.Console uses a Synchronised TextWriter to avoid this. System.Commandline might need to do the same. Or do one call to "Write(message + Env.Newline)" instead of two separate calls on the Console.

remcoros:
Fixes #1590 by doing only one Write call to IStandardStreamWriter

### DIFF
--- a/src/System.CommandLine/IO/StandardStreamWriter.cs
+++ b/src/System.CommandLine/IO/StandardStreamWriter.cs
@@ -63,8 +63,7 @@ namespace System.CommandLine.IO
                 throw new ArgumentNullException(nameof(writer));
             }
 
-            writer.Write(value);
-            writer.Write(Environment.NewLine);
+            writer.Write(value + Environment.NewLine);
         }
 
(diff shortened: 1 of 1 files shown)


jonsequitur (approved):

jonsequitur:
Thanks @remcoros!

PR #1606, opened 2022-02-01
remove warnings for single file publishing

jonsequitur:


src/System.CommandLine/RootCommand.cs
```diff
@@ -23,7 +23,17 @@ public class RootCommand : Command
                 LazyThreadSafetyMode.PublicationOnly);
 
         private static readonly Lazy<string> _executablePath =
-            new(() => GetAssembly().Location,
+            new(() =>
+                {
+#if NET6_0_OR_GREATER
+                    var al = Environment.ProcessPath ?? "";
```
eerhardt, line 29: Is it OK for this to return `C:\Program Files\dotnet\dotnet.exe` when someone uses `dotnet.exe MyApp.dll` to invoke their app?

eerhardt, line 29: (nit) Is `al` the best variable name here?

src/System.CommandLine/RootCommand.cs
```diff
@@ -23,7 +23,17 @@ public class RootCommand : Command
...
+#else
+#pragma warning disable IL3000 // Avoid accessing Assembly file path when publishing as a single file
+                    var al = GetAssembly().Location ?? "";
```
Keboo, line 32: Isn't calling `GetAssembly()` here recursive? Anything I am missing?

src/System.CommandLine/RootCommand.cs
eerhardt: One option is to use Assembly.Location, and check it for `IsNullOrEmpty`. If it is empty, then assume you are in a single-file executable and fallback to ProcessPath.

For example:

https://github.com/dotnet/runtime/blob/8447c80431f091c1eb51a1468dce7638d13f7e1a/src/libraries/Microsoft.Extensions.DependencyModel/src/DependencyContextLoader.cs#L134-L140

Keboo (approved):

src/System.CommandLine.Tests/TrimmingTests.cs
```diff
@@ -24,15 +25,29 @@ public TrimmingTests(ITestOutputHelper output)
...
+    public void App_referencing_system_commandline_can_be_trimmed(string additionalArgs)
     {
         var stdOut = new StringBuilder();
         var stdErr = new StringBuilder();
+
+        var workingDirectory = Path.Combine(Directory.GetCurrentDirectory(), "TrimmingTestApp");
         
+        Process.RunToCompletion(
+            DotnetMuxer.Path.FullName,
+            "clean -c Release -r win-x64",
```
eerhardt, line 40: Does `clean` need a `-c` or a `-r`? Doesn't it just need to be `dotnet clean` and it does it all?

src/System.CommandLine/RootCommand.cs
jonsequitur: No, it's a terrible name! 😅 I'm still tinkering.

jonsequitur: The problem I'm seeing is that `ProcessPath` is returning just `dotnet` in some cases rather than the full path.

eerhardt: When? On Windows or other?

If you are seeing that, can you log an issue in dotnet/runtime? I believe that shouldn't happen.

PR #1636, opened 2022-02-11
fix #1629

### LINKED ISSUE #1629: Default parse error for an enum type should suggest the known values
Given the following option:

```fsharp
#r "nuget: RestoreSources=https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-libraries/nuget/v3/index.json"
#r "nuget: System.CommandLine, 	2.0.0-beta3.22108.1"

open System.CommandLine

let dayOpt = Option<System.DayOfWeek>("--day", "The day to alert on")
dayOpt.Arity = ArgumentArity.ExactlyOne

let r = dayOpt.Parse("--day SleepyDay")
let symR = r.Errors[0] // Cannot parse argument 'SleepyDay' for option '--day' as expected type 'System.DayOfWeek'.
```

the parse error here tells me the input value, the parent option, and the expected enum type, but it doesn't guide me towards what the acceptable values for this enum are. I think it would be helpful if it did.  We already generate [completions from the enum names](https://github.dev/dotnet/command-line-api/blob/77c170fb8a1d8a30aec7e159bf1649314f951491/src/System.CommandLine/Completions/CompletionSource.cs#L59), but I wonder if adding this to help might pose problems for trimmability?

jonsequitur:


src/System.CommandLine/Parsing/ParseResultVisitor.cs
```diff
@@ -483,22 +483,31 @@ private void ValidateAndConvertOptionResult(OptionResult optionResult)
...
+                return;
             }
 
-            if (argumentResult.GetArgumentConversionResult() is FailedArgumentConversionResult failed 
+            if (argumentResult.GetArgumentConversionResult() is FailedArgumentConversionResult failed
                 and not FailedArgumentConversionArityResult)
             {
+                if (argument.Parents.FirstOrDefault() is Option option)
+                {
+                    var completions = option.GetCompletions().ToArray();
```
baronfel, line 503: This is a way better fix than the enum-specific one I suggested on the issue. Nice reuse of existing domain concepts 👍

baronfel (approved):

PR #1641, opened 2022-02-14
remove System.Memory dependency from .NET 6.0

tibel:
This dependency is needed for .NET Standard 2.0 only.


jonsequitur:
Nice catch! Thanks @tibel.

Issue #1638, opened 2022-02-12
Make `Argument` and `Option` classes abstract


jonsequitur:
The `Option` class can currently be instantiated as follows. (`Argument` is similar so for the purpose of this issue, I'll only talk about `Option`, but the change would apply to both.)

```cs
var option = new Option("-x");
option.ValueType = typeof(int);
```

By default, the `ValueType` for `Option` is `string`.

This is largely equivalent to instantiating an `Option<int>`, and the following will return an `int` (as a compile-time `object`):

```cs
var value = parseResult.FindResultFor(option).GetValueOrDefault();
```

The non-generic `Option` can also be late-bound like this:

```cs
var value = parseResult.FindResultFor(option).GetValueOrDefault<int>();
```

This capability wasn't removed because of scenarios where parsers might be built based on a runtime reflection-dependent model, such as attribute-based app models. Forcing use of the generic types was potentially awkward here. As source generation has become more common and as we look to make the library more performant, the use case has become less compelling and the performance and complexity cost less worthwhile.

They are still needed though as base type, e.g. for the `Command.Options` collection.

PR #1656, opened 2022-02-25
Add NativeAOT test

adamsitnik:
There was one warning, about the usage of `Array.CreateInstance`. IMO we can't get rid of it until we implement https://github.com/dotnet/command-line-api/issues/1638

I've disabled this particular warning just to make sure we don't add any new warnings in the meantime. It's not perfect, but still better than no tests at all.


src/System.CommandLine/Binding/ArgumentConverter.DefaultValues.cs
```diff
@@ -19,7 +19,8 @@ internal static partial class ArgumentConverter
                   .SingleOrDefault(c => c.GetParameters().Length == 0)!);
 #endif
 
-    private static Array CreateEmptyArray(Type itemType, int capacity = 0)
+    [UnconditionalSuppressMessage("ReflectionAnalysis", "IL3050", Justification = "We are working on it")]
```
eerhardt, line 22: Link to the issue tracking it?

src/System.CommandLine/Binding/ArgumentConverter.DefaultValues.cs
```diff
@@ -19,7 +19,8 @@ internal static partial class ArgumentConverter
...
+    private static Array CreateArray(Type itemType, int capacity = 0)
```
eerhardt, line 23: Can we drop the default ‘0’ value? If you want a 0 length array use Array.Empty

src/System.CommandLine/Binding/ArgumentConverter.DefaultValues.cs
adamsitnik: Good idea, added

src/System.CommandLine/Binding/ArgumentConverter.DefaultValues.cs
adamsitnik: I've done that in the place where it was possible. Since it's used mostly from non-generic context we can't use `Array.Empty<T>`

adamsitnik:
@jonsequitur the Ubuntu leg is most likely failing due to lack of native dependencies: 

https://github.com/dotnet/runtime/blob/main/src/coreclr/nativeaot/docs/prerequisites.md

Should we add these dependencies or disable the test for this OS?

adamsitnik:
@jonsequitur there are other issues that we need to consider:

1. contributors who don't have the https://github.com/dotnet/runtime/blob/main/src/coreclr/nativeaot/docs/prerequisites.md installed are going to get test failures
2. it takes a lot of time to build an AOT app
3. the test uses latest version of NativeAOT compiler. If there is a new bug introduced to it, the CI might suddenly start failing. It happened to us in BDN and we actually have two kind of tests: using hardcoded version to ensure to regressions in BDN and using latest to ensure it's supported

src/System.CommandLine.Tests/NativeAOTTests.cs
```diff
@@ -0,0 +1,79 @@
...
+    public NativeAOTTests(ITestOutputHelper output)
+    {
+        _output = output;
+
+        _systemCommandLineDllPath = typeof(Command).Assembly.Location;
+    }
+
+    [ReleaseBuildOnlyTheory]
+    [InlineData("-p:IlcOptimizationPreference=Speed")]
+    [InlineData("-p:IlcOptimizationPreference=Size")]
```
eerhardt, line 31: Do we need to do both? This test is just to catch warnings, right? Will it produce different warnings between these two options? If we don't need to do both, that will cut the amount of time in half.

src/System.CommandLine.Tests/TestApps/NativeAOT/NativeAOT.csproj
```diff
@@ -0,0 +1,21 @@
...
+
+	<PropertyGroup>
+		<SystemCommandLineDllPath Condition="'$(SystemCommandLineDllPath)'==''">..\..\..\System.CommandLine\bin\Release\net6.0\System.CommandLine.dll</SystemCommandLineDllPath>
+	</PropertyGroup>
+
+	<ItemGroup>
+		<Reference Include="SystemCommandLineDll">
+			<HintPath>$(SystemCommandLineDllPath)</HintPath>
+		</Reference>
+		<PackageReference Include="Microsoft.DotNet.ILCompiler" Version="7.0.0-*" />
```
eerhardt, line 18: Can we instead hard-code a version, and then set up a Maestro update to keep it updated?

src/System.CommandLine/Binding/ArgumentConverter.DefaultValues.cs
eerhardt:
```suggestion
    private static Array CreateArray(Type itemType, int capacity)
```

I don't see any calls to this method that use the default value.

jonsequitur:
> @jonsequitur the Ubuntu leg is most likely failing due to lack of native dependencies:
> 
> https://github.com/dotnet/runtime/blob/main/src/coreclr/nativeaot/docs/prerequisites.md
> 
> Should we add these dependencies or disable the test for this OS?

I'd prefer to add the dependencies. I'll look into it.

jonsequitur:
> @jonsequitur there are other issues that we need to consider:
> 
> 1. contributors who don't have the https://github.com/dotnet/runtime/blob/main/src/coreclr/nativeaot/docs/prerequisites.md installed are going to get test failures
> 2. it takes a lot of time to build an AOT app

We've been using `ReleaseBuildOnlyFact` and `ReleaseBuildOnlyTheory` for tests that are important to run during CI but would cause friction during the dev loop, such as complex setup or tests that are necessarily slow. 

> 3. the test uses latest version of NativeAOT compiler. If there is a new bug introduced to it, the CI might suddenly start failing. It happened to us in BDN and we actually have two kind of tests: using hardcoded version to ensure to regressions in BDN and using latest to ensure it's supported

My thought is the hardcoded version to prevent regression is simpler and more important, since here we're testing System.CommandLine and not the AOT compiler. Ingesting new dependencies dynamically can break this CI pipeline, so I would tend to avoid that.

jonsequitur:
@adamsitnik The update to Ubuntu 20.04 is merged. Could you rebase this and see if the tests now pass?

adamsitnik:
> The update to Ubuntu 20.04 is merged. Could you rebase this and see if the tests now pass?

@jonsequitur I've rebased the branch but the test is till failing on Ubuntu. Is this repo using CI machines from MS internal feed? Or some public ones and we can define what is needed?

This is what we needed in BDN:

https://github.com/dotnet/BenchmarkDotNet/blob/ca51035a62ea89de6059c2822c4a5e5df58560a3/azure-pipelines.Ubuntu.yml#L17-L22

jonsequitur:
Do you want to try including that in your PR?

jonsequitur (approved):

PR #1671, opened 2022-03-12
Fix #1663

### LINKED ISSUE #1663: First argument cannot be the executable name
When using RootCommand, if the first argument matches the command name it is stripped. This is one of at least two characteristics that differentiate RootCommand from Command. This includes the name of the executable without a suffix:

```
> MyExe MyExe
```

will strip the argument if RootCommand is the root of the parse tree. The current workaround if this is a problem is to use Command as the root. The name you use in the constructor will not matter. 

This was done so that scenarios, including passing System.Environment.GetCommandLineArgs(), that include the executable will be correctly handled. 

There are a couple of problems with this approach, such that I think we should consider a minor-ish change. 

* How args are handled becomes a characteristic of the parse tree, not the callsite. This means that the same parse tree can't work in both scenarios. The callsite knows the source of the args.
* Telling folks to enter garbage as a name is silly. But we certainly do not want to make name optional on command, as havoc would ensue.

I propose: 

* Command continue to work as today, no changes. It can be a root and we can use a meaningful unused names like "Root". In this case, the parse tree is marked to _never_ remove the first arg.
* Alter the logic of RootCommand so that the callsite can control whether the arg is removed if it matches the exe name.

I think the syntax of this deserves discussion, but I'll throw out a couple of ideas:

cmd.Invoke(args) // No longer strips - this is a breaking change, but I think will affect few folks
cmd.InvokeStripExename(args) // ok, it is a stinky name, but I'd like to be clear here.
cmd.Invoke(stripExeName, args) // because a bool in that arg position is never currently valid?

If we can do the last, it is prettiest

@jkotas @jonsequitur @baronfel

jonsequitur:
Fixes #1663.

### DIFF
--- a/src/System.CommandLine/Parsing/StringExtensions.cs
+++ b/src/System.CommandLine/Parsing/StringExtensions.cs
@@ -324,8 +324,6 @@ namespace System.CommandLine.Parsing
 
             var list = new List<string>();
 
-            string? potentialRootCommand = null;
-
             if (args.Count > 0)
             {
                 if (inferRootCommand &&
@@ -338,7 +336,7 @@ namespace System.CommandLine.Parsing
                 {
                     try
                     {
-                        potentialRootCommand = Path.GetFileName(args[0]);
+                        var potentialRootCommand = Path.GetFileName(args[0]);
 
                         if (rootCommand.HasAlias(potentialRootCommand))
                         {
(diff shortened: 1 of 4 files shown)


src/System.CommandLine/Parsing/StringExtensions.cs
```diff
@@ -322,32 +328,45 @@ void ReadResponseFile(string filePath, int i)
...
 
             int startAt = 0;
 
-            if (FirstArgMatchesRootCommand())
+            var firstArgMatchesRootCommand = FirstArgMatchesRootCommand();
+
+            if (inferRootCommand)
             {
-                startAt = 1;
+                if (firstArgMatchesRootCommand)
```
Keboo, line 366: nit: Combine if statements to reduce nesting?
```C#
if (inferRootCommand && firstArgMatchesRootCommand)
```

Keboo (approved):

src/System.CommandLine/Parsing/StringExtensions.cs
jonsequitur: Or just delete the whole `if` block... :)

PR #1677, opened 2022-03-15
Fix #1638

### LINKED ISSUE #1638: Make `Argument` and `Option` classes abstract

jonsequitur:
This addresses #1638.


src/System.CommandLine.Tests/GlobalOptionTests.cs
```diff
@@ -1,31 +1,13 @@
...
-
-using System.Threading.Tasks;
 using FluentAssertions;
 using Xunit;
 
 namespace System.CommandLine.Tests
 {
     public class GlobalOptionTests
     {
-        [Fact]
```
baronfel, line 15: Can you walk me through why this doesn't apply anymore? I'm not following that thread.

jonsequitur: This test was green for the wrong reason after we stopped performing this kind of validation up front and introduced `CommandLineConfiguration.ThrowIfInvalid`. I'll be adding some additional global option tests and validations in a future PR.

PR #1684, opened 2022-03-23
Fix #1209

### LINKED ISSUE #1209: UseCommandHandler not works with sub-class.
codes like this won't work.
```cs
abstract class Base : ICommandHandler
{
    protected virtual void Run()
    {}

    {
        Run();
        return Task.FromResult(0);
    }
}

class Sub : Base {}

host.UseCommandHandler<MyCommand, Sub>()
```
exception:
```
Unhandled exception. System.Reflection.TargetException: Non-static method requires a target.
   at System.Reflection.RuntimeMethodInfo.CheckConsistency(Object target)
   at System.Reflection.RuntimeMethodInfo.InvokeArgumentsCheck(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at System.Reflection.MethodBase.Invoke(Object obj, Object[] parameters)
   at System.CommandLine.Invocation.ModelBindingCommandHandler.InvokeAsync(InvocationContext context)
   at System.CommandLine.Invocation.InvocationPipeline.<>c__DisplayClass4_0.<<BuildInvocationChain>b__0>d.MoveNext()
```

eisbaer66:
`UseCommandHandler(...)` binds base-classes of the CommandHandler-type, to allow implementing ICommandHandler in a base-class.
fixes #1209

### DIFF
--- a/src/System.CommandLine.Hosting/HostingExtensions.cs
+++ b/src/System.CommandLine.Hosting/HostingExtensions.cs
@@ -108,6 +108,12 @@ namespace System.CommandLine.Hosting
                 && command.GetType() == commandType)
             {
                 invocation.BindingContext.AddService(handlerType, c => c.GetService<IHost>().Services.GetService(handlerType));
+                var baseType = handlerType.BaseType;
+                while (baseType != null && baseType != typeof(object))
+                {
+                    invocation.BindingContext.AddService(baseType, c => c.GetService<IHost>().Services.GetService(handlerType));
+                    baseType = baseType.BaseType;
+                }
                 builder.ConfigureServices(services =>
                 {
                     services.AddTransient(handlerType);
--- a/src/System.CommandLine.Hosting.Tests/HostingHandlerTest.cs
+++ b/src/System.CommandLine.Hosting.Tests/HostingHandlerTest.cs
@@ -113,6 +113,28 @@ namespace System.CommandLine.Hosting.Tests
             service.StringValue.Should().Be("TEST");
         }
 
+        [Fact]
+        public static async Task Invokes_DerivedClass()
+        {
+            var service = new MyService();
+
+            var parser = new CommandLineBuilder(
+                                                new MyCommand()
+                                               )
+                         .UseHost((builder) => {
+                                      builder.ConfigureServices(services =>
+                                                                {
+                                                                    services.AddTransient(x => service);
+                                                                })
+                                             .UseCommandHandler<MyCommand, MyCommand.MyDerivedHandler>();
+                                  })
+                         .Build();
+
+            var result = await parser.InvokeAsync(new string[] { "--int-option", "54" });
+
+            service.Value.Should().Be(54);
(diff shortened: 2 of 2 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/command-line-api?pullRequest=1684) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/command-line-api?pullRequest=1684) before we can accept your contribution.<br/><br/>:x: eisbaer66 [sign now](https://cla.dotnetfoundation.org/dotnet/command-line-api?pullRequest=1684)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/command-line-api?pullRequest=1684) it.</sub>

src/System.CommandLine.Hosting/HostingExtensions.cs
```diff
@@ -108,6 +108,12 @@ public static IHostBuilder UseCommandHandler(this IHostBuilder builder, Type com
                 && command.GetType() == commandType)
             {
                 invocation.BindingContext.AddService(handlerType, c => c.GetService<IHost>().Services.GetService(handlerType));
+                var baseType = handlerType.BaseType;
+                while (baseType != null && baseType != typeof(object))
+                {
+                    invocation.BindingContext.AddService(baseType, c => c.GetService<IHost>().Services.GetService(handlerType));
```
jonsequitur, line 114: If I use two different derived types this way, will one not overwrite the other's base type registration?

eisbaer66: Very true. Was a bit too short-sighted when the code got rid of the initial error sorry for that :/
I have explored an other option, though:
1. adding a overload for `CommandHandler.Create(...)` that takes the concrete handler type (called inside `UseCommandHandler`)
2. adding similar overloads all the way to `ModelBindingCommandHandler`
3. using the concrete type inside of `ModelBindingCommandHandler` to resolve the handler from the concrete type , falling back to the previously used `_handlerMethodInfo!.DeclaringType` if no concrete type was specified (here `_handlerMethodInfo!.DeclaringType` is the base-type, that implements `ICommandHandler.InvokeAsync`)

I think we should ultimately get rid of this fallback behavior and require the handler type to be specified. Even if a overload with a concrete `_invocationTarget` was used, since it could be null.
Didn't do that yet, since it comes with breaking changes on some lower level classes (`.UseCommandHandler()` signature would not change) and a bunch of calls needing changes.
So I wanted to hear what you have to say about this before going forward.

Should i add the new code here? Or make a  new PR?

PR #1701, opened 2022-04-05
Apply comments from #7859

tdykstra:
https://github.com/dotnet/dotnet-api-docs/pull/7859 was merged before @gewarren's comments were addressed. This PR applies the same changes but makes them in the /// comments.

### DIFF
--- a/src/System.CommandLine/IdentifierSymbol.cs
+++ b/src/System.CommandLine/IdentifierSymbol.cs
@@ -60,9 +60,12 @@ namespace System.CommandLine
         }
 
         /// <summary>
-        /// Adds an alias. Multiple aliases can be added, most often used to provide a shorthand alternative.
+        /// Adds an <see href="/dotnet/standard/commandline/syntax#aliases">alias</see>.
         /// </summary>
         /// <param name="alias">The alias to add.</param>
+        /// <remarks>
+        /// You can add multiple aliases for a symbol.
+        /// </remarks>
         public void AddAlias(string alias)
         {
             ThrowIfAliasIsInvalid(alias);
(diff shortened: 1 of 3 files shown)


src/System.CommandLine/Parsing/Parser.cs
```diff
@@ -23,22 +23,22 @@ public Parser(Command command) : this(new CommandLineConfiguration(command))
         }
         
         /// <summary>
-        /// Initializes a new instance of the Parser class with using the default <seealso cref="RootCommand"/>.
+        /// Initializes a new instance of the <see cref="T:System.CommandLine.Parsing.Parser" /> class using the default <see cref="T:System.CommandLine.RootCommand" />.
```
gewarren, line 26: In the ///, it's best not to add prefixes to crefs, and also not to fully qualify the type. mdoc will convert it as needed.

```suggestion
        /// Initializes a new instance of the <see cref="Parser" /> class using the default <see cref="RootCommand" />.
```

src/System.CommandLine/IdentifierSymbol.cs
```diff
@@ -60,9 +60,12 @@ public override string Name
         }
 
         /// <summary>
-        /// Adds an alias. Multiple aliases can be added, most often used to provide a shorthand alternative.
+        /// Adds an <see href="/dotnet/standard/commandline/syntax#aliases">alias</see>.
         /// </summary>
         /// <param name="alias">The alias to add.</param>
+        /// <remarks>
+        /// You can add multiple aliases for a symbol.
```
gewarren, line 67: I thought there was some info added about what aliases could be used for. Not serious though.

gewarren (approved):
Looks good, thanks Tom!

src/System.CommandLine/IdentifierSymbol.cs
tdykstra: I replaced that in the summary with a link on "alias" to the aliases section of the syntax article https://docs.microsoft.com/dotnet/standard/commandline/syntax#aliases

jonsequitur (approved):

jonsequitur:
Thanks, @tdykstra!

PR #1714, opened 2022-04-20
Fix #1665

### LINKED ISSUE #1665: simplify response file parsing
Response file parsing can currently be customized by a developer using System.CommandLine using the `CommandLineBuilder.ParseRespondFilesAs(ResponseFileHandling)` method. The `ResponseFileHandling` enum provides two modes (in addition to `Disabled`) but the behavior is not apparent to end users, and is occasionally surprising. 

The proposal is to support only a single response file parsing behavior by removing the existing methods and supporting only enable/disable. The new behavior would follow these rules:

* Multiple-line files will be parsed treating newlines as equivalent to spaces.
* Multiple tokens on a single line will be parsed as though passed verbatim on the command line.
  * Including a space within a token will now require enclosing the token in quotes, e.g. `"this is one token"`. 
  * Shell-specific escaping of quotes will not be required.
* Allow comment lines using a `#` prefix at the beginning of the line  

Open questions:

* How should quotation marks be escaped?

jonsequitur:
This removes the distinction between line-delimited and space-delimited response files, fixing #1665.

It does this by generalizing the mechanism where tokens prefixed with `@` can be replaced with other tokens. Response files are one example of this but now alternative mechanisms can be introduced.

```csharp
var parser = new CommandLineBuilder(command)
             .UseTokenReplacer((string tokenToReplace, out IReadOnlyList<string> tokens, out string message) =>
             {
                 tokens = new[] { "123" };
                 message = null;
                 return true;
             })
             .Build();

var result = parser.Parse("@interpolate-me");

result.GetValueForArgument(argument).Should().Be(123);
```


src/System.CommandLine.Tests/CustomTokenReplacerTests.cs
```diff
@@ -0,0 +1,36 @@
...
+{
+    [Fact]
+    public void Custom_token_replacer_can_expand_argument_values()
+    {
+        var argument = new Argument<int>();
+
+        var command = new RootCommand { argument };
+
+        var parser = new CommandLineBuilder(command)
+                     .UseTokenReplacer((string tokenToReplace, out IReadOnlyList<string> tokens, out string message) =>
```
Keboo, line 22: Should this test also assert on the value of `tokenToReplace`? I assume in this situation it would be "@interpolate-me" or would it be "interpolate-me"?

Keboo (approved):

src/System.CommandLine.Tests/CustomTokenReplacerTests.cs
jonsequitur: I'll add a test for this. The received `tokenToReplace` will not include the `@`. Open to changing that. My assumption was that this should be transparently removed, sort of like quotation marks.

Keboo: That sounds good to me

Keboo (approved):

PR #1722, opened 2022-04-29
Updating HelpBuilder.WriteColumns to be easier to extend

Keboo:
While updating ClangSharp to the latest System.CommandLine I ran into the issue of a custom HelpBuilder wanting to leverage the two column layout for some arbitrary items. This required building a HelpContext despite the method only needing a TextWriter to output to.

See: https://github.com/dotnet/ClangSharp/pull/343

### DIFF
--- a/src/System.CommandLine.ApiCompatibility.Tests/ApiCompatibilityApprovalTests.System_CommandLine_api_is_not_changed.approved.txt
+++ b/src/System.CommandLine.ApiCompatibility.Tests/ApiCompatibilityApprovalTests.System_CommandLine_api_is_not_changed.approved.txt
@@ -312,7 +312,7 @@ System.CommandLine.Help
     public System.Void CustomizeSymbol(System.CommandLine.Symbol symbol, System.Func<HelpContext,System.String> firstColumnText = null, System.Func<HelpContext,System.String> secondColumnText = null, System.Func<HelpContext,System.String> defaultValue = null)
     public TwoColumnHelpRow GetTwoColumnRow(System.CommandLine.Symbol symbol, HelpContext context)
     public System.Void Write(HelpContext context)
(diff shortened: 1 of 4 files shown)


jonsequitur:
I wonder if another helper method like this work:

https://github.com/dotnet/command-line-api/blob/74ae46ea8571269a66945049ce9f134a91d67d1e/src/System.CommandLine/Help/HelpBuilderExtensions.cs#L31-L40

`HelpBuilder.Write` also takes a `HelpContext` and keeping things consistent on this type leaves an opening to add things to the `HelpContext` later.

Keboo:
The only issue with that type of extension is it makes is very difficult for anyone to use the built-in two column layout with arbitrary content. For the PR linked above, there is a [large bit of content](https://github.com/dotnet/ClangSharp/blob/main/sources/ClangSharpPInvokeGenerator/Program.cs#L586-L592) simply wants to be written out using the two-column layout. To work around the API [I did something similar](https://github.com/dotnet/ClangSharp/blob/main/sources/ClangSharpPInvokeGenerator/CustomHelpBuilder.cs#L33-L39), the suggested extension method still assumes that the input to the function will come from a command. Where in this case what is desired is a way for an _option_ to give "help-like" behavior with a lot of additional details. Really that code path doesn't really need to even go through a help builder, it is only using it to get some of the formatting niceness, but the input is "custom".

jonsequitur:
I wonder if that suggests that the problem isn't so much the `HelpContext` as the fact that it's so `Command`-centric.

Keboo:
I think that is a good summary. Which brings up two questions. 1. Should it be less command centric? 2. Any thoughts on what that should like?

jonsequitur:
One approach off the top of my head could be to change `HelpBuilder.Write` to accept `Symbol` and change `HelpContext.Command` to `HelpContext.Symbol`. That might have a bit of a ripple effect, though.

PR #1729, opened 2022-05-04
SetHandler improvements

jonsequitur:
These changes attempt to improve the clarity of the `SetHandler` methods by removing the `params` parameters and replacing them with a number of `IValueDescriptor<T>` parameters matching the number of generic type arguments.

Since DI-like behaviors are fairly non-obvious, I'll likely also experiment here with removing the internal service provider and see if `BinderBase<T>` can fulfill its role.


src/System.CommandLine.Rendering.Tests/TerminalModeTests.cs
```diff
@@ -44,7 +44,7 @@ public async Task Sets_output_mode_to_Ansi_when_specified_by_output_directive(Ou
             OutputMode detectedOutputMode = OutputMode.Auto;
 
             var command = new Command("hello");
-            command.SetHandler((InvocationContext ctx) =>
+            command.SetHandler(ctx =>
```
baronfel, line 47: it's really nice for the simple handlers to be naturally-typed here (and in the other 1000 callsites in this PR)

baronfel (commented):
Nothing major, but man those simple callsites look a lot nicer.

src/System.CommandLine.Rendering.Tests/TerminalModeTests.cs
jonsequitur: Agreed. Settling on `InvocationContext` for the case where we do this (i.e. removing support for `ParseResult`, etc. as parameters) reduces a lot of DI-ish complexity in these handlers. If we pull this thread hard enough, the need for the service provider might disappear.

src/System.CommandLine.NamingConventionBinder/ModelBinder.cs
```diff
@@ -312,12 +312,26 @@ private bool UpdateInstanceInternalNotifyIfNonDefaultsUsed<T>(T instance, Bindin
...
         return (null, false);
     }
 
+    private static BoundValue DefaultForValueDescriptor(
+        IValueDescriptor valueDescriptor,
+        BindingContext context)
+    {
+        var valueSource = ValueDescriptorDefaultValueSource.Instance;
+
+        valueSource.TryGetValue(valueDescriptor, context, out var value);
```
Keboo, line 327: nit: I dislike `var` here as it is not clear what the type of `value` is.

Keboo, line 327: Should we care about the case where `TryGetValue` returns false? Should there be something line `BoundValue.None` to return in that case?

Keboo (commented):
Looks good. Might be worth reviewing System.CommandLine.Generator after this merges, as I suspect it could be cleaned up a little bit to leverage the InvocationContext SetHandler overload.

PR #1756, opened 2022-06-06
Add support for timeout in process termination handling

JanKrivanek:
**Background:**

related to https://github.com/dotnet/templating/issues/4799

It would be helpfull to be able to specify optional maximum timeout for handling the process termination - in order to prevent unwanted 'unkillable' command - especially for cases where command builder and and actual implementation of the command have different maintainers (e.g. a case of dotnet sdk).

The process that don't repond to Ctrl-C can still be killed by other means - but it may create unpleasant and inconsistent experience to users. The idea is to have top level handler that gives command chance to perform cleanup in correct way, but bail out if they do not return in timely manner (e.g. due to not using the cancellation patter at all).

If the optinal timeout is not specified, the original behavior remains (no timeout enforced).

### DIFF
--- a/src/System.CommandLine.ApiCompatibility.Tests/ApiCompatibilityApprovalTests.System_CommandLine_api_is_not_changed.approved.txt
+++ b/src/System.CommandLine.ApiCompatibility.Tests/ApiCompatibilityApprovalTests.System_CommandLine_api_is_not_changed.approved.txt
@@ -240,7 +240,7 @@ System.CommandLine.Builder
   public static class CommandLineBuilderExtensions
     public static CommandLineBuilder AddMiddleware(this CommandLineBuilder builder, System.CommandLine.Invocation.InvocationMiddleware middleware, System.CommandLine.Invocation.MiddlewareOrder order = Default)
     public static CommandLineBuilder AddMiddleware(this CommandLineBuilder builder, System.Action<System.CommandLine.Invocation.InvocationContext> onInvoke, System.CommandLine.Invocation.MiddlewareOrder order = Default)
-    public static CommandLineBuilder CancelOnProcessTermination(this CommandLineBuilder builder)
+    public static CommandLineBuilder CancelOnProcessTermination(this CommandLineBuilder builder, System.Nullable<System.TimeSpan> cancelationProcessingTimeout = null)
     public static CommandLineBuilder EnableDirectives(this CommandLineBuilder builder, System.Boolean value = True)
     public static CommandLineBuilder EnableLegacyDoubleDashBehavior(this CommandLineBuilder builder, System.Boolean value = True)
     public static CommandLineBuilder EnablePosixBundling(this CommandLineBuilder builder, System.Boolean value = True)
--- a/src/System.CommandLine.ApiCompatibility.Tests/ApiCompatibilityApprovalTests.System_CommandLine_api_is_not_changed.approved.txt
+++ b/src/System.CommandLine.ApiCompatibility.Tests/ApiCompatibilityApprovalTests.System_CommandLine_api_is_not_changed.approved.txt
@@ -240,7 +240,7 @@ System.CommandLine.Builder
   public static class CommandLineBuilderExtensions
     public static CommandLineBuilder AddMiddleware(this CommandLineBuilder builder, System.CommandLine.Invocation.InvocationMiddleware middleware, System.CommandLine.Invocation.MiddlewareOrder order = Default)
     public static CommandLineBuilder AddMiddleware(this CommandLineBuilder builder, System.Action<System.CommandLine.Invocation.InvocationContext> onInvoke, System.CommandLine.Invocation.MiddlewareOrder order = Default)
-    public static CommandLineBuilder CancelOnProcessTermination(this CommandLineBuilder builder, System.Nullable<System.TimeSpan> cancelationProcessingTimeout = null)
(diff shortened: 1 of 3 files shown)


src/System.CommandLine.ApiCompatibility.Tests/ApiCompatibilityApprovalTests.System_CommandLine_api_is_not_changed.approved.txt
```diff
@@ -240,7 +240,7 @@ System.CommandLine.Builder
   public static class CommandLineBuilderExtensions
     public static CommandLineBuilder AddMiddleware(this CommandLineBuilder builder, System.CommandLine.Invocation.InvocationMiddleware middleware, System.CommandLine.Invocation.MiddlewareOrder order = Default)
     public static CommandLineBuilder AddMiddleware(this CommandLineBuilder builder, System.Action<System.CommandLine.Invocation.InvocationContext> onInvoke, System.CommandLine.Invocation.MiddlewareOrder order = Default)
-    public static CommandLineBuilder CancelOnProcessTermination(this CommandLineBuilder builder)
+    public static CommandLineBuilder CancelOnProcessTermination(this CommandLineBuilder builder, System.Nullable<System.TimeSpan> cancelationProcessingTimeout = null)
```
jonsequitur, line 243: It's fine to just call this `timeout` I think.

src/System.CommandLine.Tests/Invocation/CancelOnProcessTerminationTests.cs
```diff
@@ -91,6 +91,169 @@ public async Task CancelOnProcessTermination_cancels_on_process_termination(int
...
+            bool processExited = process.WaitForExit(10000);
+            if (!processExited)
+            {
+                process.Kill();
+                process.WaitForExit();
+            }
+            processExited.Should().Be(true);
+
+            // Verify the process exit code
+            process.ExitCode.Should().Be(timeOutMs.HasValue ? ForceTerminationCode : CancelledExitCode);
```
jonsequitur, line 254: `timeOutMs` always has a value in this `Theory`. I'd suggest removing the branching logic or, if the two theories' implementations are substantially similar, hoist the exit codes into the `InlineData` definitions and only have one `Theory`.

JanKrivanek: Unfortunately it's not possible to have that parameter in `InlineData` as `RemoteExecutor` doesn't capture the closure variables - so parameters would not get propagated (hence the reason for 2 test methods).

I'll remove that parameter to make it less confusing

JanKrivanek:
@jonsequitur thank you for the initial feedback (should be addressed now). Are there any further unaddressed concerns or suggestions?

src/System.CommandLine.Tests/Invocation/CancelOnProcessTerminationTests.cs
```diff
@@ -91,6 +91,170 @@ public async Task CancelOnProcessTermination_cancels_on_process_termination(int
             process.ExitCode.Should().Be(CancelledExitCode);
         }
 
+        [LinuxOnlyTheory]
+        [InlineData(SIGINT)]
+        [InlineData(SIGTERM)]
+        public async Task CancelOnProcessTermination_null_timeout_on_cancel_processing(int signo)
```
jonsequitur, line 97:
```suggestion
        public async Task CancelOnProcessTermination_cancels_on_process_termination_when_no_timeout_is_specified(int signo)
```

src/System.CommandLine.Tests/Invocation/CancelOnProcessTerminationTests.cs
```diff
@@ -91,6 +91,170 @@ public async Task CancelOnProcessTermination_cancels_on_process_termination(int
...
+            processExited.Should().Be(true);
+
+            // Verify the process exit code
+            process.ExitCode.Should().Be(CancelledExitCode);
+        }
+
+        [LinuxOnlyTheory]
+        [InlineData(SIGINT)]
+        [InlineData(SIGTERM)]
+        public async Task CancelOnProcessTermination_timeout_on_cancel_processing(int signo)
```
jonsequitur, line 176:
```suggestion
        public async Task CancelOnProcessTermination_provides_CancellationToken_that_signals_termination_and_is_cancelled_at_the_specified_timeout(int signo)
```

jonsequitur:
> @jonsequitur thank you for the initial feedback (should be addressed now). Are there any further unaddressed concerns or suggestions?

I suggested a couple of test renames but feel free to adjust if you have something more  precise. The goal is to  embed the expectations about the scenario in the test name, like a spec.

src/System.CommandLine.Tests/Invocation/CancelOnProcessTerminationTests.cs
jonsequitur, line 97: Is this test redundant with the pre-existing test above?

JanKrivanek: It is redundand with the current `CancelOnProcessTermination` implementation. But it guards the possibility of contract/behavior change on explict passing of `null` as `timeout`.
I can remove it if you feel it's superfluous.

JanKrivanek:
@jonsequitur Are there any further unaddressed concerns or suggestions?

PR #1761, opened 2022-06-12
Enable nullable reference types for Hosting

### LINKED ISSUE #1760: Hosting Nullable reference types support
It would be better to support nullable reference types for all libraries

WeihanLi:
Fixes #1760 

- Enable nullable reference types for Hosting
- Renaming `commandline` => `commandLine` to fix typo
- Remove unnecessary `ToList` for configuration extension

### DIFF
--- a/src/System.CommandLine.Hosting/Directory.Build.props
+++ b/src/System.CommandLine.Hosting/Directory.Build.props
@@ -1,8 +1,7 @@
 <Project>
 
   <PropertyGroup>
-    <UseAlphaVersion>true</UseAlphaVersion>
-    <Nullable>enable</Nullable>
+    <UseAlphaVersion>true</UseAlphaVersion>    
   </PropertyGroup>
 
   <Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.props', '$(MSBuildThisFileDirectory)../'))" />
--- a/src/System.CommandLine.Hosting/Directory.Build.props
+++ b/src/System.CommandLine.Hosting/Directory.Build.props
@@ -1,7 +1,7 @@
 <Project>
 
   <PropertyGroup>
-    <UseAlphaVersion>true</UseAlphaVersion>    
+    <UseAlphaVersion>true</UseAlphaVersion>
   </PropertyGroup>
 
   <Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.props', '$(MSBuildThisFileDirectory)../'))" />
--- a/src/System.CommandLine.Hosting/Directory.Build.props
+++ b/src/System.CommandLine.Hosting/Directory.Build.props
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <UseAlphaVersion>true</UseAlphaVersion>
+    <Nullable>enable</Nullable>
   </PropertyGroup>
 
   <Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.props', '$(MSBuildThisFileDirectory)../'))" />
--- a/src/System.CommandLine.Hosting/HostingExtensions.cs
+++ b/src/System.CommandLine.Hosting/HostingExtensions.cs
@@ -145,10 +145,7 @@ namespace System.CommandLine.Hosting
         {
             _ = invocationContext ?? throw new ArgumentNullException(paramName: nameof(invocationContext));
             var hostModelBinder = new ModelBinder<IHost>();
-            var instance = hostModelBinder.CreateInstance(invocationContext.BindingContext);
-            if (instance is IHost host)
-              return host;            
-            return null;
+            return hostModelBinder.CreateInstance(invocationContext.BindingContext) as IHost;
         }
     }
 }
--- a/src/System.CommandLine.Hosting/HostingExtensions.cs
+++ b/src/System.CommandLine.Hosting/HostingExtensions.cs
@@ -17,8 +17,8 @@ namespace System.CommandLine.Hosting
         private const string ConfigurationDirectiveName = "config";
 
         public static CommandLineBuilder UseHost(this CommandLineBuilder builder,
-            Func<string[], IHostBuilder> hostBuilderFactory,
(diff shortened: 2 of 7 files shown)


src/System.CommandLine.Hosting/Directory.Build.props
```diff
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <UseAlphaVersion>true</UseAlphaVersion>
+    <Nullable>enable</Nullable>
```
jonsequitur, line 5: Please make this change in the `.csproj` file for consistency with the other projects.

src/System.CommandLine.Hosting/HostingExtensions.cs
```diff
@@ -107,7 +107,7 @@ public static IHostBuilder UseCommandHandler(this IHostBuilder builder, Type com
                 && invocation.ParseResult.CommandResult.Command is Command command
                 && command.GetType() == commandType)
             {
-                invocation.BindingContext.AddService(handlerType, c => c.GetService<IHost>().Services.GetService(handlerType));
+                invocation.BindingContext.AddService(handlerType, c => c.GetRequiredService<IHost>().Services.GetService(handlerType));
```
jonsequitur, line 110: Why this change? I would think the `IHost` is a required service here.

src/System.CommandLine.Hosting/HostingExtensions.cs
```diff
@@ -141,11 +141,14 @@ public static InvocationContext GetInvocationContext(this HostBuilderContext con
...
-        public static IHost GetHost(this InvocationContext invocationContext)
+        public static IHost? GetHost(this InvocationContext invocationContext)
         {
             _ = invocationContext ?? throw new ArgumentNullException(paramName: nameof(invocationContext));
             var hostModelBinder = new ModelBinder<IHost>();
-            return (IHost)hostModelBinder.CreateInstance(invocationContext.BindingContext);
+            var instance = hostModelBinder.CreateInstance(invocationContext.BindingContext);
+            if (instance is IHost host)
+              return host;            
+            return null;
```
jonsequitur, line 151: I would think we should  not allow null here.

src/System.CommandLine.Hosting/InvocationLifetime.cs
```diff
@@ -17,7 +17,7 @@
 
 namespace System.CommandLine.Hosting
 {
-    public class InvocationLifetime : IHostLifetime
+    public sealed class InvocationLifetime : IHostLifetime
```
jonsequitur, line 20: What is  the reason for this change?

src/System.CommandLine.Hosting/InvocationLifetimeOptions.cs
```diff
@@ -2,7 +2,7 @@
 
 namespace System.CommandLine.Hosting
 {
-    public class InvocationLifetimeOptions : ConsoleLifetimeOptions
+    public sealed class InvocationLifetimeOptions : ConsoleLifetimeOptions
```
jonsequitur, line 5: What is the reason for this change?

src/System.CommandLine.Hosting/HostingExtensions.cs
WeihanLi: `GetRequiredService` would return service and ensure the service is not null, `GetService` would return `null` when no service found, think we expect `IHost` exists, so maybe better to use `GetRequiredService` 
<https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.serviceproviderserviceextensions.getrequiredservice?view=dotnet-plat-ext-6.0>

src/System.CommandLine.Hosting/InvocationLifetime.cs
WeihanLi: From this issue https://github.com/dotnet/runtime/issues/49944, sealed types would get performance gains and there're no virtual methods to override so think it could be a sealed class, same for the `InvocationLifetimeOptions`

src/System.CommandLine.Hosting/HostingExtensions.cs
WeihanLi: I think so before, from test code <https://github.com/dotnet/command-line-api/blob/209b724a3c843253d3071e8348c353b297b0b8b5/src/System.CommandLine.Hosting.Tests/HostingTests.cs#L401-L403>
it may be null, if this should not be `null`, guess we should throw an `InvalidOperationException`?

src/System.CommandLine.Hosting/InvocationLifetimeOptions.cs
WeihanLi: Reverted

src/System.CommandLine.Hosting/InvocationLifetime.cs
WeihanLi: Reverted

src/System.CommandLine.Hosting/Directory.Build.props
WeihanLi: Updated

src/System.CommandLine.Hosting/InvocationLifetime.cs
jonsequitur: This is good performance advice but generally something I'd avoid applying universally without a benchmark indicating that these specific types will benefit from being `sealed`.

src/System.CommandLine.Hosting/HostingExtensions.cs
WeihanLi: Think it should not be registered when `IHost` or the handler is `null`, maybe I could add a null check for that

src/System.CommandLine.Hosting/HostingExtensions.cs
Keboo, line 151: More succinct syntax could be (assuming we keep the null return):
```suggestion
            return hostModelBinder.CreateInstance(invocationContext.BindingContext) as IHost;
```

PR #1767, opened 2022-06-18
Fix #1759

### LINKED ISSUE #1759: `ArgumentResult.OnlyTake(0)` doesn't work with a scalar argument followed by an enumerable argument
If you've got two arguments, one of arity ZeroOrOne and the second of arity ZeroOrMore, the token passed to the first argument isn't returned to the pool (and therefore passed to the second argument) when `OnlyTake(0)` is used to reject the token passed to the scalar argument.

<details>
<summary>Sample C# code</summary>

```csharp
using System.CommandLine;

class TestCommand : Command
{
    private static ParseArgument<string> Nope =
            (ctx) =>
            {
                ctx.OnlyTake(0);
                return null;
            };
    public static readonly Argument<string> Scalar = new Argument<string>( parse: Nope);
    public static readonly Argument<string[]> Multiple = new();

    public TestCommand() : base("test", "do the testy") {
        this.Add(Scalar);
        this.Add(Multiple);

        this.SetHandler(ctx => {
            var parseResult = ctx.ParseResult;
            var forward = parseResult.GetValueForArgument(Multiple);
            Console.WriteLine("Forwarded:");
            foreach (var arg in forward ?? Array.Empty<string>()) {
                Console.WriteLine($"\t{arg}");
            }
        });
    }
}

public static partial class Program {
    public static int Main(string[] argv) {
        return new TestCommand().Invoke(argv);
    }
}
```
</details>

If you `dotnet run [parse] a b c` this code,  you get the output `[ test [ string <> ] [ string[] <b> <c> ] ]`. Note that the `a` token is lost.
I would expect `a b c` to be the tokens that the second argument collects.

jonsequitur:


src/System.CommandLine/Parsing/ParseResultVisitor.cs
```diff
@@ -256,6 +256,18 @@ private void ValidateAndConvertArgumentResults(IReadOnlyList<Argument> arguments
...
+                if (i > 0 && _argumentResults.Count > 1)
+                {
+                    var previousArgumentResult = _argumentResults[i - 1];
+
+                    var passedOnTokensCount = previousArgumentResult.PassedOnTokens?.Count;
+
+                    if (passedOnTokensCount > 0)
+                    {
+                        ShiftPassedOnTokensToNextResult(previousArgumentResult, _argumentResults[i], passedOnTokensCount);
+                    }
```
baronfel, line 268: This makes perfect sense, nice find!

Issue #1653, opened 2022-02-22
remove the legacy double dash behavior and `ParseResult.UnparsedTokens`


jonsequitur:
This functionality exists primarily to support the pre-existing behavior of the `dotnet` CLI, but that can likely be reimplemented using `Argument<string[]>`.

baronfel:
As a case study for users of this library, [here](https://github.com/dotnet/sdk/pull/23889/commits/3250361d1d3e39b5b3354dc20b21bdb8078186fb) is a commit where explicit -- behavior was superseded with an `Argument<string[]>`.

Tunduk:
@jonsequitur Hi!
`dotnet/sdk` team merged [PR](https://github.com/dotnet/sdk/pull/26761) with removing legacy double dash behavior.
If you don't mind I want to try to fix this issue.

jonsequitur:
Closed by #1806.

PR #1806, opened 2022-07-26
[1653] remove the legacy double dash behavior and ParseResult.UnparsedTokens

Tunduk:
This PR fixes an issue https://github.com/dotnet/command-line-api/issues/1653

### DIFF
--- a/src/System.CommandLine.Suggest/SuggestionDispatcher.cs
+++ b/src/System.CommandLine.Suggest/SuggestionDispatcher.cs
@@ -74,9 +74,8 @@ namespace System.CommandLine.Suggest
                 RegisterCommand,
                 CompleteScriptCommand
             };
-
+            root.TreatUnmatchedTokensAsErrors = false;
             Parser = new CommandLineBuilder(root)
-                     .EnableLegacyDoubleDashBehavior()
                      .UseVersionOption()
                      .UseHelp()
                      .UseParseDirective()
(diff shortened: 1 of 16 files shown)


src/System.CommandLine.Suggest/SuggestionDispatcher.cs
```diff
@@ -74,9 +74,8 @@ public SuggestionDispatcher(ISuggestionRegistration suggestionRegistration, ISug
                 RegisterCommand,
                 CompleteScriptCommand
             };
-
+            root.TreatUnmatchedTokensAsErrors = false;
```
jonsequitur, line 77: This looks like it will be a behavioral change. We shouldn't alter the parser configuration here.

src/System.CommandLine/Parsing/ParseOperation.cs
```diff
@@ -243,7 +235,7 @@ private void ParseRemainingTokens()
                 }
                 else if (foundEndOfArguments)
                 {
-                    AddCurrentTokenToUnparsed();
+                    AddCurrentTokenToUnmatched();
```
jonsequitur, line 238: I don't think this is the right behavior. This should probably not doing anything special at this point.

Tunduk: Yeah, you are right. The whole method become useless. I removed it.

jonsequitur (approved):

Tunduk:
@jonsequitur Thanks for you review!
Can we close the issue too?

jonsequitur:
Yes we can!

PR #1862, opened 2022-09-30
Fix IsOptional method for arguments

### LINKED ISSUE #1748: [BUG?] Reused arguments are considered optional
I see in the help builder that an argument used in multiple commands is flagged as optional in usage help

https://github.com/dotnet/command-line-api/blob/55dbf39074a7563e25e3409c9feb7afb8c8d2cc4/src/System.CommandLine/Help/HelpBuilder.cs#L332
```
            bool IsOptional(Argument argument) =>
                IsMultiParented(argument) ||
                argument.Arity.MinimumNumberOfValues == 0;
```
Is there a reason for this? We reuse arguments when it makes sense between commands, and they are still required.
I believe this is a bug unless I'm missing something

iBicha:
Fixes #1748 

This bug is causing confusion on what it optional and what it not. Reusing args between commands has nothing to do with them being optional or required.

### DIFF
--- a/src/System.CommandLine.Tests/Help/HelpBuilderTests.cs
+++ b/src/System.CommandLine.Tests/Help/HelpBuilderTests.cs
@@ -250,6 +250,34 @@ namespace System.CommandLine.Tests.Help
         }
 
         [Fact]
+        public void Usage_section_for_subcommand_shows_arguments_for_subcommand_and_parent_command_not_optional()
+        {
+            var arg = new Argument<string[]>
+            {
+                Name = "shared-args",
+                Arity = ArgumentArity.OneOrMore
+            };
+
+            var inner = new Command("inner", "command help")
+            {
+                arg
+            };
+            _ = new Command("outer", "command help")
+            {
+                inner,
+                arg
+            };
+
+            _helpBuilder.Write(inner, _console);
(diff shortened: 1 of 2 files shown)


iBicha:
cc @jonsequitur

jonsequitur:
Can you add a test that demonstrates the behavior?

iBicha:
> Can you add a test that demonstrates the behavior?

Mostly covered by other tests, but I added a case where the arg is shared, it should not be optional.

src/System.CommandLine.Tests/Help/HelpBuilderTests.cs
```diff
@@ -932,7 +932,38 @@ public void Command_arguments_with_default_values_that_are_enumerable_display_pi
             _console.ToString().Should().Contain(expected);
         }
 
+        [Fact]
+        public void Command_shared_arguments_with_one_or_more_arity_are_required()
```
jonsequitur, line 936:
```suggestion
        public void Command_shared_arguments_with_one_or_more_arity_are_displayed_as_being_required()
```

jonsequitur (approved):

jonsequitur (approved):

PR #1890, opened 2022-11-02
Make Option constructors protected

### LINKED ISSUE #1889: Cannot implement Option because the only constructors are internal
A [recent change](https://github.com/dotnet/command-line-api/commit/6c71bd788d9c8f8628787792d865b9eb683f851f#diff-72cee93846ef63ae468a9f0b8c214ec5989e9f7e71268505248d9834596f146aL30) in `2.0-beta4` removed the public constructors for `Option`. Because the constructors are declared with `internal` accessibility, 3rd party libraries and applications can no longer extend the abstract `Option` class.

The following constructors should be changed to `protected` accessibility to allow custom extensions of `Option`:
```csharp
internal Option(
    string name,
    string? description,
    Argument argument)
    : base(description)
{
    ...
}

internal Option(
    string[] aliases,
    string? description,
    Argument argument)
    : base(description)
{
    ...
}
```

tom-mckinney:
Change the accessibility of the `Option` constructors to `protected` so that it can be extended by 3rd party libraries

Resolves #1889

### DIFF
--- a/src/System.CommandLine/Option.cs
+++ b/src/System.CommandLine/Option.cs
@@ -19,7 +19,7 @@ namespace System.CommandLine
         private List<ValidateSymbolResult<OptionResult>>? _validators;
         private readonly Argument _argument;
 
-        internal Option(
+        protected Option(
             string name,
             string? description,
             Argument argument)
@@ -38,7 +38,7 @@ namespace System.CommandLine
             _argument = argument;
         }
 
-        internal Option(
+        protected Option(
             string[] aliases,
             string? description,
             Argument argument)


adamsitnik (changes requested):
Hello @tom-mckinney !

We don't expect our users to derive from `Option`, but from `Option<T>` which defines a bunch of public constructors:

https://github.com/dotnet/command-line-api/blob/cd7b452191629397ea3c1fbc6e7f6c0f4cf36dce/src/System.CommandLine/Option%7BT%7D.cs#L18-L64

I am going to close the PR. If there is any scenario which is currently impossible to implement with generic `Option`, please create an issue and tag me (I am in the middle of improving these APIs right now)

PR #1947, opened 2022-11-04
Rename ThrowIfInvalid to Validate, change thrown exception

### LINKED ISSUE #1908: Consider renaming `ThrowIfInvalid` to clarify the intent

### LINKED ISSUE #1909: Remove `CommandLineConfigurationException`
```C#
    // Is this something people actually need to catch?  If not, consider not making it public.
    // And if they do need to catch it, is it actually different enough from any existing exception to warrant a new type?
    public class CommandLineConfigurationException : Exception, System.Runtime.Serialization.ISerializable 
```

adamsitnik:
fixes #1908
fixes #1909


jonsequitur:
Is the name `Validate` consistent with other usages of that word? I see validation as typically referring to validation of user input. Other usage of the word in System.CommandLine prior to this refers only to validating input. Since valid input is expected and not exceptional, throwing an exception when a "validate" method is called might be confusing.

Jozkee:
I agree `Validate` may not be the best, there's also the property `Command.Validators`, so, `Validate()` may convey that you will run the validators for all commands and subcommands, and the method impl. doesn't appear to me to be doing that.

Is `ThrowIfInvalid` really necessary? Looks like is meant to avoid object cycles and disallow duplicate subcommands/options. Couldn't we just run this method implicitly when `Parser.Parse(...)` is called? Or is there ascenario where you don't want to validate prior to that?

jonsequitur:
> Is ThrowIfInvalid really necessary? Looks like is meant to avoid object cycles and disallow duplicate subcommands/options. Couldn't we just run this method implicitly when Parser.Parse(...) is called? Or is there ascenario where you don't want to validate prior to that?

Once you compile your code, the parser configuration is either valid or it's not, and it will never change, so performing this validation on every run is pointless and wasteful. The primary use case for the method is testing, but it's also potentially useful when dynamically assembling a parser, as some users are doing.

Using the word "test" in the name might be helpful in clarifying its intent.

Jozkee:
> Once you compile your code, the parser configuration is either valid or it's not, and it will never change, so performing this validation on every run is pointless and wasteful.

In that case, we can run this only once on `Parser.Parse()` :)

KathleenDollard:
`Validate` seems too short. I would like a two word name here: `ValidateCliStructure`? or perhaps `TestCliStructure`? `TestIfCliStructureIsValid`? (last is joke)

Jozkee:
> > Once you compile your code, the parser configuration is either valid or it's not, and it will never change, so performing this validation on every run is pointless and wasteful.
> 
> In that case, we can run this only once on Parser.Parse(). I haven't seen APIs on BCL added just for testing purposes but is possible that I just haven't seen many APIs.
> 
> Maybe this could be an internal method called only on #if DEBUG.

Spoke with @bartonjs about this, so it seems there is no way to say "I'm going to verify your inputs only when you're built in debug". He also suggested that there may be a way with source generators. IIRC @jonsequitur also menioned sourcegen. You can disregard my comment.

PR #1948, opened 2022-11-04
replace IInvocationResult  with Action<InvocationContext>

### LINKED ISSUE #1936: Consider using `Action<InvocationContext>` and removing `IInvocationResult`
```csharp
    // Usual "classes over interfaces" comment.
    // Definitely do not make a public interface that the library doesn't already provide at least three different implementations of (non-public ones count), otherwise we don't have confidence in the shape.
    public interface IInvocationResult 
    {
        void Apply(InvocationContext context);
    }
 
        // Where does this cancellation token come from?
        public System.Threading.CancellationToken GetCancellationToken();
    }
 
    // Even with a renamed concept of "middleware", I'm not sure a delegate feels like the right thing.
    // And what should a middleware do with "next"?  (Was the return type of the delegate supposed to be Task?)
    // I'm afraid this whole thing needs a full redesign (or, preferably, to be cut).
    public delegate void InvocationMiddleware(InvocationContext context, Func<InvocationContext,Task> next);
    
    // You are aware that callers can pass any value between int.MinValue and int.MaxValue, inclusive, right?
    public enum MiddlewareOrder
    {
        Default = 0,
        ErrorReporting = 1000,
        ExceptionHandler = -2000,
        Configuration = -1000,
    }
}
 
```

adamsitnik:
fixes #1936

@jonsequitur I don't know if redesigned invocation is still going to keep the concept of `IInvocationResult`. If not, this PR can be rejected, as it does not make sense to break the customers twice (now by replacing the interface with delegate, and leter when we remove the delegate too)


jonsequitur:
I don't anticipate that the concept of a result will go away with the invocation redesign, so this is still a good simplification.

jonsequitur (approved):

Issue #1900, opened 2022-11-02
Rename `LegalFileNamesOnly` to include a verb


jonsequitur:
```C#
        // Missing a verb.  "PermitLegalFileNamesOnly", maybe?
        public static TArgument LegalFileNamesOnly<TArgument>(this TArgument argument);
        public static TArgument LegalFilePathsOnly<TArgument>(this TArgument argument);
```

Issue #1898, opened 2022-11-02
Rename parameters to align on the word "completion"


jonsequitur:
```C#
        // Why is this one "values" and the other two are "complete"?
        // Since the method is called "AddCompletions" I'd expect things to be called a "completion".
        public static TArgument AddCompletions<TArgument>(this TArgument argument, string[] values);
        public static TArgument AddCompletions<TArgument>(this TArgument argument, Func<CompletionContext,IEnumerable<string>> complete);
        public static TArgument AddCompletions<TArgument>(this TArgument argument, CompletionDelegate complete);
```

Issue #1896, opened 2022-11-02
Move extension methods into the extended type


jonsequitur:
```C#
    // Guidline: DO NOT put extension methods in the same namespace as the extended type unless the intention is to add methods to interfaces, for generic type restriction, or for dependency management.
    //
    // The extension methods on TArgument should all move to be instance methods on the Argument class.  Since they were implemented as extension methods, I presume they would be non-virtual.
    // The ExistingOnly extension methods are "generic type restriction", so they're OK.  
```

PR #1951, opened 2022-11-08
Moves and renames

Jozkee:
Fixes https://github.com/dotnet/command-line-api/issues/1900
Fixes https://github.com/dotnet/command-line-api/issues/1899
Fixes https://github.com/dotnet/command-line-api/issues/1898
Fixes https://github.com/dotnet/command-line-api/issues/1897
Fixes https://github.com/dotnet/command-line-api/issues/1896


src/System.CommandLine.ApiCompatibility.Tests/ApiCompatibilityApprovalTests.System_CommandLine_api_is_not_changed.approved.txt
```diff
@@ -32,19 +40,11 @@ System.CommandLine
...
-    public static Argument<System.IO.FileInfo> ExistingOnly(this Argument<System.IO.FileInfo> argument)
-    public static Argument<System.IO.DirectoryInfo> ExistingOnly(this Argument<System.IO.DirectoryInfo> argument)
-    public static Argument<System.IO.FileSystemInfo> ExistingOnly(this Argument<System.IO.FileSystemInfo> argument)
-    public static Argument<T> ExistingOnly<T>(this Argument<T> argument)
-    public static TArgument FromAmong<TArgument>(this TArgument argument, System.String[] values)
-    public static TArgument LegalFileNamesOnly<TArgument>(this TArgument argument)
-    public static TArgument LegalFilePathsOnly<TArgument>(this TArgument argument)
-    public static ParseResult Parse(this Argument argument, System.String commandLine)
-    public static ParseResult Parse(this Argument argument, System.String[] args)
+  public static class ArgumentValidationExtensions
```
jonsequitur, line 43: Would `ArgumentValidation` be a more standard class name?

Jozkee: @bartonjs may know better, but I think is a common pattern to use `Extensions` suffix for extension classes.

bartonjs: Adding the Extension suffix is common, but is actually an anti-pattern.  You should, ideally, name the containing type on the basis everything in it gets invoked as a static at the callsite, then can optionally make things be extension methods.

Also, extension methods themselves are generally an anti-pattern.

* If the type is an interface, see the design guidelines that basically suggest "don't add new interfaces, they're always full of regret".
* If the type isn't generic, you shouldn't have any extension methods for it in the same assembly.
* If the type is generic, and you have members that only apply for a smaller subset of Ts, those can/have-to be generic.
* If the type is generic, expects to be instantiated with nearly value-type under the sun, and needs to be small (e.g. it's Span) then extension methods tend to be preferred so that the T-substitutions are reduced to only the combination of T and method that got used.

So, basically, System.CommandLine should have almost no extension methods.

Jozkee: > If the type is generic, and you have members that only apply for a smaller subset of Ts, those can/have-to be generic.

That's the case here. 

I prefer the `Extensions` suffix just because it glimpses that is a class meant to contain extensions but is not a strong opinion. Thanks, @bartonjs.

jonsequitur (approved):

PR #1953, opened 2022-11-08
default completions should be used only when user did not provide any

### LINKED ISSUE #1884: Cannot replace default completions of Option<T> where T is enum
When `T` is an enum type, `Option<T>` gets a default completion source that returns the names of the defined enum values. If I call OptionExtensions.AddCompletions to add a completion callback that returns some of those names but not all, that has no effect because the default completion source returns all of them anyway and Argument.GetCompletions discards my completion items as duplicates, even if they contain documentation and the default completion items don't. Because the Option.Argument property is internal, I cannot call option.Argument.Completions.Clear() to delete the default completion source before I add mine.

Possibly I could hack around this by deriving my own class from `Option<T>` and overriding GetCompletions there.

## To reproduce

```XML
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net6.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="System.CommandLine" Version="2.0.0-beta4.22272.1" />
  </ItemGroup>

</Project>
```

```C#
namespace EnumCompletion;

using System.CommandLine;
using System.CommandLine.Completions;

class Program
{
    {
        var option = new Option<DayOfWeek>("-d")
            .AddCompletions(CompleteDayOfWeek);
        foreach (var completion in option.GetCompletions())
        {
            Console.WriteLine($"{completion.Label}: {completion.Documentation}");
        }
    }

    static IEnumerable<CompletionItem> CompleteDayOfWeek(CompletionContext context)
    {
        Console.Error.WriteLine("CompleteDayOfWeek was called.");

        // Omit Saturday and Sunday.
        return new[]
        {
           new CompletionItem("Monday", sortText: "1", documentation: "maanantai"),
           new CompletionItem("Tuesday", sortText: "2", documentation: "tiistai"),
           new CompletionItem("Wednesday", sortText: "3", documentation: "keskiviikko"),
           new CompletionItem("Thursday", sortText: "4", documentation: "torstai"),
           new CompletionItem("Friday", sortText: "5", documentation: "perjantai"),
        };
    }
}
```

## Actual output

```
CompleteDayOfWeek was called.
Friday:
Monday:
Saturday:
Sunday:
Thursday:
Tuesday:
Wednesday:
```

## Expected output

```
CompleteDayOfWeek was called.
Monday: maanantai
Tuesday: tiistai
Wednesday: keskiviikko
Thursday: torstai
Friday: perjantai
```

adamsitnik:
My proposal for fixing #1884: use the default completions only when user has not provided their own.

Cons: we lose the possibility to extend the default completions with custom ones.

fixes #1884

cc @KalleOlaviNiemitalo

### DIFF
--- a/src/System.CommandLine.Tests/CompletionTests.cs
+++ b/src/System.CommandLine.Tests/CompletionTests.cs
@@ -888,7 +888,25 @@ namespace System.CommandLine.Tests
         }
 
         [Fact]
-        public void Default_completions_can_be_appended_to()
+        public void Default_completions_are_not_used_when_user_provides_their_own()
+        {
+            var argument = new Argument<DayOfWeek>();
+            argument.Completions.Add(new[] { "mon", "tues", "wed", "thur", "fri", "sat", "sun" });
+            var command = new Command("the-command")
+            {
+                argument
+            };
+
+            var completions = command.Parse("the-command s")
+                                     .GetCompletions();
+
+            completions.Select(item => item.Label)
+                       .Should()
+                       .BeEquivalentTo("sat", "sun", "tues");
+        }
+
+        [Fact]
+        public void Default_completions_can_not_be_appended_to()
         {
             var command = new Command("the-command")
             {
@@ -906,13 +924,8 @@ namespace System.CommandLine.Tests
                 .Should()
                 .BeEquivalentTo(
                     "sat",
-                    nameof(DayOfWeek.Saturday),
                     "sun", 
-                    nameof(DayOfWeek.Sunday),
-                    "tues",
-                    nameof(DayOfWeek.Tuesday),
-                    nameof(DayOfWeek.Thursday),
-                    nameof(DayOfWeek.Wednesday));
+                    "tues");
         }
 
         [Fact]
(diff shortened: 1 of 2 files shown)


jonsequitur:
The behavior seems unintuitive. If I know for example that `Option<DayOfWeek>` returns seven well-known values, I would not expect calling `Completions.Add` to replace them.

Jozkee:
>  If I know for example that Option<DayOfWeek> returns seven well-known values, I would not expect calling Completions.Add to replace them.

That's what @KalleOlaviNiemitalo describes as expected in the [linked issue](https://github.com/dotnet/command-line-api/issues/1884). I was first thinking that a possibility could be to only use defaults as long as users don't `Add` their own completions.

If that's not good, we could also just expose a method to `Clear()` the completions.

adamsitnik:
I've opened https://github.com/dotnet/command-line-api/issues/1972 to discuss better solution. Thank you all for your feedback!

Issue #1891, opened 2022-11-02
API review issues


jonsequitur:
- [ ] Rename `Argument`, `Option`, and `Command` to disambiguate from other single-word types
```C#
    // Consider adding a disambiguator to this type name, so it doesn't feel like "ArgumentException" is related to this type.
    public abstract class Argument : Symbol, IValueDescriptor, ICompletionSource 
```

- [ ] Consider a more commonly-known word than "arity"
```C#
        // I know what arity is, but only from working on a compiler project.  Do "real people" know what it means?
        public ArgumentArity Arity { get; set; }
```

- [ ] Rename `getDefaultValue` parameter to  `defaultValueFactory`
```C#
        // Feels like `defaultValueFactory` is a better name for the parameter, given the method name.
        public void SetDefaultValueFactory(Func<object> getDefaultValue);
        public void SetDefaultValueFactory(Func<ArgumentResult,object> getDefaultValue);
```

- [ ] NEEDS DISCUSSION: Does an argument with no name make sense?
```C#
    public class Argument<T> : Argument, IValueDescriptor<T>, IValueDescriptor, ICompletionSource 
    {
        // Does an argument with no name make sense?
        public Argument();
```

- [ ] Move extension methods into the extended type 
```C#
    // Guidline: DO NOT put extension methods in the same namespace as the extended type unless the intention is to add methods to interfaces, for generic type restriction, or for dependency management.
    //
    // The extension methods on TArgument should all move to be instance methods on the Argument class.  Since they were implemented as extension methods, I presume they would be non-virtual.
    // The ExistingOnly extension methods are "generic type restriction", so they're OK.  
```

- [ ] Rename `ExistingOnly` 
```C#
    // I'm not sold on the naming around [ExistingOnly], because
    // * Guideline: DO give methods names that are verbs or verb phrases.
    // * Guideline: CONSIDER naming the type that holds extension methods for itsfunctionality—for example, use “Routing” instead of “[ExtendedType]Extensions.”
    //
    // The only existing "ExistingOnly" in the BCL is System.Security.Cryptography.X509Certificates.OpenFlags.OpenExistingOnly, which is an enum member.
    // "AcceptExistingOnly".  "RequirePathExists"
    // Static methods should also consider their static invocation model:  ArgumentExtensions.ExistingOnly(inputFile)... that has no meaning to me.
[...]

PR #1954, opened 2022-11-08
introduce CompletionContext.Empty, remove Symbol.GetCompletions(void)

adamsitnik:
introduce static `CompletionContext.Empty`:

* cache it so it does not get created more than once
* make it public so it can be reused by our users who write tests (https://github.com/dotnet/command-line-api/issues/1891#issuecomment-1307575196)
* use it places where it can improve performance (including benchmarks, which should not include this allocation as this is not what they are supposed to benchmark)

remove Symbol.GetCompletions(void): mostly to simplify `Symbol`

@jonsequitur if we don't want to delete `Symbol.GetCompletions(void)` we should keep caching the empty completion context (perf gains in `HelpBuilder` and `ParseResultVisitor`)

### DIFF
--- a/src/System.CommandLine/Symbol.cs
+++ b/src/System.CommandLine/Symbol.cs
@@ -80,10 +80,6 @@ namespace System.CommandLine
         /// <summary>
         /// Gets completions for the symbol.
         /// </summary>
-        public IEnumerable<CompletionItem> GetCompletions() => 
-            GetCompletions(CompletionContext.Empty());
-
-        /// <inheritdoc />
         public abstract IEnumerable<CompletionItem> GetCompletions(CompletionContext context);
 
         /// <inheritdoc/>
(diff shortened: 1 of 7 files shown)


src/System.CommandLine/Symbol.cs
```diff
@@ -80,10 +80,6 @@ public IEnumerable<Symbol> Parents
         /// <summary>
         /// Gets completions for the symbol.
         /// </summary>
-        public IEnumerable<CompletionItem> GetCompletions() => 
-            GetCompletions(CompletionContext.Empty());
```
Jozkee, line 84: This feels like an unnecessary breaking change to me.

Jozkee (commented):
> if we don't want to delete Symbol.GetCompletions(void) we should keep caching the empty completion context (perf gains in HelpBuilder and ParseResultVisitor)

+1 to not deleting `Symbol.GetCompletions()`. Otherwise; LGTM.

src/System.CommandLine/Symbol.cs
adamsitnik: > This feels like an unnecessary breaking change to me.

The main point of my changes is simplification of the APIs which should help us get them approved and moved into BCL.

One of the users raised a concern about removing this method: https://github.com/dotnet/command-line-api/issues/1891#issuecomment-1307575196 but by exposing `CompletionContext.Empty` we are going to make it testable.

Jozkee: Ok that's fair, if is proven later on that this method is valuable, we can bring it back.

Jozkee (approved):

PR #1959, opened 2022-11-10
Remove Argument.AllowedValues, use Validator to implement the logic

adamsitnik:
follow up to https://github.com/dotnet/command-line-api/issues/1891#issuecomment-1310520011


src/System.CommandLine/Parsing/ArgumentResult.cs
```diff
@@ -83,7 +83,7 @@ public void OnlyTake(int numberOfTokens)
 
                 if (!string.IsNullOrWhiteSpace(ErrorMessage))
                 {
-                    return new ParseError(ErrorMessage!, this);
+                    return new ParseError(ErrorMessage!, Parent is OptionResult option ? option : this);
```
adamsitnik, line 86: This change is not obvious. Before this change, the `SymbolResult` of `ParseError` for an Argument with unmatched tokens owned by an Option was the `OptionResult`, not `ArgumentResult`:

https://github.com/dotnet/command-line-api/blob/0adeb26f4ce52c454280d9f3a5d6975e5d081fec/src/System.CommandLine/Parsing/ParseResultVisitor.cs#L509

but this was true only "non-custom errors" (user defined, coming from validators):

https://github.com/dotnet/command-line-api/blob/0adeb26f4ce52c454280d9f3a5d6975e5d081fec/src/System.CommandLine/Parsing/ParseResultVisitor.cs#L510

Without this change, one test was failing:

```log
  Failed System.CommandLine.Tests.ParseDiagramTests.Parse_result_diagram_displays_unmatched_tokens [283 ms]
  Error Message:
   Expected string to be
"[ command ![ -x <ar> ] ]", but
"[ command [ -x !<ar> ] ]" differs near "[ -" (index 10).
```

Since we hide the `Option.Argument` from the user, I think it's reasonable to use the `OptionResult` in such cases for all validations.

jonsequitur:
If there's no test for case-insensitive enum parsing, we should definitely add one.

src/System.CommandLine/Argument.cs
```diff
@@ -105,14 +103,15 @@ private protected override string DefaultName
...
+        internal IReadOnlyList<Action<ArgumentResult>> Validators
+            => _validators is null ? Array.Empty<Action<ArgumentResult>>() : _validators;
 
         /// <summary>
         /// Adds a custom validator to the argument. Validators can be used
         /// to provide custom errors based on user input.
         /// </summary>
         /// <param name="validate">The action to validate the parsed argument.</param>
-        public void AddValidator(Action<ArgumentResult> validate) => Validators.Add(validate);
+        public void AddValidator(Action<ArgumentResult> validate) => (_validators ??= new()).Add(validate);
```
Jozkee, line 114: There are other places that call Validator.Add, which would now add the Action to the Array.Empty<Action<ArgumentResult>>.

https://github.com/dotnet/command-line-api/blob/350a618ab44932c568ae2392d7a571281baed59a/src/System.CommandLine/Argument%7BT%7D.cs#L215

https://github.com/dotnet/command-line-api/blob/350a618ab44932c568ae2392d7a571281baed59a/src/System.CommandLine/OptionValidation.cs#L21

adamsitnik:
I've addressed all the feedback, I am going to merge this PR now to unblock other PRs.

> AcceptOnlyFromAmong is case-sensitive, unlike the Enum.Parse call in ArgumentConverter

Please create an issue for that. We can just extend `AcceptOnlyFromAmong ` with an `StringComparison comparisonType` argument and let the users decide

PR #1964, opened 2022-11-14
fluent APIs should return least restrictive result

### LINKED ISSUE #1963: All fluent APIs defined by Option<T> and Argument<T> should return generic value
While working on unblocking SDK with latest S.CL update in https://github.com/dotnet/sdk/pull/29012 I've realized that in #1951 we have moved argument and option extension methods to the types themselves, but unfortunately, we have also made these methods non-generic. It does not allow to re-use the returned value when we need a generic value.

Example:

```cs
public static readonly Argument<string> CmdPackageArgument = new Argument<string>(LocalizableStrings.CmdPackage)
{
    Description = LocalizableStrings.CmdPackageDescription
}.AddCompletions((context) => QueryNuGet(context.WordToComplete).Select(match => new CompletionItem(match)));

// currently `AddCompletions` returns an `Argument` rather than `Argument<string>`
```

```cmd
 Cannot implicitly convert type 'System.CommandLine.Argument' to 'System.CommandLine.Argument<string>'. An explicit
 conversion exists
```

cc @Jozkee @jonsequitur

adamsitnik:
fixes #1963


src/System.CommandLine/Option{T}.cs
```diff
@@ -57,8 +61,96 @@ public class Option<T> : Option, IValueDescriptor<T>
...
+            string[] aliases,
+            string? description,
+            Argument<T> argument)
+            : base(aliases, description)
+        {
+            argument.AddParent(this);
+            _argument = argument;
+        }
+
+        internal sealed override Argument Argument => _argument;
```
adamsitnik, line 88: By making this property `sealed` we enforce all `Option<T>` derived types to always pass an instance of `Argument<T>` to the `ctor`. By doing that, we are always sure that  `_argument` (the field) and `Argument` (the property) always refer to the same instance of an object. In the past it was not always a thing?

Another benefit is that when the derived types are being created, only one `Argument<T>` is being allocated.

src/System.CommandLine/Help/HelpOption.cs
```diff
@@ -9,7 +9,7 @@ internal class HelpOption : Option<bool>
         private string? _description;
 
         public HelpOption(string[] aliases, Func<LocalizationResources> getLocalizationResources)
-            : base(aliases)
+            : base(aliases, null, new Argument<bool> { Arity = ArgumentArity.Zero })
```
Jozkee, line 12: Could `new Argument<bool> { Arity = ArgumentArity.Zero }` be in a static and be re-used here and on VerisonOption?

Jozkee (approved):
Otherwise, LGTM. Sorry for breaking the SDK 😅.

src/System.CommandLine/Option.cs
```diff
@@ -58,15 +46,12 @@ public abstract class Option : IdentifierSymbol, IValueDescriptor
...
-
-            argument.AddParent(this);
-            _argument = argument;
         }
 
         /// <summary>
         /// Gets the <see cref="Argument">argument</see> for the option.
         /// </summary>
-        internal virtual Argument Argument => _argument;
+        internal abstract Argument Argument { get; }
```
Jozkee, line 54: This will break custom implementations of `Option`, we don't have any in S.CL source code, expect for Option<T>, but there may be users that do have them. 

Is that not an issue?

adamsitnik: @Jozkee this method is `internal` so it won't be an issue

src/System.CommandLine/Help/HelpOption.cs
adamsitnik: IIRC I've tried it in the past (when I was working on SCL perf) and it was impossible as some of the methods were allowing for mutation, but I am not 100% sure if this is the case now.

Once we are done with API-redesign and re-layering we are going to profile the startup scenario again and if it pops up, try to cache it. But initializing static fields also has a cost, so this will have to be carefully verified.

PR #1969, opened 2022-11-15
Move version option out of middleware

jonsequitur:


src/System.CommandLine/Builder/CommandLineBuilderExtensions.cs
```diff
@@ -643,6 +624,7 @@ public static CommandLineBuilder UseHelp(this CommandLineBuilder builder, int? m
             builder.VersionOption = versionOption;
             builder.Command.AddOption(versionOption);
 
+#if false
```
Jozkee, line 627:
```suggestion
```

src/System.CommandLine/Builder/CommandLineBuilderExtensions.cs
```diff
@@ -661,6 +643,7 @@ public static CommandLineBuilder UseHelp(this CommandLineBuilder builder, int? m
                     await next(context);
                 }
             }, MiddlewareOrderInternal.VersionOption);
+#endif
```
Jozkee, line 646:
```suggestion
```

Same for the other `#if` on line 670.

src/System.CommandLine/Builder/CommandLineBuilderExtensions.cs
Jozkee: Actually I misunderstood, this is disabling the code, right?

Do we want to keep it that way or remove the whole block?

Keboo: I would propose removal of the whole block

adamsitnik (approved):
LGTM! The allocations have dropped nicely!

|        Method |      Args |   before  |     after |
|-------------- |---------- |----------:|----------:|
|  DefaultsSync | String[0] |  10,584 B |  10,264 B |
| DefaultsAsync | String[0] |  10,584 B |  10,264 B |
|   MinimalSync | String[0] |   3,256 B |   3,264 B |
|  MinimalAsync | String[0] |   3,256 B |   3,264 B |
|  DefaultsSync | String[4] |  12,080 B |  11,760 B |
| DefaultsAsync | String[4] |  12,080 B |  11,760 B |
|   MinimalSync | String[4] |   4,752 B |   4,760 B |
|  MinimalAsync | String[4] |   4,752 B |   4,760 B |

PR #1989, opened 2022-12-12
Expose Command collection properties as mutable IList, remove AddCommand, AddOption and AddArugment methods

adamsitnik:
This is very similar to #1987, but since adding a new Symbol to Command requires setting a parent I had to implement a custom `IList<T>` that sets the parent on every add.


src/System.CommandLine/ChildList.cs
```diff
@@ -0,0 +1,65 @@
...
+
+using System.Collections;
+using System.Collections.Generic;
+
+namespace System.CommandLine
+{
+    /// <summary>
+    /// a wrapper of List<typeparamref name="T"/> that sets parent for every added element
+    /// </summary>
+    internal sealed class ChildList<T> : IList<T> where T : Symbol
```
Jozkee, line 12: Did you consider this?
```suggestion
    internal sealed class ChildList : IList<Symbol>
```

src/System.CommandLine/Command.cs
```diff
@@ -113,26 +82,32 @@ public void AddOption(Option option)
         public void AddGlobalOption(Option option)
```
Jozkee, line 82: Is weird that now that you are removing `AddOption` we keep `AddGlobalOption`. If we follow the pattern this should be named `AddGlobal`, no? But that sounds weird.

Not asking for a change, just wanted to point that out.

src/System.CommandLine.ApiCompatibility.Tests/ApiCompatibilityApprovalTests.System_CommandLine_api_is_not_changed.approved.txt
```diff
@@ -49,19 +49,14 @@ System.CommandLine
     public static Argument<T> AcceptExistingOnly<T>(this Argument<T> argument)
   public class Command : IdentifierSymbol, System.Collections.Generic.IEnumerable<Symbol>, System.Collections.IEnumerable
     .ctor(System.String name, System.String description = null)
-    public System.Collections.Generic.IReadOnlyList<Argument> Arguments { get; }
```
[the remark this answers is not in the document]
adamsitnik: They were mutable anyway, but via methods like `AddArguments` and `Add(Argument argument)`. Now we just expose the list and let the user add and delete if they want to

src/System.CommandLine/ChildList.cs
adamsitnik: It would be impossible to enforce the right types without perf overhead. Example:

```cs
Option<bool> option = new ("--some);
RootCommand root = new ();
// in the line below Arguments would be a list of Symbols, which would allow for adding not just arguments but also options and commands
root.Arguments.Add(option); // no compilation error, at runtime we would later find out that it's not an Argument
```

src/System.CommandLine/Command.cs
adamsitnik: That is on my TODO list, we just need to find a good name for "global" options. In the near future it might look like this:

```cs
Option<bool> help = new ("--help);
help.IsCommon = true;
command.Options.Add(help);
```

adamsitnik:
I believe I've answered all the questions, merging.

PR #1990, opened 2022-12-12
First name and aliases cleanup

adamsitnik:
- Multiple types derived from `Symbol` were defining it's own name field. I've removed these and now all of them are just using `_name` defined by `Symbol`
- As agreed with Jon (not tagging him as he is on vacations) removing `Option.HasAliasIgnoringPrefix` as it's not for general purpose
- making `GetLongestAlias` part of base type and re-using it where possible
- encapsulating `_aliases` field, as it will most likely change once we introduce the concept of a hidden alias

### DIFF
--- a/src/System.CommandLine/Binding/ArgumentConversionResult.cs
+++ b/src/System.CommandLine/Binding/ArgumentConversionResult.cs
@@ -56,7 +56,7 @@ namespace System.CommandLine.Binding
             if (argument.FirstParent?.Symbol is IdentifierSymbol identifierSymbol &&
                 argument.FirstParent.Next is null)
             {
-                var alias = identifierSymbol.Aliases.First();
+                var alias = identifierSymbol.GetLongestAlias(removePrefix: false);
 
                 switch (identifierSymbol)
                 {
--- a/src/System.CommandLine/IdentifierSymbol.cs
+++ b/src/System.CommandLine/IdentifierSymbol.cs
@@ -12,7 +12,6 @@ namespace System.CommandLine
     public abstract class IdentifierSymbol : Symbol
     {
         private protected readonly HashSet<string> _aliases = new(StringComparer.Ordinal);
-        private string? _specifiedName;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="IdentifierSymbol"/> class.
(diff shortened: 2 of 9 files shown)


Keboo (approved):
:shipit:

src/System.CommandLine/IdentifierSymbol.cs
```diff
@@ -42,19 +42,18 @@ protected IdentifierSymbol(string name, string? description = null)
...
-            get => _specifiedName ??= DefaultName;
             set
             {
-                if (_specifiedName is null || !string.Equals(_specifiedName, value, StringComparison.Ordinal))
+                if (_name is null || !string.Equals(_name, value, StringComparison.Ordinal))
                 {
                     AddAlias(value);
 
-                    if (_specifiedName is { })
+                    if (_name is { })
```
Jozkee, line 51: Does this checks if _name is not null?
```suggestion
                    if (_name != null)
```

src/System.CommandLine/Binding/ArgumentConversionResult.cs
```diff
@@ -56,7 +56,7 @@ private ArgumentConversionResult(Argument argument)
             if (argument.FirstParent?.Symbol is IdentifierSymbol identifierSymbol &&
                 argument.FirstParent.Next is null)
             {
-                var alias = identifierSymbol.Aliases.First();
+                var alias = identifierSymbol.GetLongestAlias(removePrefix: false);
```
Jozkee, line 59: This wasn't really getting the longest alias but the first one.

src/System.CommandLine/Parsing/StringExtensions.cs
```diff
@@ -32,7 +32,7 @@ internal static string RemovePrefix(this string alias)
                        : alias;
         }
 
-        internal static int GetPrefixLength(this string alias)
+        private static int GetPrefixLength(this string alias)
```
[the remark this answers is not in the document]
adamsitnik: `DragonFruit` will be deleted soon and replaced by new source-generator based solution

src/System.CommandLine/IdentifierSymbol.cs
```diff
@@ -30,7 +30,7 @@ protected IdentifierSymbol(string? description = null)
         /// <param name="description">The description of the symbol, which is displayed in command line help.</param>
         protected IdentifierSymbol(string name, string? description = null) 
         {
-            Name = name;
+            Name = name ?? throw new ArgumentNullException(nameof(name));
```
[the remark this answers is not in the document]
adamsitnik: The `ctor` does not allow for nullable `name` (`description` can be null). If we move this check here the derived types don't need to perform it.

src/System.CommandLine/IdentifierSymbol.cs
adamsitnik: It does and I also believe that it's very unintuitive way of checking for `null` ;)

src/System.CommandLine/Binding/ArgumentConversionResult.cs
adamsitnik: It was inconsistent. As you can see other places in the source code were using longest option alias to get its default alias. Here we were using the first one but it was a bug (at least in my opinion).

adamsitnik: Option.Name should not contain the prefix, while here we care about prefix.

Jozkee (approved):

PR #1991, opened 2022-12-13
make CompletionItem implement IEquatable<CompletionItem>

adamsitnik:


Jozkee (approved):

adamsitnik:
> What is the rationale for this change?

It helps to avoid casting and `EqualityComparer<CompletionItem>.Default` returns `GenericEqualityComparer` rather than `ObjectEqualityComparer` (better perf when used in dictionaries/hashsets).

PR #1994, opened 2022-12-15
Fluent APIs are not allowed in non-builder types

adamsitnik:
Based on the feedback from @bartonjs

### DIFF
--- a/src/System.CommandLine.Tests/CompletionTests.cs
+++ b/src/System.CommandLine.Tests/CompletionTests.cs
@@ -972,9 +972,9 @@ namespace System.CommandLine.Tests
 
         private static Option<string> CreateOptionWithAcceptOnlyFromAmong(string name, params string[] values)
         {
-            Option<string> argument = new(name);
-            argument.AcceptOnlyFromAmong(values);
-            return argument;
+            Option<string> option = new(name);
+            option.AcceptOnlyFromAmong(values);
+            return option;
         }
     }
 }
--- a/src/System.CommandLine.Tests/CompletionTests.cs
+++ b/src/System.CommandLine.Tests/CompletionTests.cs
@@ -698,15 +698,15 @@ namespace System.CommandLine.Tests
             {
                 new Command("one")
                 {
-                    new Argument<string>().AcceptOnlyFromAmong("one-a", "one-b", "one-c")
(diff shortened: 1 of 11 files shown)


src/System.CommandLine.Tests/CompletionTests.cs
```diff
@@ -968,5 +962,19 @@ public void When_option_completions_are_available_then_they_are_suggested_when_a
...
+            Argument<string> argument = new();
+            argument.AcceptOnlyFromAmong(values);
+            return argument;
+        }
+
+        private static Option<string> CreateOptionWithAcceptOnlyFromAmong(string name, params string[] values)
+        {
+            Option<string> argument = new(name);
+            argument.AcceptOnlyFromAmong(values);
+            return argument;
```
Jozkee, line 977: nit
```suggestion
            Option<string> option = new(name);
            option.AcceptOnlyFromAmong(values);
            return option;
```

src/System.CommandLine.Tests/OptionTests.cs
```diff
@@ -372,10 +372,8 @@ public void Option_of_enum_can_limit_enum_members_as_valid_values()
         [Fact]
         public void Option_of_T_fluent_APIs_return_Option_of_T()
```
Jozkee, line 373: You might as well remove the test

Jozkee (approved):
> Should the ArgumentValidation.AcceptExistingOnly and OptionValidation.AcceptExistingOnly methods likewise return void?

Up to Adam's choice, but that could be addressed in a follow-up PR if required.

adamsitnik:
> Should the ArgumentValidation.AcceptExistingOnly and OptionValidation.AcceptExistingOnly methods likewise return void?

That is a very good question. To be honest I don't like `AcceptLegalFilePathsOnly` and `AcceptLegalFileNamesOnly` as they don't restrict the argument/option type in any way. Example:

```cs
Option<bool> flag = new ("--flag");
flag.AcceptLegalFilePathsOnly(); // it should work only with files, not things like boolean
```

`ArgumentValidation.AcceptExistingOnly` and `OptionValidation.AcceptExistingOnly ` restrict the types, but our BCL guidelines are against introducing extension methods for types that we own (they should be just instance methods). When @bartonjs and @jonsequitur are back from holidays I am going to discuss it with them.

PR #1995, opened 2022-12-15
Lambdas cleanup

adamsitnik:
* don't create unnecessary lambdas (based on feedback from @KalleOlaviNiemitalo)
* Option ctor simplification (Argument ctor will throw for null anyway)
* avoid closure allocations

### DIFF
--- a/src/System.CommandLine/Argument{T}.cs
+++ b/src/System.CommandLine/Argument{T}.cs
@@ -39,7 +39,7 @@ namespace System.CommandLine
             Func<T> defaultValueFactory, 
             string? description = null) : this(name, description)
         {
-            SetDefaultValueFactory(() => defaultValueFactory());
+            SetDefaultValueFactory(defaultValueFactory);
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/System.CommandLine/Argument{T}.cs
```diff
@@ -210,10 +210,10 @@ void UnrecognizedArgumentError(ArgumentResult argumentResult)
         /// <returns>The configured argument.</returns>
         public Argument<T> AcceptLegalFilePathsOnly()
         {
-            var invalidPathChars = Path.GetInvalidPathChars();
```
Jozkee, line 213: Wasn't this meant to cache `Path.GetInvalidPathChars()` which returns a new char[] on each call?

adamsitnik: What we had was:

1. calling `Path.GetInvalidPathChars()` for every `AcceptLegalFilePathsOnly` call
2. storing it's result in local variable (not a static field of the class, which could allow for caching the value)
3. allocating a closure (a lambda with a state)

What I am offering here is lack of closure allocation. I would prefer to avoid caching the result of `Path.GetInvalidPathChars()` in a static field as long as profiling don't prove that it's needed (adding static fields can hurt startup scenarios, it's not always a free lunch)

Jozkee (approved):

PR #1998, opened 2022-12-20
remove default Parser ctor

adamsitnik:


Jozkee (approved):
Assuming this is just to simplify API surface, LGTM.

PR #1999, opened 2022-12-20
move cycle detection logic to Symbol.AddParent

adamsitnik:
as it should be very cheap

related to #1909

### DIFF
--- a/src/System.CommandLine/Symbol.cs
+++ b/src/System.CommandLine/Symbol.cs
@@ -41,6 +41,24 @@ namespace System.CommandLine
         
         internal void AddParent(Symbol symbol)
         {
+            if (ReferenceEquals(this, symbol))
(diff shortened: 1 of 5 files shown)


src/System.CommandLine/Symbol.cs
```diff
@@ -54,6 +72,8 @@ internal void AddParent(Symbol symbol)
                 }
                 current.Next = new ParentNode(symbol);
             }
+
+            static void Throw(Symbol symbol) => throw new ArgumentException($"Cycle detected in symbol tree. Symbol '{symbol.Name}' would be its own ancestor.");
```
jonsequitur, line 76: The message should maybe say "Command ____ would be its own ancestor" as it should be impossible for this to happen with other symbol types.

PR #2001, opened 2022-12-20
Reducing allocations in parsing

adamsitnik:
While reading parsing logic I found plenty of places where we could remove some small allocations.


src/System.CommandLine/ParseResult.cs
```diff
@@ -99,7 +100,7 @@ public class ParseResult
         /// Gets the directives found while parsing command line input.
         /// </summary>
         /// <remarks>If <see cref="CommandLineConfiguration.EnableDirectives"/> is set to <see langword="false"/>, then this collection will be empty.</remarks>
-        public IReadOnlyDictionary<string, IReadOnlyList<string>> Directives { get; }
+        public IReadOnlyDictionary<string, IReadOnlyList<string>> Directives => _directives ??= new ();
```
adamsitnik, line 103: in the future we are going to use `IReadOnlyDictionary<TKey, TValue>.Empty` introduced very recently in .NET 8: https://github.com/dotnet/runtime/issues/76028

src/System.CommandLine/Invocation/TypoCorrection.cs
```diff
@@ -23,9 +23,10 @@ public TypoCorrection(int maxLevenshteinDistance)
 
         public void ProvideSuggestions(ParseResult result, IConsole console)
         {
-            for (var i = 0; i < result.UnmatchedTokens.Count; i++)
+            var unmatchedTokens = result.UnmatchedTokens;
```
adamsitnik, line 26: we need to store the result, because currently it's always allocating: https://github.com/dotnet/command-line-api/blob/905cb0b525a97ba753026954b4551fecd28375bd/src/System.CommandLine/ParseResult.cs#L118

src/System.CommandLine/Parsing/CommandArgumentNode.cs
```diff
@@ -10,10 +12,7 @@ internal class CommandArgumentNode : SyntaxNode
             Argument argument,
             CommandNode parent) : base(token, parent)
         {
-            if (token.Type != TokenType.Argument)
-            {
-                throw new ArgumentException($"Incorrect token type: {token}");
-            }
+            Debug.Assert(token.Type == TokenType.Argument, $"Incorrect token type: {token}");
```
adamsitnik, line 15: these types are internal and we control the input, so there is no need to the throw condition. `Debug.Assert` will keep us safe (the CI runs all the tests in debug) with 0 cost for the `Release` build (less IL to compile, possibility to inline the method)

src/System.CommandLine/Parsing/Parser.cs
```diff
@@ -41,9 +41,11 @@ public Parser() : this(new RootCommand())
         /// <param name="rawInput">The complete command line input prior to splitting and tokenization. This input is not typically available when the parser is called from <c>Program.Main</c>. It is primarily used when calculating completions via the <c>dotnet-suggest</c> tool.</param>
         /// <returns>A <see cref="ParseResult"/> providing details about the parse operation.</returns>
         public ParseResult Parse(
-            IReadOnlyList<string> arguments,
+            IReadOnlyList<string>? arguments,
```
adamsitnik, line 44: there is a test that verifies that the user can pass `null` here. So I made it more visible.

src/System.CommandLine/Parsing/StringExtensions.cs
```diff
@@ -71,23 +71,29 @@ internal static (string? Prefix, string Alias) SplitPrefix(this string rawAlias)
...
 
-            var argList = NormalizeRootCommand(args, configuration.RootCommand, inferRootCommand);
-
-            var tokenList = new List<Token>(argList.Count);
+            var tokenList = new List<Token>(args.Count);
 
             var knownTokens = configuration.RootCommand.ValidTokens();
 
-            for (var i = 0; i < argList.Count; i++)
+            int i = FirstArgumentIsRootCommand(args, configuration.RootCommand, inferRootCommand)
```
adamsitnik, line 88: the `NormalizeRootCommand` was allocating a new list just to ensure that the list of arguments starts with the root command alias.

src/System.CommandLine/Parsing/StringExtensions.cs
```diff
@@ -131,12 +137,17 @@ internal static (string? Prefix, string Alias) SplitPrefix(this string rawAlias)
                             out var newTokens,
                             out var error))
                     {
-                        argList.InsertRange(i + 1, newTokens ?? Array.Empty<string>());
+                        if (newTokens is not null && newTokens.Count > 0)
+                        {
+                            List<string> listWithReplacedTokens = args.ToList();
```
adamsitnik, line 142: this should be rare

src/System.CommandLine/Parsing/StringExtensions.cs
```diff
@@ -71,23 +71,29 @@ internal static (string? Prefix, string Alias) SplitPrefix(this string rawAlias)
         internal static TokenizeResult Tokenize(
             this IReadOnlyList<string> args,
             CommandLineConfiguration configuration,
-            bool inferRootCommand = true)
+            bool inferRootCommand)
         {
-            var errorList = new List<string>();
+            const int FirstArgIsNotRootCommand = -1;
+
+            List<string>? errorList = null;
```
adamsitnik, line 78: usually there are no errors, so we delay the allocation until first error is found

src/System.CommandLine/Command.cs
```diff
@@ -151,34 +155,44 @@ public override IEnumerable<CompletionItem> GetCompletions(CompletionContext con
 
             if (context.WordToComplete is { } textToMatch)
             {
-                var commands = Subcommands;
-                for (int i = 0; i < commands.Count; i++)
+                if (HasSubcommands)
                 {
-                    AddCompletionsFor(commands[i]);
+                    var commands = Subcommands;
```
Jozkee, line 160: For all the `var`s that don't comply with dotnet/runtime guidelines, are you planning on addressing them all at once in the future? 
```suggestion
                    IList<Command> commands = Subcommands;
```

src/System.CommandLine/Invocation/TypoCorrection.cs
```diff
@@ -23,9 +23,10 @@ public TypoCorrection(int maxLevenshteinDistance)
...
+            for (var i = 0; i < unmatchedTokens.Count; i++)
             {
-                var token = result.UnmatchedTokens[i];
+                var token = unmatchedTokens[i];
                 var suggestions = GetPossibleTokens(result.CommandResult.Command, token).ToList();
                 if (suggestions.Count > 0)
                 {
```
Jozkee, line 32: This is one less allocation, right?
```suggestion
                var suggestions = GetPossibleTokens(result.CommandResult.Command, token);
                if (suggestions.Any())
                {
```

src/System.CommandLine/ParseResult.cs
```diff
@@ -14,44 +14,41 @@ namespace System.CommandLine
...
-                for (var i = 0; i < tokenizeResult.Tokens.Count - 1; i++)
-                {
-                    var token = tokenizeResult.Tokens[i + 1];
-                    tokens[i] = token;
-                }
-
+                // Since TokenizeResult.Tokens is not public and not used anywhere after the parsing,
+                // we take advantage of its mutability and remove the root command token
+                // instead of creating a copy of the whole list.
+                tokens.RemoveAt(0);
```
Jozkee, line 44: Is there only one ParseResult for a TokenizeResult? i.e: mutating it here won't affect other ParseResult instances, right?

src/System.CommandLine/ParseResult.cs
```diff
@@ -99,7 +98,7 @@ public class ParseResult
         /// Gets the directives found while parsing command line input.
         /// </summary>
         /// <remarks>If <see cref="CommandLineConfiguration.EnableDirectives"/> is set to <see langword="false"/>, then this collection will be empty.</remarks>
-        public IReadOnlyDictionary<string, IReadOnlyList<string>> Directives { get; }
+        public IReadOnlyDictionary<string, IReadOnlyList<string>> Directives => _directives ??= new ();
```
Jozkee, line 101:
```suggestion
        public IReadOnlyDictionary<string, IReadOnlyList<string>> Directives => _directives ??= new();
```

src/System.CommandLine/Parsing/ParseOperation.cs
```diff
@@ -202,16 +200,14 @@ private void ParseDirectives(CommandNode rootCommandNode)
...
-                {
-                    ':'
-                }, 2);
-
-                var key = keyAndValue[0];
-                var value = keyAndValue.Length == 2
-                                ? keyAndValue[1]
-                                : null;
+                ReadOnlySpan<char> withoutBrackets = token.Value.AsSpan(1, token.Value.Length - 2);
+                int commaIndex = withoutBrackets.IndexOf(':');
```
Jozkee, line 204:
```suggestion
                int indexOfColon = withoutBrackets.IndexOf(':');
```

src/System.CommandLine/Command.cs
adamsitnik: We need to discuss it with other contributors. And the usage of `is { }` instead `is not null` which drives me crazy ;p

src/System.CommandLine/ParseResult.cs
adamsitnik: Yes, as each parsing operation creates it's own tokens and returns a single `ParseResult`:

https://github.com/dotnet/command-line-api/blob/23a3b3f20c558afb75a5d0af913512c657babec4/src/System.CommandLine/Parsing/Parser.cs#L40-L58

src/System.CommandLine/Invocation/TypoCorrection.cs
adamsitnik: Actually there was a simpler solution: https://github.com/dotnet/command-line-api/pull/2001/commits/96c8e77a6334b2d7ba93b4d71e37f26f33f9ae24

src/System.CommandLine/ParseResult.cs
adamsitnik: I think that this is a matter of personal preference. We should discuss it with others, choose one policy and enforce it with static code analysis tools.

adamsitnik: https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1000.md

PR #2016, opened 2023-01-12
More parsing improvements

adamsitnik:
Most likely the best way to review this PR is to read each commits description


src/System.CommandLine/Parsing/SymbolResult.cs
```diff
@@ -111,24 +108,34 @@ public LocalizationResources LocalizationResources
...
+
+        private SymbolResult GetRoot()
+        {
+            SymbolResult result = this;
+            while (result.Parent is not null)
+            {
+                result = result.Parent;
+            }
+
+            Debug.Assert(result is RootCommandResult);
```
jonsequitur, line 135: We've avoided using `Debug.Assert` in the past. We should discuss.

jonsequitur (approved):

PR #2024, opened 2023-01-18
SymbolResults improvements

adamsitnik:
SymbolResults improvements based on feedback from our last meeting ([notes](https://gist.github.com/adamsitnik/b7531299c37753c3ccc3bb37e832036f)):

- single, readonly LocalizationResources for entire symbol tree
- removed Symbol from SymbolResult, as each derived type defines a more specific property (OptionResult.Option etc)
- moved SymbolResult.Children to Command.Children, as it's the only type that officially has them (the fact that Option has them is an implementation detail)
- removed a lot of allocations by not storing a list of children per symbol/command and re-using the information present in symbol dictionary (we just search for instances with Parent equal to given symbol result)

I still need to implement a copy-free approach for `Symbol.Tokens` and I don't like the fact that many of the SymbolResults properties/methods are virtual and just calling `Root.MethodAbc`. I'll send another PR tomorrow or just update this one.

### DIFF
--- a/src/System.CommandLine.Tests/ArgumentTests.cs
+++ b/src/System.CommandLine.Tests/ArgumentTests.cs
@@ -131,7 +131,7 @@ namespace System.CommandLine.Tests
                 argument.Parse("x")
                         .Errors
                         .Should()
-                        .ContainSingle(e => e.SymbolResult.Symbol == argument)
+                        .ContainSingle(e => ((ArgumentResult)e.SymbolResult).Argument == argument)
                         .Which
                         .Message
                         .Should()
@@ -150,7 +150,7 @@ namespace System.CommandLine.Tests
                 argument.Parse("")
                         .Errors
                         .Should()
-                        .ContainSingle(e => e.SymbolResult.Symbol == argument)
+                        .ContainSingle(e => ((ArgumentResult)e.SymbolResult).Argument == argument)
                         .Which
                         .Message
                         .Should()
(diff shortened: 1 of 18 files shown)


src/System.CommandLine/Parsing/ParseResultVisitor.cs
```diff
@@ -179,31 +162,33 @@ private void VisitOptionNode(OptionNode optionNode)
                     optionNode.Token,
                     _innermostCommandResult);
 
-                AddToResult(optionResult);
+                _symbolResults.Add(optionNode.Option, optionResult);
+
+                if (optionNode.Children is null) // no Arguments
+                {
+                    if (optionResult.Option.Argument.HasCustomParser)
```
adamsitnik, line 169: previously this logic was performed as part of the validation (`ValidateAndConvertOptionResult`). Since here we know that given `OptionNode` was parsed with no arguments, we can create the argument result now and let validation method just validate without performing mutations.

https://github.com/dotnet/command-line-api/blob/9c48247db71bfbf7309ec48d87929a8fe425a64d/src/System.CommandLine/Parsing/ParseResultVisitor.cs#L489-L500

src/System.CommandLine.Tests/ArgumentTests.cs
```diff
@@ -333,9 +339,12 @@ public void Command_ArgumentResult_Parent_is_set_correctly_when_token_is_implici
 
                 argumentResult
                     .Parent
-                    .Symbol
                     .Should()
-                    .Be(command);
+                    .BeAssignableTo<CommandResult>()
```
jonsequitur, line 343: I think you can use `BeOfType` here, no?

jonsequitur (approved):

src/System.CommandLine.Tests/ArgumentTests.cs
adamsitnik: iirc it's `RootCommandResult` here (which is internal)

adamsitnik:
@jonsequitur thank you for the review!

PR #2029, opened 2023-01-20
UnrecognizedArgument error should mention first invalid argument, not the last one

adamsitnik:
In #1959 I've introduce a bug that SDK has hit in https://github.com/dotnet/sdk/pull/29131

Previously, the method was returning error for the first mismatch:

https://github.com/dotnet/command-line-api/blob/82273cb56c83b589e8e5b63da0ac9745ffc6e105/src/System.CommandLine/Parsing/SymbolResult.cs#L132-L135

After my changes it was the last one. This fix just helps me to unblock the SDK update, it does not solve the problem of producing multiple errors by a single validator.

### DIFF
--- a/src/System.CommandLine.Tests/ParsingValidationTests.cs
+++ b/src/System.CommandLine.Tests/ParsingValidationTests.cs
@@ -178,12 +178,12 @@ namespace System.CommandLine.Tests
             option.Arity = new ArgumentArity(1, 4);
             option.AllowMultipleArgumentsPerToken = true;
 
-            var command = new Command("--list")
+            var command = new Command("list")
             {
                 option
             };
 
-            var result = command.Parse("--list --columns c1 c2");
+            var result = command.Parse("list --columns c1 c2");
 
             // Currently there is no possibility for a single validator to produce multiple errors,
             // so only the first one is checked.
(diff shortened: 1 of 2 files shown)


src/System.CommandLine.Tests/ParsingValidationTests.cs
```diff
@@ -170,6 +170,29 @@ public void When_FromAmong_is_used_for_multiple_arguments_and_invalid_input_is_p
...
 
+        [Fact]
+        public void When_FromAmong_is_used_and_multiple_invalid_inputs_are_provided_the_error_mentions_first_invalid_argument()
+        {
+            Option<string[]> option = new(new[] { "--columns" });
+            option.AcceptOnlyFromAmong("author", "language", "tags", "type");
+            option.Arity = new ArgumentArity(1, 4);
+            option.AllowMultipleArgumentsPerToken = true;
+
+            var command = new Command("--list")
```
jonsequitur, line 181:
```suggestion
            var command = new Command("list")
```

jonsequitur (approved):

src/System.CommandLine.Tests/ParsingValidationTests.cs
```diff
@@ -170,6 +170,29 @@ public void When_FromAmong_is_used_for_multiple_arguments_and_invalid_input_is_p
...
+            {
+                option
+            };
+
+            var result = command.Parse("--list --columns c1 c2");
```
adamsitnik, line 186:
```suggestion
            var result = command.Parse("list --columns c1 c2");
```

PR #2030, opened 2023-01-23
Parent search bug fix

adamsitnik:
Every symbol can have multiple parents. When getting the completions for commands, the global options defined by the parent commands should be included as well. The problem was that we were searching only the parents of current command (flat, one level above), rather going up in the hierarchy to reach the root command

Discovered by https://github.com/dotnet/sdk/pull/29131


adamsitnik:
@KalleOlaviNiemitalo thank you for excellent feedback! I've added a failing test based on it and provided the fix. PTAL again

jonsequitur (approved):

PR #2033, opened 2023-01-24
more flexible error reporting for SymbolResult

adamsitnik:
I've removed `SymbolResult.ErrorMessage` as it did not allow for providing multiple errors for single result and introduced `SymbolResult.ReportError` (name can be changed, I don't have strong preferences about it).

`SymbolResult.ReportError` adds a new error to the list of errors (so multiple errors per result are supported) and in case of `ArgumentResult` also sets `ArgumentConversionResult` (perf, simplicity) and encapsulates the fact that `Option.Argument` is internal and for Arguments owned by Option sets the parent to `OptionResult` rather than `ArgumentResult`

I also did a lot of refactoring related to `ArgumentConversionResult`, I am going to add comments explaining why.


src/System.CommandLine/Binding/ArgumentConversionResult.cs
```diff
@@ -1,73 +1,94 @@
...
 
-                switch (identifierSymbol)
+                if (completionItems.Length > 0)
                 {
-                    case Command _:
-                        return localizationResources.ArgumentConversionCannotParseForCommand(value, alias, expectedType);
-                    case Option _:
-                        return localizationResources.ArgumentConversionCannotParseForOption(value, alias, expectedType);
+                    return argumentResult.LocalizationResources.ArgumentConversionCannotParseForCommand(
+                        value, alias, expectedType, completionItems.Select(ci => ci.Label));
```
adamsitnik, line 67: previously it was done by `ParseResultVisitor` (the error message was mutated, which did not play well with the new way of reporting erros) and was not using localization for the extra information received from completions. I moved entire logic here but this required having `ArgumentResult` and the information it provides

https://github.com/dotnet/command-line-api/blob/edd256b2fe63b57a79b39f10d0c63c81d43218ce/src/System.CommandLine/Parsing/ParseResultVisitor.cs#L522-L533

src/System.CommandLine/Parsing/ArgumentResult.cs
```diff
@@ -80,90 +80,66 @@ public void OnlyTake(int numberOfTokens)
...
-
-                if (!string.IsNullOrWhiteSpace(ErrorMessage))
-                {
-                    return new ParseError(ErrorMessage!, Parent is OptionResult option ? option : this);
-                }
-            }
-
-            return null;
+            SymbolResultTree.ReportError(new ParseError(errorMessage, Parent is OptionResult option ? option : this));
+            _conversionResult = ArgumentConversionResult.Failure(this, errorMessage, ArgumentConversionResultType.Failed);
```
adamsitnik, line 87: at this moment we know that it's a failure

src/System.CommandLine.ApiCompatibility.Tests/ApiCompatibilityApprovalTests.System_CommandLine_api_is_not_changed.approved.txt
```diff
@@ -434,6 +436,7 @@ System.CommandLine.Parsing
     public System.Object GetValue(System.CommandLine.Argument argument)
     public T GetValue<T>(Option<T> option)
     public System.Object GetValue(System.CommandLine.Option option)
+    public System.Void ReportError(System.String errorMessage)
```
jonsequitur, line 439: I'd suggest `AddError` since nothing is actually being reported at this point.

src/System.CommandLine.Tests/ParseDiagramTests.cs
```diff
@@ -59,7 +59,7 @@ public void Parse_diagram_shows_type_conversion_errors()
 
             result.Diagram()
                   .Should()
-                  .Be($"[ {RootCommand.ExecutableName} [ -f !<not-an-int> ] ]");
+                  .Be($"[ {RootCommand.ExecutableName} ![ -f <not-an-int> ] ]");
```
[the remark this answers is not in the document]
jonsequitur: Nice catch!

src/System.CommandLine/Parsing/ArgumentResult.cs
```diff
@@ -55,115 +53,152 @@ public sealed class ArgumentResult : SymbolResult
...
-            return null;
+        /// <inheritdoc/>
+        public override void AddError(string errorMessage)
+        {
+            SymbolResultTree.AddError(new ParseError(errorMessage, Parent is OptionResult option ? option : this));
+            _conversionResult = ArgumentConversionResult.Failure(this, errorMessage, ArgumentConversionResultType.Failed);
         }
 
-        private ArgumentConversionResult Convert(Argument argument)
+        private ArgumentConversionResult ValidateAndConvert(bool potentialRecursion)
```
adamsitnik, line 134: this method is now responsible for entire validation. The problem is that it can be called by not only the `ParseResultVisitor` but also custom parsers and validator which can do things like `ArgumentResult.GetDefaultValue()`. For the latter case I've added `potentialRecursion` argument to ensure we don't have endless recursion (validator->getvalue->validator)

src/System.CommandLine/Parsing/CommandResult.cs
```diff
@@ -36,32 +36,147 @@ public sealed class CommandResult : SymbolResult
         /// </summary>
         public IEnumerable<SymbolResult> Children => SymbolResultTree.GetChildren(this);
 
-        internal sealed override int MaximumArgumentCapacity
+        internal override bool UseDefaultValueFor(ArgumentResult argumentResult)
+            => argumentResult.Argument.HasDefaultValue && argumentResult.Tokens.Count == 0;
+
+        /// <param name="completeValidation">Only the inner most command goes through complete validation.</param>
+        internal void Validate(bool completeValidation)
```
adamsitnik, line 43: I've moved this logic from the visitor and re-wrote in a way that we don't iterate over every argument more than once (previously we were setting the default values and then validating).

src/System.CommandLine.Tests/ParsingValidationTests.cs
```diff
@@ -185,12 +192,17 @@ public void When_FromAmong_is_used_and_multiple_invalid_inputs_are_provided_the_
 
             var result = command.Parse("list --columns c1 c2");
 
-            // Currently there is no possibility for a single validator to produce multiple errors,
-            // so only the first one is checked.
+            result.Errors.Count.Should().Be(2);
```
adamsitnik, line 195: > Was a test added to pin the expected behavior when ReportError is called more than once?

@jonsequitur I've extended this test to cover this scenario. Is that enough?

src/System.CommandLine.ApiCompatibility.Tests/ApiCompatibilityApprovalTests.System_CommandLine_api_is_not_changed.approved.txt
adamsitnik: @jonsequitur done

src/System.CommandLine.Tests/ParsingValidationTests.cs
jonsequitur: The test makes sense but the test name no longer matches the scenario.

src/System.CommandLine/Binding/ArgumentConversionResult.cs
jonsequitur: Very nice.

src/System.CommandLine/Parsing/ArgumentResult.cs
```diff
@@ -55,115 +53,152 @@ public sealed class ArgumentResult : SymbolResult
...
+        /// <exception cref="NotSupportedException">Thrown if this method is called by Option-owned ArgumentResult.</exception>
         public void OnlyTake(int numberOfTokens)
         {
             if (numberOfTokens < 0)
             {
                 throw new ArgumentOutOfRangeException(nameof(numberOfTokens), numberOfTokens, "Value must be at least 1.");
             }
 
-            if (PassedOnTokens is { })
+            if (_passedOnHasBeenCalled)
```
jonsequitur, line 64: Does this mean `OnlyTake` has been called?

src/System.CommandLine/Parsing/ArgumentResult.cs
```diff
@@ -55,115 +53,152 @@ public sealed class ArgumentResult : SymbolResult
...
             {
                 throw new InvalidOperationException($"{nameof(OnlyTake)} can only be called once.");
             }
 
-            if (_tokens is not null)
+            if (Parent is OptionResult)
             {
-                var passedOnTokensCount = _tokens.Count - numberOfTokens;
+                throw new NotSupportedException($"TakeOnly is supported only by Command-owned ArgumentResults");
```
jonsequitur, line 71:
```suggestion
                throw new NotSupportedException($"{nameof(OnlyTake)} is supported only for a {nameof(Command)}-owned {nameof(ArgumentResult)}");
```

...mainly because at least one of these is likely to get renamed.

src/System.CommandLine/Parsing/ArgumentResult.cs
jonsequitur: Very nice. As long as we don't later allow clearing of the errors.

src/System.CommandLine/Parsing/ArgumentResult.cs
jonsequitur: The name `potentialRecursion` is a little unclear to me. Is this saying whether to allow recursion?

src/System.CommandLine/Parsing/CommandResult.cs
```diff
@@ -36,32 +36,147 @@ public sealed class CommandResult : SymbolResult
...
+            if (completeValidation)
+            {
+                if (Command.Handler is null && Command.HasSubcommands)
+                {
+                    SymbolResultTree.InsertFirstError(
+                        new ParseError(LocalizationResources.RequiredCommandWasNotProvided(), this));
+                }
+            }
+
+            if (completeValidation && Command.HasValidators)
```
jonsequitur, line 54: This looks like it can be moved into the above block since you're rechecking `completeValidation`.

src/System.CommandLine/Parsing/CommandResult.cs
```diff
@@ -36,32 +36,147 @@ public sealed class CommandResult : SymbolResult
...
+            var arguments = Command.Arguments;
+            for (var i = 0; i < arguments.Count; i++)
             {
-                ArgumentResult arg => arg.Argument.HasDefaultValue && 
-                                      arg.Tokens.Count == 0,
-                _ => false
-            };
+                Argument argument = arguments[i];
+
+                if (!completeValidation && !argument.HasDefaultValue)
```
jonsequitur, line 153: It looks like you can just early return from the whole method without entering the loop if `completeValidation == false`.

src/System.CommandLine/Parsing/ParseResultVisitor.cs
```diff
@@ -34,87 +28,57 @@ internal sealed class ParseResultVisitor
...
+            if (node is OptionNode optionNode)
+                VisitOptionNode(optionNode);
+            else if (node is OptionArgumentNode optionArgumentNode)
+                VisitOptionArgumentNode(optionArgumentNode);
+            else if (node is CommandArgumentNode commandArgumentNode)
+                VisitCommandArgumentNode(commandArgumentNode);
+            else if (node is CommandNode commandNode)
+                VisitCommandNode(commandNode);
+            else if (node is DirectiveNode directiveNode)
+                VisitDirectiveNode(directiveNode);
```
jonsequitur, line 70: Please add braces.

jonsequitur (approved):

PR #2042, opened 2023-02-02
use Validator to set Handler for VersionOption

adamsitnik:
@jonsequitur while reviewing #1969 I reminded myself about the concept of `ActiveOption` and got to conclusion that it can be implemented by using existing public APIs.

During validaiton of an OptionResult we know that it's Parent must be CommandResult. CommandResult exposes Command, which exposes Handler. So we can set it during validation.

In case of VersionOption we can validate that there are no errors and then just set the handler.

I am not sure whether Validators should have side effects like this, I just want to hear your thoughts on this.

### DIFF
--- a/src/System.CommandLine/Help/VersionOption.cs
+++ b/src/System.CommandLine/Help/VersionOption.cs
@@ -1,8 +1,9 @@
 ﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.CommandLine.Invocation;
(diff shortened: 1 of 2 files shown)


jonsequitur:
I agree with @KalleOlaviNiemitalo. 

As a principle, we should avoid validators (or anything that happens during parsing) creating side effects.

Another consideration to keep in mind is that analysis of the parser configuration, without parsing things, should be able to provide information such as which options are "active" / invokable.

adamsitnik:
> What concept is that?

An option which performs an action when it's provided by the user. Examples: `--help` (displays helps) and `--version` (displays version)

PR #2044, opened 2023-02-06
Moving CancelOnProcessTermination out of middleware, better cancellation support

### LINKED ISSUE #1783: How to propagate custom CancellationToken?
I'm using CommandLine library in embedded scenario where input and output are fully controlled by the host (and not associated with stdin and stdout). In this case, `CancelOnProcessTermination` makes no sense. Instead, I want to propagate cancellation token from the host. At the moment, it is not possible. `InvocationContext.GetCancellationToken` returning token associated with internal CTS. As a result, there is no way to cancel the operation by the host.

To address this issue, I would like to propose some changes in public API:
```csharp
public static class ParserExtensions
{
    // added token as the last parameter
    public static int Invoke(this Parser parser, string commandLine, IConsole? console = null, CancellationToken token = default);
    public static int Invoke(this Parser parser, string[] args, IConsole? console = null, CancellationToken token = default);
    public static Task<int> InvokeAsync(this Parser parser, string commandLine, IConsole? console = null, CancellationToken token = default);
    public static Task<int> InvokeAsync(this Parser parser, string[] args, IConsole? console = null, CancellationToken token = default);
}

public static class CommandExtensions
{
    // added token as the last parameter
    public static int Invoke(this Command command, string[] args, IConsole? console = null, CancellationToken token = default);
    public static int Invoke(this Command command, string commandLine, IConsole? console = null, CancellationToken token = default);
    public static Task<int> InvokeAsync(this Command command, string[] args, IConsole? console = null, CancellationToken token = default);
    public static Task<int> InvokeAsync(this Command command, string commandLine, IConsole? console = null, CancellationToken token = default);
}
```

If the changes will be approved, I'm ready to make a PR.

### LINKED ISSUE #1921: Consider adding a general capability, independent of current invocation concepts, to align shutdown handling with standard cancellation APIs. (For example, should `CancellationToken` support be available in `async Main`?)

### LINKED ISSUE #2019: Ctrl+C gets ignored
```cs
using System.CommandLine;

new MyRootCommand().Invoke(args);

sealed class MyRootCommand : RootCommand
{
    public MyRootCommand()
    {
        this.SetHandler(() =>{
            System.Console.WriteLine("Sleeping forever.");
            Thread.Sleep(int.MaxValue);
        });
    }
}
```

This app should terminate when the user presses `Ctrl+C `.

It does so when using the package from nuget.org (`2.0.0-beta4.22272.1`), but it does not with the package from https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-libraries/nuget/v3/index.json (`2.0.0-beta4.22613.1`).

It may have regressed in https://github.com/dotnet/command-line-api/pull/1502, see https://github.com/dotnet/command-line-api/pull/1502#issuecomment-985775907.

cc @Keboo @jonsequitur @adamsitnik

adamsitnik:
Asynchronous APIs have two advantages: improved scalability and cancellation support. In the case of command line applications, we mostly care about the latter.

In my opinion storing the `CancellationToken` inside `InvocationContext` is not intuitive, as C# devs are just used to passing an instance of CT to async methods. I expect that most of the S.CL users use `InvokeAsync` and never check for the cancellation, so they don’t take advantage of any async benefits.

I’ve made `CancellationToken` a mandatory argument for the async `SetHandler` overload. Now, when the users don’t pass it to any async method used inside the handler, they are going to get a compiler warning:

```cs
command.SetHandler(async (context, cancellationToken) =>
{
    await Task.Delay(Timeout.InfiniteTimeSpan);
});
```

```log
Forward the 'cancellationToken' parameter to the 'Delay' method or pass in 'CancellationToken.None' explicitly to indicate intentionally not propagating the token
```

It allows us for removing `GetCancellationToken`, `LinkToken` and `Dispose` from `InvocationContext` and gets us closer to the goal of replacing all usages of `InvocationContext` with just `ParseResult`.

Another goal was moving `CancelOnProcessTermination` out of middleware. This was not easy, but I've managed to get it working and make it easier to use.

What I don't like is adding **another** argument to `GetHandler` extension methods. Since we are going to remove all of them soon, I hope it's not an issue.

The app I've used for testing:

<details>

```cs
// Copyright (c) .NET Foundation and contributors. All rights reserved.

using System.Threading.Tasks;
using System.Threading;

namespace System.CommandLine.Benchmarks
{
    class Program
    {
        {
            Argument<bool> argument = new()
            {
                Arity = ArgumentArity.ExactlyOne
            };

            RootCommand command = new ()
            {
                argument
            };

            command.SetHandler(async (context, cancellationToken) =>
            {
                try
                {
                    bool value = context.GetValue(argument);
                    string tokenDesc = value ? "the right token" : "the wrong token";
                    context.Console.WriteLine($"Starting infinite delay using {tokenDesc}");

                    CancellationToken token = value 
                        ? cancellationToken // the cancellation will work fine
                        : CancellationToken.None; // the timeout passed to CancelOnProcessTermination will be used

                    await Task.Delay(Timeout.InfiniteTimeSpan, token);
                    context.ExitCode = 1;
                }
                catch (OperationCanceledException)
                {
                    context.Console.WriteLine("Cancellation worked as expected");
                }
            });

            var built = new CommandLineBuilder(command)
               .CancelOnProcessTermination(TimeSpan.FromSeconds(1))
               .UseExceptionHandler()
               .Build();

            int exitCode = await built.InvokeAsync(args);
            Console.WriteLine($"Exit code={exitCode}");
            return exitCode;
        }
    }
}
```

</details>

fixes #1783
fixes #1921
fixes #2019


adamsitnik:
@jonsequitur the Linux CI leg might be red, but please take a look at the PR description and let me know what do you think about such changes. I want to make sure I am going in the right direction before I spend more time working on it.

jonsequitur:
There's a third reason why people use `async` methods and based on feedback from people using this API, this one is much more common than the two mentioned: the need to call `async` APIs. Given this, I think requiring the parameter might be adding some ceremony where we're hoping to reduce it.

jonsequitur:
Any redesign of cancelation probably deserves its own discussion as there are a few different use cases to consider. 

Related:  #1783, #1921

src/System.CommandLine.Hosting/InvocationLifetime.cs
```diff
@@ -65,7 +58,7 @@ public Task WaitForStartAsync(CancellationToken cancellationToken)
                 }, this);
             }
 
-            invokeCancelReg = invokeCancelToken.Register(state =>
+            invokeCancelReg = cancellationToken.Register(state =>
```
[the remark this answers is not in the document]
adamsitnik: Very good point, I've debugged it to find out how it works and extended middleware with `CancellationToken`.

It's now being passed here:

https://github.com/dotnet/command-line-api/blob/5905a9bc428b5d25a1ec6af831afda53add0fd86/src/System.CommandLine.Hosting/HostingExtensions.cs#L21

https://github.com/dotnet/command-line-api/blob/5905a9bc428b5d25a1ec6af831afda53add0fd86/src/System.CommandLine.Hosting/HostingExtensions.cs#L47

src/System.CommandLine.Tests/Invocation/CancelOnProcessTerminationTests.cs
```diff
@@ -15,222 +15,76 @@ namespace System.CommandLine.Tests.Invocation
...
-            // Request termination
-            kill(process.Id, signo).Should().Be(0);
+            SIGINT = 2, // Console.CancelKeyPress
+            SIGTERM = 15 // AppDomain.CurrentDomain.ProcessExit
+        }
 
-            // Verify the process terminates timely
-            bool processExited = process.WaitForExit(10000);
-            if (!processExited)
+        [Fact]
```
Keboo, line 29: Any reason for using the runtime platform check rather than making a `LinuxOnlyFact`?

src/System.CommandLine/Builder/CommandLineBuilderExtensions.cs
```diff
@@ -26,7 +26,7 @@ public static class CommandLineBuilderExtensions
         /// <param name="builder">A command line builder.</param>
         /// <param name="timeout">
         /// Optional timeout for the command to process the exit cancellation.
-        /// If not passed, or passed null or non-positive timeout (including <see cref="Timeout.InfiniteTimeSpan"/>), no timeout is enforced.
+        /// If not passed , a default timeout of 2 seconds is enforced.
```
Keboo, line 29:
```suggestion
        /// If not passed, a default timeout of 2 seconds is enforced.
```

src/System.CommandLine/Invocation/InvocationContext.cs
```diff
@@ -1,78 +1,41 @@
...
-            get
-            {
-                if (_console is null)
-                {
-                    _console = new SystemConsole();
-                }
-
-                return _console;
-            }
+            get =>_console ??= new SystemConsole();
```
Keboo, line 38: nit:
```suggestion
            get => _console ??= new SystemConsole();
```

Keboo (approved):

src/System.CommandLine.Tests/Invocation/CancelOnProcessTerminationTests.cs
adamsitnik: My initial intention was to enable it on macOS too, but then it turned out that `RemoteExecutor` does not support this scenario, so instead of adding new attribute (This repo defines only `LinuxOnlyTheory` attribute. ) I've added the ugly if statement.

adamsitnik:
@Keboo thanks a lot for the review!

adamsitnik:
> Anyway, this is a risk for developers upgrading from prior versions of System.CommandLine.

Thanks for the feedback! I currently don't see a way to avoid that, but in the very near future we are most likely going to replace `InvocationContext` accepting handlers with `ParseResult` based ones. I hope that this will force the users to switch to the right overload.

And in general the `SetHandler` is still far for perfect, we are opened to suggestions in improving it.

PR #2060, opened 2023-02-17
Rename Option.IsGlobal to AppliesToSelfAndChildren, make it public

### LINKED ISSUE #2055: Rename IsGlobal to AppliesToSelfAndChildren and expose it
- [ ] rename `Option.IsGlobal` to `Option.AppliesToSelfAndChildren`
- [ ] make it public
- [ ] remove `Command.AddGlobalOption`

This might be not the final name, we might rename it once the API gets approved by .NET Team API Review Board (.NET 8 Preview 3-4 timeframe)

adamsitnik:
fixes #2055

Note from #2055: "This might be not the final name, we might rename it once the API gets approved by .NET Team API Review Board (.NET 8 Preview 3-4 timeframe)"

### DIFF
--- a/src/System.CommandLine.Tests/ParsingValidationTests.cs
+++ b/src/System.CommandLine.Tests/ParsingValidationTests.cs
@@ -458,7 +458,7 @@ namespace System.CommandLine.Tests
         [InlineData("subcommand --file \"Foo\"")]
         public void Validators_on_global_options_are_executed_when_invoking_a_subcommand(string commandLine)
         {
-            var option = new Option<FileInfo>("--file");
+            var option = new Option<FileInfo>("--file") { AppliesToSelfAndChildren = true };
             option.Validators.Add(r =>
             {
                 r.AddError("Invoked validator");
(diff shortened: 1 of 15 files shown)


src/System.CommandLine.Tests/ParsingValidationTests.cs
```diff
@@ -495,7 +495,11 @@ public async Task A_custom_validator_added_to_a_global_option_is_checked(string
         {
             var handlerWasCalled = false;
 
-            var globalOption = new Option<int>("--value");
+            var globalOption = new Option<int>("--value")
```
Keboo, line 498: Do we want to rename this local since we are removing the global concept?

src/System.CommandLine/Option.cs
```diff
@@ -75,7 +75,7 @@ public ArgumentArity Arity
         /// Global options are applied to the command and recursively to subcommands.
         /// They do not apply to parent commands.
         /// </summary>
```
Keboo, line 77:
```suggestion
        /// When set to true, this option will be applied to the command and recursively to subcommands.
        /// It will not apply to parent commands.
        /// </summary>
```

Keboo (approved):

src/System.CommandLine.Tests/ParsingValidationTests.cs
adamsitnik: I was thinking about it, but I was unable to come up with a better name. `appliesToSelfAndChildrenOption`? `inheritableOption`?

Issue #2057, opened 2023-02-17
Make CommandLineConfiguration mutable


adamsitnik:
- [ ] #2054 needs to be implemented first, as it's not clear whether we will be able to remove the `CommandLineBuilder` type. `Builder` types are the only types in BCL that are allowed to use fluent APIs (returning reference to `this` to allow for calls chaining) and I am currently not sure if we are going to be able to repsent all configuration options as properties (many extension methods set two things at a time).
- [ ] #2056 needs to be implemented first as it will have great impact on how `Directives` will be configured. We don't want to have `EnableDirectiveXYZ` property for every Directive
- [ ] make all the internal fields and properties public, with nice names
- [ ] make all the public properties mutable

PR #2062, opened 2023-02-20
introduce Command.Parse(CommandLineConfiguration)

### LINKED ISSUE #2054: Add Symbol.Parse, make Parser static
- [ ] expose `ParseResult Symbol.Parse(string[] args, CommandLineConfiguration = default)`, so users can easily discover parsing
- [ ] make `Parser` a static type with a single `ParseResult Parse(string[] args, CommandLineConfiguration = default)` method
- [ ] make `CommandLineBuilder.Build` return `CommandLineConfiguration`, move all extension methods from `CommandLineBuilderExtensions` to `CommandLineBuilder` itself
- [ ] Move the actual parsing logic from `StringExtensions` to `Parser`:

https://github.com/dotnet/command-line-api/blob/main/src/System.CommandLine/Parsing/StringExtensions.cs#L72

adamsitnik:
Initially, I just wanted  to extend `Symbol` with `Parse(string[], CommandLineConfiguration)`.

I've realized that the existing `Argument.Parse` and `Option.Parse` methods have side effects, namely creating a `RootCommand` for given option/argument which sets the Symbol parent:

https://github.com/dotnet/command-line-api/blob/704e6408c2bd050581a7982453c9c9d93e89eea5/src/System.CommandLine/SymbolExtensions.cs#L86-L88

https://github.com/dotnet/command-line-api/blob/64622889f74f14e8c515fb72be41846671c569ac/src/System.CommandLine/ChildList.cs#L37-L41

Since these methods were used only for testing and can be very easily replaced with explicit one liner I've decided to remove them:

```diff
- argument.Parse("x")
+ new RootCommand { argument }.Parse("x")
```

After doing that I've just moved `CommandExtensions.Parse` to `Command` to make it easier to discover (and follow BCL guidelines).

When I wanted to make `Parser` a static class, I've realized that we had two different default configs:

https://github.com/dotnet/command-line-api/blob/704e6408c2bd050581a7982453c9c9d93e89eea5/src/System.CommandLine/SymbolExtensions.cs#L41

https://github.com/dotnet/command-line-api/blob/704e6408c2bd050581a7982453c9c9d93e89eea5/src/System.CommandLine/SymbolExtensions.cs#L54

I've decided to use the same default config everywhere (the one that uses `UseDefaults`) and `CompletionTests` tests started failing, as they were not expecting the suggestions to contain `--help` and `--version` (global options added by default in the default config). I've just fixed the tests by passing the "simple" config in explicit way.

Another change that I've made was making `Command` a mandatory argument for `CommandLineBuilder` ctor:

https://github.com/dotnet/command-line-api/blob/704e6408c2bd050581a7982453c9c9d93e89eea5/src/System.CommandLine/Builder/CommandLineBuilder.cs#L26-L28

I did that because it was making it easy to fall into a trap of not passing the existing command to the ctor and ending up with "disconnected" root command being used by the config:

```cs
RootCommand command = new()
{
    new Option<bool>(new[] { "--bool", "-b" }, "Bool option");
};

var builder = new CommandLineBuilder(); // not passing the command we already have, BUG
return builder.UseDefaults().Build().Invoke(args);
```

Since `Command.Parse` accepts `CommandLineConfiguration` and the class is not mutable yet, I wanted to make it easier to discover the builder type, so I've added:

```cs
{
     public static CommandLineBuilder CreateBuilder(Command rootCommand);
}
```

I've also moved a bunch of extension methods to the types they were extending (BCL guideline).

I hope that all these changes are going to improve the discoverability of:

- configuration and its customization
- the separation of parsing and execution

fixes #2054
makes a progress toward #2057


src/System.CommandLine.Benchmarks/CommandLine/Perf_Parser_NestedCommands.cs
```diff
@@ -62,19 +62,10 @@ public void SetupRootCommand()
...
 
             _rootCommand = rootCommand;
+            _configuration = CommandLineConfiguration.CreateBuilder(rootCommand).UseDefaults().Build(); ;
         }
 
-        [GlobalSetup(Target = nameof(Parser_Parse))]
-        public void SetupParser()
-        {
-            SetupRootCommand();
-            _testParser = new Parser(_rootCommand);
```
adamsitnik, line 71: Parser is now static, we can't benchmark it's creation ;)

src/System.CommandLine/Builder/CommandLineBuilderExtensions.cs
```diff
@@ -9,59 +9,48 @@
...
-using static System.Environment;
 using Process = System.CommandLine.Invocation.Process;
 
 namespace System.CommandLine
 {
-    /// <summary>
-    /// Provides extension methods for <see cref="CommandLineBuilder"/>.
-    /// </summary>
-    public static class CommandLineBuilderExtensions
+    public partial class CommandLineBuilder
```
adamsitnik, line 16: to keep the diff smaller, I've decided to make the class partial rather than moving all the code to `CommandLineBuilder`

src/System.CommandLine/Parsing/Parser.cs
```diff
@@ -2,56 +2,64 @@
...
         /// <summary>
-        /// Gets the configuration on which the parser's grammar and behaviors are based.
+        /// Parses a list of arguments.
         /// </summary>
-        public CommandLineConfiguration Configuration { get; }
+        /// <param name="command">The command to use to parse the command line input.</param>
+        /// <param name="arguments">The string array typically passed to a program's <c>Main</c> method.</param>
+        /// <param name="configuration">The configuration on which the parser's grammar and behaviors are based.</param>
+        /// <returns>A <see cref="ParseResult"/> providing details about the parse operation.</returns>
+        public static ParseResult Parse(Command command, IReadOnlyList<string>? arguments, CommandLineConfiguration? configuration = null)
```
adamsitnik, line 21: TBH I don't like the fact that we accept `null` arguments. I would be happy to remove it if others think similar of it.

src/System.CommandLine.Tests/ParseResultTests.cs
```diff
@@ -25,7 +25,7 @@ public void An_option_with_a_default_value_and_no_explicitly_provided_argument_h
         }
 
         [Fact]
-        public void HasOption_can_be_used_to_check_the_presence_of_an_option()
+        public void FindResultForHasOption_can_be_used_to_check_the_presence_of_an_option()
```
Keboo, line 28:
```suggestion
        public void FindResult_can_be_used_to_check_the_presence_of_an_option()
```

src/System.CommandLine/Builder/CommandLineBuilderExtensions.cs
```diff
@@ -231,251 +211,230 @@ void Default(Exception exception, InvocationContext context)
...
-            builder.Command.Options.Add(builder.VersionOption);
+            return this;
+        }
+
+        /// <summary>
+        /// Creating a config from Command might cause side effects for Command.
+        /// The config type may add Options to the Command.
+        /// Since single command can be parsed multiple times with different configs,
+        /// we need to handle it properly.
+        /// Ideally Config should not mutate Command at all.
```
Keboo, line 421:
```suggestion
        /// Ideally config should not mutate Command at all.
```

src/System.CommandLine/Parsing/Parser.cs
Keboo: No objection from me.

src/System.CommandLine.Tests/ArgumentTests.cs
```diff
@@ -128,7 +128,7 @@ public void Validation_failure_message_can_be_specified_when_parsing_tokens()
                     return null;
                 });
 
-                argument.Parse("x")
+                new RootCommand { argument }.Parse("x")
```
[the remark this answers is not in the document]
adamsitnik: `Command` requires to provide a name in the `ctor`, so the keep it simpler I decided to use `RootCommand` here. In theory I could use `new Command(string.Empty)` here?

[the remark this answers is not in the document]
Keboo: This was a question not a suggested change. I know we have the other item to address `RootCommand` so this can probably wait until then.

Keboo (approved):

src/System.CommandLine/Parsing/Parser.cs
jonsequitur: The term "arguments" here is potentially confusing because of the `Argument` class. The common name for the `string[]` parameter to `Main` is `args` which is possibly clearer about its intent.

jonsequitur: Disallowing `null` here makes sense.

src/System.CommandLine/Builder/CommandLineBuilder.cs
```diff
@@ -12,10 +12,10 @@ namespace System.CommandLine
     /// <summary>
     /// Enables composition of command line configurations.
     /// </summary>
-    public class CommandLineBuilder 
+    public partial class CommandLineBuilder
```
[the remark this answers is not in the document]
jonsequitur: Our hope is to remove the builder class in any case, yes?

src/System.CommandLine.Tests/ArgumentTests.cs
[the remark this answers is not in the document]
jonsequitur: `new Command(string.Empty)` results (correctly) in an `ArgumentException`.

src/System.CommandLine/Command.cs
```diff
@@ -135,11 +135,11 @@ public void Add(Symbol symbol)
         /// <summary>
         /// Parses an array strings using the command.
         /// </summary>
-        /// <param name="args">The string arguments to parse.</param>
+        /// <param name="arguments">The string arguments to parse.</param>
         /// <param name="configuration">The configuration on which the parser's grammar and behaviors are based.</param>
         /// <returns>A parse result describing the outcome of the parse operation.</returns>
-        public ParseResult Parse(string[]? args, CommandLineConfiguration? configuration = null)
-            => Parser.Parse(this, args, configuration);
+        public ParseResult Parse(IReadOnlyList<string> arguments, CommandLineConfiguration? configuration = null)
```
jonsequitur, line 141: I would suggest the name `args` here as it aligns with the naming convention for `Main(string[] args)`.

adamsitnik: > I would suggest the name args here as it aligns with the naming convention for Main(string[] args).

done

jonsequitur (approved):

PR #2063, opened 2023-02-21
Introduce Directive Symbol type

### LINKED ISSUE #2056: Introduce Directive symbol type
From #2046:

Currently the only way for the users to implement custom directives is enabling directives parsing in the config, using this property to get parsed output:

https://github.com/dotnet/command-line-api/blob/76437b04511d88543df5cde2c7910e8d40e30888/src/System.CommandLine/ParseResult.cs#L86

and provide a custom, expensive Middleware.

We need to introduce a dedicated `Directive` type that derives from `Symbol`.

Initial requirements:

- [ ] introducing new `Directive` symbol type, with a possibility to define an `Action<ParseResult, string>` which should be executed when directive is parsed (the list of arguments for the action is an open question, when implementing the feature it should became more obvious)
- [ ] introducing `Command.Directives` property to allow users for defining directives
- [ ] turning existing code into new directives https://github.com/dotnet/command-line-api/blob/76437b04511d88543df5cde2c7910e8d40e30888/src/System.CommandLine/Parsing/ParseOperation.cs#L350-L378
- [ ] extending tokenization and parsing with the concept of Directive

adamsitnik:
fixes #2056


src/System.CommandLine.Tests/DirectiveTests.cs
```diff
@@ -100,63 +95,67 @@ public void Directives_must_have_a_non_empty_key(string directive)
...
+        //    var result = command.Parse("[one] [two] [one:a] [one:b]");
 
-            result.Directives.Should().HaveCount(2);
-        }
+        //    result.Directives.Should().HaveCount(2);
+        //}
 
         [Fact]
-        public void Directives_can_be_disabled()
+        public void When_directives_are_not_enabled_they_are_treated_as_regular_tokens()
```
jonsequitur, line 134: I would expect them to be present as tokens now in any case, though maybe only on the root command result.

src/System.CommandLine/Directive.cs
```diff
@@ -0,0 +1,75 @@
...
+            {
+                Handler = new AnonymousCommandHandler(asyncHandler);
+            }
+        }
+
+        internal ICommandHandler? Handler { get; }
+
+        /// <summary>
+        /// Method executed when given Directive is being parsed.
+        /// Useful for Directives that want to perform an action without setting the Handler for ParseResult.
```
jonsequitur, line 63: This makes me wonder how this would interact with a handler. 

* Is it always called? 
* Is it called during parsing (as the name implies)? We should avoid encouraging side effects during parsing.

Also (maybe related), how do we want to surface the difference between directives are "handlers" and completely take over the behavior of the app (e.g. `[parse]`) versus directives that have some orthogonal behavior (e.g. `[env]`).

adamsitnik: I've removed the `OnParsed`, but it required an ugly hack. The directive handler is now calling the parsed command handler in a direct way:

```cs
context.ParseResult.CommandResult.Command.Handler?.Invoke(context);
```

It's just a workaround, we need a better way to chain the handlers and also a possibility to not do that.

src/System.CommandLine/EnvironmentVariablesDirective.cs
```diff
@@ -0,0 +1,43 @@
...
+                    {
+                        string value = parsedValue.AsSpan(indexOfSeparator + 1).Trim().ToString();
+
+                        Environment.SetEnvironmentVariable(variable.ToString(), value);
+                    }
+                }
+            }
+
+            // we need a cleaner, more flexible and intuitive way of continuing the execution
+            context.ParseResult.CommandResult.Command.Handler?.Invoke(context);
```
adamsitnik, line 40: @jonsequitur I've reverted the ability to handle continuations and hardcoded it for now as we have agreed

https://github.com/dotnet/command-line-api/pull/2063/commits/fc987d4802440d9b407c8bc610919f1163a15ff2

jonsequitur (approved):

Issue #2048, opened 2023-02-16
Changes to invocation to remove middleware pipeline


KathleenDollard:
This is my representation of a bunch of conversations that have happened in imagining a replacement to the middleware pipeline. _If you customize the middleware pipeline, please read this and imagine your use case with this approach - or ask us about it._ Thanks to @jonsequitur, @adamsitnik, @Keboo, @patriksvensson, @HowardvanRooijen and @mhutch for the conversations that got us this far and to you for feedback, if you customize middleware. If you do not customize middleware, everything will work as before.

# Execution changes

We are planning significant changes to the internals of how System.CommandLine determines what to execute. 

* This will not affect you unless you customize the middleware pipeline. Most System.CommandLine users will not need to understand the new pipeline or that we have changed it.
* If you only customize the pipeline to add or remove existing elements like version, this will be a small change.
* *We need to hear from you if you do customizations to the pipeline that you do not think are supported in the new model!*

## Why the change

The current System.CommandLine internal execution uses a middleware pipeline which you can alter. There is a default so many programmers have been able to ignore the internal design. That design is a series of steps that can do work and then indicate whether they terminate that runs after parsing on invocation.  FOr example, the `UseParseErrorReporting` step displays validation issues and if there are issues does not call the next step in the pipeline - thus terminating.

A few problems with this approach:

* The code could be intimidating and it was not always obvious how to accomplish things.
* Key steps in the pipeline - like reporting validation issues and help - were tied to invocation and could not easily be used in scenarios that did not use System.CommandLine to invoke actions.
* It had a complicated order dependency. Issues were created because CLI authors incorrectly put their step before or after ParseErrorReporting, for example.
* It put items on the callstack which surprised users and could make stepping awkward during debugging.
* Version and Help options were internal, so it was difficult to customize without copying code.
* There are concerns around performance of this approach.

While these things have proved problematic, the middleware approach has been helpful in us gaining insight into what most people want to do.

*One purpose of this issue is to find scenarios we do not yet understand.*

This experience, and requests as part of generation and from other sources, has led to a deep redesign. For example, we have confidence in the set of features that are part of the common set:

| Name                            | In default set? | Sometimes terminates? | Modifies             |
|---------------------------------|-----------------|-----------------------|----------------------|
| UseVersionOption                | Yes             | Yes                   |                      |
| UseHelp                         | Yes             | Yes                   |                      |
| UseParseDirective               | Yes             | Yes                   |                      |
[...]

PR #2065, opened 2023-02-23
CLI actions prototype

jonsequitur:
These changes are a prototype for #2071.

As a next step, custom actions types should be able to be either generated or bound to command line values using source generators. More investigation is needed there.

Feedback and suggestions are welcome.

### DIFF
--- a/src/System.CommandLine.NamingConventionBinder.Tests/System.CommandLine.NamingConventionBinder.Tests.csproj
+++ b/src/System.CommandLine.NamingConventionBinder.Tests/System.CommandLine.NamingConventionBinder.Tests.csproj
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <TargetFrameworks>net7.0</TargetFrameworks>
     <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">$(TargetFrameworks);net462</TargetFrameworks>
-    <LangVersion>10</LangVersion>
+    <LangVersion>latest</LangVersion>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
     <DefineConstants>DEBUG;TRACE</DefineConstants>
--- a/src/System.CommandLine.NamingConventionBinder/System.CommandLine.NamingConventionBinder.csproj
+++ b/src/System.CommandLine.NamingConventionBinder/System.CommandLine.NamingConventionBinder.csproj
@@ -4,7 +4,7 @@
     <IsPackable>true</IsPackable>
     <PackageId>System.CommandLine.NamingConventionBinder</PackageId>
     <TargetFrameworks>net7.0;netstandard2.0</TargetFrameworks>
-    <LangVersion>10</LangVersion>
+    <LangVersion>latest</LangVersion>
     <Nullable>enable</Nullable>
     <Description>This package provides command handler support for System.CommandLine performs parameter and model binding by matching option and argument names to parameter and property names.</Description>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
--- a/src/System.CommandLine.Tests/ArgumentTests.cs
+++ b/src/System.CommandLine.Tests/ArgumentTests.cs
@@ -9,7 +9,6 @@ using FluentAssertions;
 using System.Linq;
 using System.Threading.Tasks;
 using Xunit;
-using System.CommandLine.Completions;
 
 namespace System.CommandLine.Tests
 {
--- a/src/System.CommandLine.Tests/Invocation/CliActionTests.cs
+++ b/src/System.CommandLine.Tests/Invocation/CliActionTests.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.CommandLine.Help;
 using System.CommandLine.Invocation;
 using System.CommandLine.IO;
 using System.Threading.Tasks;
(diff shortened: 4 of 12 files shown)


adamsitnik (commented):
## Single

Exposing an `Action` or a [Symbol](https://github.com/dotnet/command-line-api/issues/2048#issuecomment-1437976510) from `ParseResult`

```cs
public class ParseResult
{
    public CliAction Action { get; }
    public Symbol Symbol { get; }
}
```

has one important limitation: it assumes that only one `Symbol` provides an action that needs to be executed.

While implementing #2063 I've realised that it's not enough for directives. Example: `[env:key=value]` wants to just set the env var(s) and run the parsed command after that.
But some directives don't want to continue the execution. Example: `[parse]` just wants to print a diagram and quit.
Moreover, the directives can be combined. We may want to set env vars, current culture and even implement something like `[stopwatch]` and then just run the command.

After thinking about it for a while I believe that we already have all that is needed in order to remove middleware.

For the example provided by @KathleenDollard in #2048 we can just use `ParseResult.FindResultFor(Symbol)` which returns `null` if given symbol was not parsed. Since we care about the active symbols only, it's enough to identify every scenario. But we would need to expose `HelpOption` and `VersionOption` types, which IMO is perfectly fine.

<details>

```cs
public static int Run(Command root, string[] args, CommandLineConfiguration configuration = default)
{
    ParseResult parseResult = root.Parse(args, configuration);

    if (parseResult.Errors.Count != 0)
    {
        return DisplayValidationErrors(configuration, parseResult);
    }
    else if (parseResult.FindResultFor(new HelpOption()) is not null)
    {
        DisplayHelp(configuration, parseResult);
    }
    else if (parseResult.FindResultFor(new VersionOption()) is not null)
    {
        DisplayVersion();
    }
    else if (parseResult.FindResultFor(new ParseDirective()) is not null)
    {
        DiagramInput(parseResult);
    }
    else if (parseResult.FindResultFor(new SuggestDirective()) is not null)
    {
        SuggestResult(parseResult); // this is for tab completion
    }
    else
    {
        InvokeCommand(configuration, parseResult, invocationResult),
    }
}
```

</details>

It's not 100% intuitive yet, as it works because these types override `Equals` and `GetHashCode` and report equality for the same type, not for the same instance:

https://github.com/dotnet/command-line-api/blob/1e15ab1629ad13dee96b7a0fdb6c554f3f0528db/src/System.CommandLine/Help/HelpOption.cs#L30-L32

So we could just introduce:

```cs
public ParseResult
{
    public SymbolResult? FindResultFor<T>() where T : Symbol
}
```

But the implementation would not be 100% perf optimal (we would need to go over all parsed symbols and cast until we hit find the first instance).

So **we could introduce a name-based overload which would not require to make `HelpOption` and `VersionOption` public nor store references to symbols:**

```cs
public ParseResult
{
    public SymbolResult? FindResultFor(string name)
}
```

And have sth like this:

<details>

```cs
public static int Run(Command root, string[] args, CommandLineConfiguration configuration = default)
{
    ParseResult parseResult = root.Parse(args, configuration);

    if (parseResult.Errors.Count != 0)
    {
        return DisplayValidationErrors(configuration, parseResult);
    }
    else if (parseResult.FindResultFor("--help") is not null)
    {
        DisplayHelp(configuration, parseResult);
    }
    else if (parseResult.FindResultFor("--version") is not null)
    {
        DisplayVersion();
    }
    else if (parseResult.FindResultFor("[parse]") is not null)
    {
        DiagramInput(parseResult);
    }
    else if (parseResult.FindResultFor("[suggest]" is not null)
    {
        SuggestResult(parseResult); // this is for tab completion
    }
    else
    {
        InvokeCommand(configuration, parseResult, invocationResult),
    }
}
```

</details>

Personally I would prefer the try pattern:

<details>

```cs
public static int Run(Command root, string[] args, CommandLineConfiguration configuration = default)
{
    ParseResult parseResult = root.Parse(args, configuration);

    if (parseResult.Errors.Count != 0)
    {
        return DisplayValidationErrors(configuration, parseResult);
    }
    else if (parseResult.TryFindResultFor("--help", out SymbolResult? result))
    {
        DisplayHelp(configuration, parseResult);
    }
    else if (parseResult.TryFindResultFor("--version", out result))
    {
        DisplayVersion();
    }
    else if (parseResult.TryFindResultFor("[parse]", out result))
    {
        DiagramInput(parseResult);
    }
    else if (parseResult.TryFindResultFor("[suggest]", out result))
    {
        SuggestResult(parseResult); // this is for tab completion
    }
    else
    {
        InvokeCommand(configuration, parseResult, invocationResult),
    }
}
```

</details>

Giant if-else block does not look as pretty as switch, but I don't expect real world scenarios to be willing to customize every possible thing.
I've taken a look at the only middleware SDK is providing: https://github.com/dotnet/sdk/blob/b8f58bc235360597781de64f0c53190c80a68599/src/Cli/dotnet/Parser.cs#L163-L200

What it wants to achieve is custom error reporting for one command. This would be:

```cs
public static int Run(Command root, string[] args, CommandLineConfiguration configuration = default)
{
    ParseResult parseResult = root.Parse(args, configuration);

    if (parseResult.Errors.Count != 0 && parseResult.CommandResult.Command.Name == "specialName")
    {
        return DisplayCustomError(configuration, parseResult);
    }
    else
    {
        parseResult.Invoke();
    }
}
```

## Chaining

Having multiple actions that need to be invoked requires implementing chaining mechanism.

In my opinion, the fewer public types we expose (`CliAction`, `ICommandHandler`), the better as we can change the implementation without affecting the public contract.

What I am trying to say is that in my opinion `Command`, `Directive` and maybe `Option` should be exposing only one way to set sychronous and asychronous delegates and not exposing any property that returns something else.

```cs
public class Command 
{
    public Func<ParseResult, int> SynchronousHandler { get; set; }
    public Func<ParseResult, CancellationToken, Task<int>> AsynchronousHandler { get; set; }
}
```

And the actual chaining should be done by `Invoke` and `InvokeAsync` and being an implementation detail.

Issue #2053, opened 2023-02-17
Simplify custom parsers and default value factories


adamsitnik:
From https://github.com/dotnet/command-line-api/issues/2046 :

`Argument<T>` provides multiple ways of setting default value: as `T`, `Func<T>` and `Func<ArgumentResult, T>`. 
What I don't like is the fact that the last ctor requires the user to define whether given `Func<ArgumentResult, T>` is a default value provider or a custom parser. It's confusing. We need to find a way of providing `Func<ArgumentResult, T>` without specifying whether it's a default value provider or just a custom parser.

It applies to both `Argument<T>` and `Option<T>`

```cs
public Argument(Func<T> defaultValueFactory)
    
public Argument(
    string name,
    T defaultValue,
    string? description = null)

public Argument(
    string? name,
    Func<ArgumentResult, T> parse,
    bool isDefault = false,
    string? description = null)
```

adamsitnik:
My current best idea is following:

1. Make both `Argument<T>` and `Option<T>` have only a single ctor that always requires the name to be provided in explicit way (#2051, #2052).

```cs
public Argument(string name)
public Option(string name, params string[] aliases)
```

2. Option should accept a `params string[]` so users don't need to allocate an array in explicit way every time they create an `Option`.
3. Make other, optional information available to set and get via properties (Description etc).
4. When it comes to custom parser and default value factories, expose them also as properties:

```cs
public Func<ArgumentResult, T>? DefaultValueFactory { get; set; }
public Func<ArgumentResult, T>? ValueParser { get; set; }
```

By doing that we can:

- remove `HasDefaultValue` and `HasCustomParser` properties
- we don't need `isDefault`, as users can just assing both properties to the same method

PTAL at the example below and let me know how do you feel about it.

<details>

```cs
using System.Collections;
using System.Runtime.InteropServices;

namespace SymbolCtors
{
    internal class Program
    {
        static void Main(string[] args)
        {
            // Sample: "dotnet build"
            Command dotnetBuild = new Command("build")
            {
                new Argument<FileInfo?>("file")
                {
                    Description = "The project or solution file to operate on. If a file is not specified, the command will search the current directory for one.",
                    DefaultValueFactory = (argumentResult) =>
                    {
                        FileInfo? first = new DirectoryInfo(Directory.GetCurrentDirectory()).EnumerateFiles("*.sln|*.proj").FirstOrDefault();

                        if (first is null)
                        {
[...]

Issue #2052, opened 2023-02-17
Introduce ParseResult.GetValue<T>(string name)


adamsitnik:
- [ ] #2051 needs to be implemented first. The method should work only for the names (no short aliases). So if user defines `Option<bool>(name: "help", aliases: new [] { "--help", "-h", "-?" })` the lookup should match only `help`.
- [ ] The users might define a parse tree where there are multiple arguments/options with the same name. But conflicts are not allowed on the Command level. The method should handle that properly.
- [ ] The method needs to handle default values properly (value not specified in args, default value factory used to create it).
- [ ] The lookup needs to be fast, so it should be backed by a `Dictionary`. No need to use a concurrent version of dictionary, as the creation will be single-threaded.

One way of implementing it (just an idea) is creating a new dedicated class (not a Value Tuple because it's a `struct` and JIT compiles a dedicated version of generic collections for every value type and we don't want that as it's a performance hit for startup scenarios).
The class should store `string name` and `CommandResult parent` as fields, implement IEqutable and override `GetHashCode` and be internal and sealed. Instances of this class could be keys of the mentioned dictionary and they should be added to the dictionary at the same time when `ArgumentResult` instances are being added to `SymbolResultTree` (the values of the dictionary should be instances of `ArgumentResult`).

https://github.com/dotnet/command-line-api/blob/76437b04511d88543df5cde2c7910e8d40e30888/src/System.CommandLine/Parsing/ParseOperation.cs#L158

Finding all usages of `SymbolResultTree.*Add*` might be non-trivial, as the type derives from `Dictionary`. This can be changed, it's internal type and it was just a minor perf optimization.

When `ParseResult.GetValue<T>(string name)` is used, it should use the parsed CommandResult to find the value of given name in the context of parsed Command:

https://github.com/dotnet/command-line-api/blob/76437b04511d88543df5cde2c7910e8d40e30888/src/System.CommandLine/ParseResult.cs#L65

Issue #1976, opened 2022-11-23
Allow hiding tab-completion values from help


CurtisLusmore:
I have an argument for one of my commands which represents a path/ID to a resource. I want to be able to provide tab-completion to allow easily cycling between existing paths/IDs, but I don't want these values to be enumerated if I run `--help` for my command.

I'm thinking this could be an option on the `CompletionItem` object, like a `boolean` property called `Hidden`. You could then do something this like.

```csharp
var myarg = new Option<string>("--id", "The id of the item to get");
myarg.AddCompletions(ctx => GetIdSuggestions(ctx));

private IEnumerable<CompletionItem> GetIdSuggestions(CompletionContext ctx)
    => GetExistingIds(prefix: ctx.WordToComplete).Select(id => new CompletionItem(id) { Hidden = true });
```

and then when you attempt to run the command from the terminal

```
app command --id <TAB>
```
it will cycle through existing IDs returned by my method, but when I run

```
app command --help
```

It will still only print out

```
Options:
 --id <id>   The id of the item to get
```

rather than

```
Options:
 --id <id1|id2|id3|id4|...>   The id of the item to get
```

CurtisLusmore:
Thanks! For the benefit of those who stumble upon this issue trying to do the same thing, the solution is to do the following
```csharp
var myarg = new Argument<string>("arg", "An argument with completions") { HelpName = "arg" };
myarg.AddCompletions(/*...*/);
var myopt = new Option<string>("--opt", "An option with completions") { ArgumentHelpName = "opt" };
myopt.AddCompletions(/*...*/);
```

which will produce the following output when you run `--help`

```
Arguments:
  <arg>  An argument with completions

Options:
  --opt <opt>  An option with completions
```

PR #2067, opened 2023-02-24
make Symbol.Name mandatory, non-nullable and readonly

### LINKED ISSUE #1895: Does an argument with no name make sense?
```C#
    public class Argument<T> : Argument, IValueDescriptor<T>, IValueDescriptor, ICompletionSource 
    {
        // Does an argument with no name make sense?
        public Argument();
```

### LINKED ISSUE #2038: If Argument.HelpName was set, HelpBuilder.Default calls GetCompletions but ignores the result
HelpBuilder.Default.GetArgumentUsageLabel(Argument argument) calls `argument.GetCompletions()` unless `argument.ValueType` is `bool` nor `bool?`. However, if `argument.HelpName` is a non-empty string, then GetArgumentUsageLabel ignores the result of GetCompletions. In that situation, it would be better if GetArgumentUsageLabel skipped the GetCompletions call. Doing so would improve performance if the application has slow completion sources.

<https://github.com/dotnet/command-line-api/blob/6524142f1398ed78eadfb9b802ba6492c6af63c4/src/System.CommandLine/Help/HelpBuilder.Default.cs#L66-L92>

This scenario was not optimized in <https://github.com/dotnet/command-line-api/pull/1367> when HelpBuilder was changed to display Argument.HelpName.

Fixing this might solve <https://github.com/dotnet/sdk/issues/27374>, in which .NET SDK uses a completion callback that ends up resolving MSBuild SDKs from NuGet, even though the completions won't be displayed in help. Such a slow completion callback could still be intolerable in `dotnet-suggest`, though.

### LINKED ISSUE #2051: Providing name should be mandatory when creating Argument<T> and Option<T>
Based on our discussion in #2046 we have agreed that:

- [ ] `Argument<T>` should require a mandatory name to be provided in every `ctor`:
https://github.com/dotnet/command-line-api/blob/76437b04511d88543df5cde2c7910e8d40e30888/src/System.CommandLine/Argument%7BT%7D.cs#L19
- [ ] `Argument.HelpName` needs to be removed and `Argument.Name` should be used instead
- [ ] `Option<T>` should require a mandatory name to be provided in explicit way in every `ctor`. Currently, when the user provides multiple aliases we are choosing the longest one and removing the prefix. It should be explicit.
https://github.com/dotnet/command-line-api/blob/76437b04511d88543df5cde2c7910e8d40e30888/src/System.CommandLine/Option%7BT%7D.cs#L23-L24
- [ ] `Option.ArgumentHelpName` needs to be removed

This issue might seem simple to implement, but I expect it to require a LOT of test changes.

adamsitnik:
The goal of this PR it to make `Name` a mandatory, readonly and non empty value for every `Symbol`.

We need that to implement #2052: `ParseResult.GetValue<T>(string name)`, which is going to allow our users to use the symbol name to fetch parsed value without the need of storing reference to Symbol.

Before:

```cs
Argument<string> pathArgument = new();
Option<bool> flagOption = new(new[] { "--flag", "-f" });

RootCommand command = new()
{
    pathArgument,
    flagOption
};

ParseResult parseResult = command.Parse(args);

string path = parseResult.GetValue(pathArgument);
bool flag = parseResult.GetValue(flagOption);
```

After

```cs
RootCommand command = new()
{
    new Argument<string>("path"),
    new Option<bool>("flag", new[] { "--flag", "-f" })
};

ParseResult parseResult = command.Parse(args);

string path = parseResult.GetValue("path");
bool flag = parseResult.GetValue("flag");
```

Another motivation is to remove `Option.ArgumentHelpName` which is confusing, as the fact that `Option` wraps an `Argument` is an implementation detail.
Last but not least, `Argument` being a parameter without a name, but having both `Name` and `HelpName` property was also confusing.

This required changes in help. So far, the value of first column was `Argument.HelpName` if present, completions (if any) or `Argument.Name` otherwise.

https://github.com/dotnet/command-line-api/blob/1e15ab1629ad13dee96b7a0fdb6c554f3f0528db/src/System.CommandLine/Help/HelpBuilder.Default.cs#L73-L86

Not everybody like that: #1976 and some of our users were setting `Argument.HelpName` just to avoid having all completions displayed in help.

Now the first column contains the Name (because it is always present), while second chooses Description. If description is not provided, it's using the completions.

fixes #1895
fixes #2038
fixes #2051 

unblocks #2052 (required data will be always present)
unblocks #2053 (avoids conflicts related to ctors changes)


src/System.CommandLine.Tests/ParseResultTests.cs
```diff
@@ -42,7 +42,7 @@ public void FindResult_can_be_used_to_check_the_presence_of_an_option()
         [Fact]
         public void FindResultFor_can_be_used_to_check_the_presence_of_an_implicit_option()
         {
-            var option = new Option<int>(new[] { "-c", "--count" }, () => 5);
+            var option = new Option<int>("--count", new[] { "-c", "--count" }, () => 5);
```
Keboo, line 45:
```suggestion
            var option = new Option<int>("count", new[] { "-c", "--count" }, () => 5);
```

src/System.CommandLine.Tests/OptionTests.cs
```diff
@@ -2,37 +2,27 @@
...
-            option.Name.Should().Be("myname");
+            option.Name.Should().Be("short");
         }
 
         [Fact]
-        public void When_an_option_has_several_aliases_then_the_longest_alias_is_its_name()
+        public void Option_names_can_contain_prefixes()
         {
-            var option = new Option<string>(new[] { "myname", "m" });
+            var option = new Option<string>("--myname", new[] { "myname", "m" });
```
Keboo, line 23:
```suggestion
            var option = new Option<string>("myname", new[] { "myname", "m" });
```

src/System.CommandLine.Tests/OptionTests.cs
```diff
@@ -70,39 +60,39 @@ public void A_prefixed_alias_can_be_added_to_an_option()
         [Fact]
         public void Option_aliases_are_case_sensitive()
         {
-            var option = new Option<string>(new[] { "-o" });
+            var option = new Option<string>("-o", new[] { "-o" });
```
Keboo, line 63:
```suggestion
            var option = new Option<string>("o", new[] { "-o" });
```

jonsequitur:
I'm hesitant on this change. 

* It changes the meaning of `Option<T>.Name` significantly and in way that will require code changes for every current user of System.CommandLine that, unlike our other breaking changes, won't be detectable by compiler errors.

* It implies (based on past behaviors of all versions of System.CommandLine) that the `Name` property is no longer an identifier in the grammar. Take this example:

    ```csharp
    RootCommand command = new()
    {
        new Argument<string>("path"),
        new Option<bool>("flag", new[] { "--flag", "-f" })
    };

    ParseResult parseResult = command.Parse(args);

    string path = parseResult.GetValue("path");
    bool flag = parseResult.GetValue("flag");
    ```

    I assume we would treat this as an invalid command line:

    ```console
    > myapp flag
    ```

    I can see how this change potentially allows us to disambiguate lookups by alias (e.g. do I look up using `parseResult.GetValue("-f")` versus `parseResult.GetValue("--flag")`?) but it only does so by introducing a third option `"flag"`. I suspect people are still more like to use `"--flag"` as the lookup, especially given how long the API was used that way.

adamsitnik:
> It changes the meaning of `Option<T>.Name` significantly and in way that will require code changes for every current user of System.CommandLine that, 

So far to get `Option<T>.Name`, we were searching for longest alias and removing the prefix. I am just proposing to make it a mandatory value provided by the user.

> unlike our other breaking changes, won't be detectable by compiler errors.

You mean lack of compiler errors because all the ctors accept just strings?

I've a proposal for how we could simplify that: https://github.com/dotnet/command-line-api/issues/2053#issuecomment-1450831070

> that the Name property is no longer an identifier in the grammar

I am not sure if I understand. What do you mean by grammar? The parsing rules? Argument is a parameter with no name, so it was not used for parsing. Options name was not the actual alias, but a shortened version of it. So it was also not used for parsing. 

Could you please elaborate more on that?

> I can see how this change potentially allows us to disambiguate lookups by alias

I can add lookups by alias, especially since some products like SDK provide localized names for their symbols.

jonsequitur:
> Options name was not the actual alias, but a shortened version of it. So it was also not used for parsing.

`Option.Name` is set by specifying e.g. `new Option<string>("--blah")` and we've long regarded it as a bug that `Option.Name` returns it without the prefix. It's never been necessary to pass in aliases. So in that sense, the value specified in the constructor (which ideally should be used as the `Name` property without stripping the prefix) is an identifier in your program's command line grammar.

jonsequitur:
>> unlike our other breaking changes, won't be detectable by compiler errors.

> You mean lack of compiler errors because all the ctors accept just strings?

Yes.

I looked at the comment you mentioned and it includes a sample like this:

```csharp
new Option<bool>("ucr", "--ucr", "--use-current-runtime")
```

I think the repetition of `ucr` here shouldn't be necessary. It solves an indexing problem by putting some extra trivia into the API that I think is a little more ceremony than we'd prefer.

jonsequitur:
> I can add lookups by alias, especially since some products like SDK provide localized names for their symbols.

This sounds like a new scenario that deserves some discussion. Localizing actual identifiers sounds like a pretty uncommon case, and needing to know the localized identifier to perform a lookup sounds like it would invite a lot of bugs.

adamsitnik:
@jonsequitur big thanks for a very detailed answer, it was very helpful.

I think that my understanding of what `Option.Name` really is (or what it was supposed to be originally) was wrong because:

1. Unintuitive prefix removal, using the `Name` only in Help and when creating parse errors, but not for parsing as the main ID.
2. Always adding `Name` to the collection of aliases and treating it also as an alias (it makes some parts of the code simpler).


**Should separate `Name` and `Aliases` concepts, namely don't include `Name` in `Aliases` collection?**

It would simplify the design of symbol types, as we would not need to check whether given alias can be removed in `RootCommand.RemoveAlias`:

https://github.com/dotnet/command-line-api/blob/1e15ab1629ad13dee96b7a0fdb6c554f3f0528db/src/System.CommandLine/RootCommand.cs#L61-L67

and could be expressed as a single property, which would be a step forward removing `IdentifierSymbol` type.

```diff
public string Name { get; }

+ public HashSet<string> Aliases { get; }
- public IReadOnlyCollection<string> Aliases { get; }
- public void AddAlias(string alias)
- private protected virtual void RemoveAlias(string alias)
- public bool HasAlias(string alias)
```

We would also not need to allocate a hashset of strings for symbols with no aliases (perf).
We could also easily differentiate the name and aliases.

And it would make it way easier to introduce hidden aliases, as it would be just a single new property:

```cs
public HashSet<string> HiddenAliases { get; }
```

We could also use a very fancy C# syntax:

```cs
new Option<bool>("--help", "-h")
{
    Description = "Show command line help.",
    HiddenAliases = { "-?", "help" }
}
```

The downside is that all the consumer of `Aliases` property would need to take this into account:

```diff
+ tokens.Add(command.Name, new Token(command.Name, TokenType.Command, command, Token.ImplicitPosition));

foreach (string commandAlias in command.Aliases)
{
    tokens.Add(
        commandAlias,
        new Token(commandAlias, TokenType.Command, command, Token.ImplicitPosition));
}
```

But to be honest I don't expect 99% of our users to be hit by that change.


> I looked at the comment you mentioned and it includes a sample like this:

```csharp
new Option<bool>("ucr", "--ucr", "--use-current-runtime")
```

You are right, this example is wrong, it should not work in that way.

It's wrong because I've tried to replace `ArgumentHelpName` and `Name` with just `Name` without having good understanding of how `ArgumentHelpName` is used by the customers.

After looking at more SDK command help examples I am convinced that we need to keep `HelpName`.

```log
Arguments:
  <PROJECT | SOLUTION>  The project or solution file to operate on. If a file is not specified, the command will search
                        the current directory for one.
```

Setting `Argument.Name` to "PROJECT | SOLUTION" feels weird to me.
I would just rename it to `HelpName` from `ArgumentHelpName` and make it a part of every `Symbol` type.

Updated proposal:

<details>

```cs
using System.Collections;
using System.Runtime.InteropServices;

namespace SymbolCtors
{
    internal class Program
    {
        {
            // Sample: "dotnet build"
            Command dotnetBuild = new Command("build")
            {
                new Argument<FileInfo?>("file")
                {
                    Description = "The project or solution file to operate on. If a file is not specified, the command will search the current directory for one.",
                    HelpName = "PROJECT | SOLUTION",
                    DefaultValueFactory = (argumentResult) =>
                    {
                        FileInfo? first = new DirectoryInfo(Directory.GetCurrentDirectory()).EnumerateFiles("*.sln|*.proj").FirstOrDefault();

                        if (first is null)
                        {
                            argumentResult.AddError("Unable to find project or a solution file");
                        }
                        
                        return first;
                    }
                },
                new Option<bool>("--use-current-runtime", "--ucr")
                {
                    Description = "Use current runtime as the target runtime."
                },
                new Option<string>("--framework", "-f")
                {
                    Description = "The target framework to build for. The target framework must also be specified in the project file.",
                    HelpName = "FRAMEWORK",
                },
                new Option<Architecture>("--arch", "-a")
                {
                    Description = "The target architecture.",
                    HelpName = "arch",
                    ValueParser = (argumentResult) => Enum.Parse<Architecture>(argumentResult.Tokens[0]),
                    DefaultValueFactory = (_) => RuntimeInformation.OSArchitecture,
                },
                new Option<string>("--os")
                {
                    Description = "The target operating system.",
                    HelpName = "os",
                    ValueParser = GetOsName,
                    DefaultValueFactory = GetOsName,
                },
                new Option<bool>("--help")
                {
                    Description = "Show command line help.",
                    // just to show that this syntax would be allowed if Aliases was a mutable collection wiith .Add method
                    Aliases = { "-h", "-?" }
                }
            };

            // an example of both a parser and a custom value provider
            static string GetOsName(ArgumentResult argumentResult)
            {
                if (argumentResult.Tokens.Count > 0)
                {
                    string provided = argumentResult.Tokens[0];

                    if (provided is "win" or "lin" or "osx")
                    {
                        return provided;
                    }

                    argumentResult.AddError($"Unknown Operating System: {provided}");
                }

                return OperatingSystem.IsWindows() ? "win" : OperatingSystem.IsLinux() ? "lin" : "osx";
            }
        }
    }

    public class Symbol
    {
        public string Name { get; }

        public string? Description { get; set; }

        public string? HelpName { get; set; }

        private protected Symbol(string name)
        {
            Name = name;
        }
    }

    public class Argument<T> : Symbol
    {
        public Argument(string name) : base(name)
        {
        }

        public Func<ArgumentResult, T>? DefaultValueFactory { get; set; }
        public Func<ArgumentResult, T>? ValueParser { get; set; }
    }

    public class Option<T> : Symbol
    {
        private readonly Argument<T> _argument;
        private HashSet<string>? _aliases, _hiddenAliases;

        public HashSet<string> Aliases => _aliases ??= new(StringComparer.Ordinal);

        public HashSet<string> HiddenAliases => _hiddenAliases ??= new(StringComparer.Ordinal);

        // Using "params" to avoid forcing users to allocate a new array
        //
        // An alternative is having a ctor with optional arguments
        // public Option(string name, string? firstAlias = null, string? secondAlias = null) : base(name)
        public Option(string name, params string[] aliases) : base(name)
        {
            _argument = new Argument<T>(name);

            _aliases = new HashSet<string>(aliases, StringComparer.Ordinal);
        }

        public Func<ArgumentResult, T>? DefaultValueFactory
        {
            get => _argument.DefaultValueFactory;
            set => _argument.DefaultValueFactory = value;
        }

        public Func<ArgumentResult, T>? ValueParser
        {
            get => _argument.ValueParser;
            set => _argument.ValueParser = value;
        }
    }

    public class Command : Symbol, IEnumerable<Symbol>
    {
        private readonly List<Symbol> _symbols;
        private HashSet<string>? _aliases, _hiddenAliases;

        public HashSet<string> Aliases => _aliases ??= new(StringComparer.Ordinal);
        public HashSet<string> HiddenAliases => _hiddenAliases ??= new(StringComparer.Ordinal);

        public Command(string name, string? description = null) : base(name)
        {
            Description = description;
        }

        public void Add(Symbol symbol) => _symbols.Add(symbol);
        public IEnumerator<Symbol> GetEnumerator() => _symbols.GetEnumerator();
        IEnumerator IEnumerable.GetEnumerator() => _symbols.GetEnumerator();
    }

    public class ArgumentResult
    {
        public List<string> Tokens { get; } = new();

        public void AddError(string errorMessage) { }
    }
}
```

</details>

adamsitnik:
Closing in favor of #2073

PR #2073, opened 2023-03-03
Name and aliases separation

### LINKED ISSUE #1895: Does an argument with no name make sense?

### LINKED ISSUE #2038: If Argument.HelpName was set, HelpBuilder.Default calls GetCompletions but ignores the result

### LINKED ISSUE #2051: Providing name should be mandatory when creating Argument<T> and Option<T>

### LINKED ISSUE #2053: Simplify custom parsers and default value factories

adamsitnik:
It's a second attempt (the first one was #2067). For motivation please go to https://github.com/dotnet/command-line-api/issues/2053#issuecomment-1451992830

Changes:
* Separation of `Name` and `Aliases` concepts. Name needs to be always provided in an explicit way and can not be empty or changed later. We don't remove prefixes from Name anymore (as is). Name is also not added to `Aliases` anymore. 
* Simplification of the `Symbol` constructors. Each symbol type provides now only a single public ctor. The ctor requires to provide only the mandatory data (name), everything else needs to be provided by using the properties.
* Removal of `SetDefaultValue`, `SetDefaultValueFactory` methods, they have been repalced with a single property. 
* Removal of `IdentifierSymbol` type. The aliases are now represented as a single property. It makes it very easy to implement `HiddenAliases` feature. Once this PR gets merged I am going to create an issue for that (or find an existing one and add description)
* Rename of `Option.ArgumentHelpName` to just `HelpName` (the fact that we are using an Argument is an implementation detail)

Breaking change: users who were using `Option(string name, string description)` ctor, are now going to hit a silent error, as compiler won't warn them that `description` has became an `alias` now.
We are soon going to rename `Option` to `CliOpton` and expect the end users to not hit the silent error (we are currently not releasing any packages to nuget.org).
Those who are going to hit this issue are most likely going to get an exception at runtime, as description typically contains whitespaces and aliases can't.

fixes #1895
fixes #2038
fixes #2051 
fixes #2053

unblocks #2052 (required data will be always present)


src/System.CommandLine/Argument.cs
```diff
@@ -19,24 +19,10 @@ public abstract class Argument : Symbol, IValueDescriptor
         private List<Func<CompletionContext, IEnumerable<CompletionItem>>>? _completionSources = null;
         private List<Action<ArgumentResult>>? _validators = null;
 
-        /// <summary>
-        /// Initializes a new instance of the Argument class.
-        /// </summary>
-        protected Argument()
+        private protected Argument(string name) : base(name, canContainWhitespaces: true)
```
adamsitnik, line 22: since arguments name is not used for parsing, I've decided to keep allowing the whitespaces. I don't have a strong opinion about that.

src/System.CommandLine/Help/HelpBuilder.Default.cs
```diff
@@ -50,63 +50,46 @@ public static string GetArgumentDefaultValue(Argument argument)
...
-                {
-                    return $"<{argument.Name}>";
-                }
-                else
-                {
-                    return "";
-                }
+                return $"<{argument.HelpName}>";
             }
+            else if (!argument.IsBoolean() && argument.CompletionSources.Count > 0)
```
adamsitnik, line 58: this change fixes #2038

First of all we check `HelpName`, if it's not provided and there are some `CompletionSources` defined for given `Argument`, then we are getting the completions.

src/System.CommandLine.Tests/CompletionTests.cs
```diff
@@ -41,9 +41,9 @@ public void Command_GetCompletions_returns_available_option_aliases()
         {
             var command = new Command("command")
             {
-                new Option<string>("--one", "option one"),
-                new Option<string>("--two", "option two"),
-                new Option<string>("--three", "option three")
+                new Option<string>("--one") { Description = "option one" },
+                new Option<string>("--two") { Description = "option two" },
+                new Option<string>("--three") { Description = "option three" },
```
jonsequitur, line 46: The change to the `params` constructor for aliases results in some extra verbosity here and I wonder if it's the right tradeoff, given that people using `Option` set a description much more often than aliases.

src/System.CommandLine/Argument.cs
jonsequitur: I agree.

src/System.CommandLine/Symbol.cs
```diff
@@ -11,11 +12,11 @@ namespace System.CommandLine
     /// </summary>
     public abstract class Symbol
     {
-        private protected string? _name;
         private ParentNode? _firstParent;
 
-        private protected Symbol()
+        private protected Symbol(string name, bool canContainWhitespaces = false)
```
jonsequitur, line 17:
```suggestion
        private protected Symbol(string name, bool allowWhitespace = false)
```

src/System.CommandLine.Tests/OptionTests.cs
```diff
@@ -9,160 +9,155 @@
...
+            var option = new Option<string>("o");
 
-            option.HasAlias("o").Should().BeTrue();
+            option.Name.Should().Be("o");
+            option.Aliases.Should().BeEmpty();
         }
 
         [Fact]
-        public void An_option_must_have_at_least_one_alias()
+        public void An_option_doesno_not_need_to_have_at_least_one_alias()
```
jonsequitur, line 103:
```suggestion
        public void An_option_does_not_need_to_have_at_any_aliases()
```

src/System.CommandLine.Tests/OptionTests.cs
```diff
@@ -210,39 +205,34 @@ public void When_options_use_different_prefixes_they_still_work(string prefix)
...
         [Fact]
         public void Option_T_default_value_can_be_set_via_the_constructor()
         {
-            var option = new Option<int>(
-                "-x",
-                parseArgument: parsed => 123,
-                isDefault: true);
+            var option = new Option<int>("-x")
+            {
+                DefaultValueFactory = _ => 123
```
jonsequitur, line 234: I'd love to come up with a better name for this.

src/System.CommandLine/Help/HelpBuilder.Default.cs
```diff
@@ -50,63 +50,46 @@ public static string GetArgumentDefaultValue(Argument argument)
...
         /// </summary>
         /// <param name="symbol">The symbol to get a help item for.</param>
-        /// <param name="context">The help context, used for localization purposes.</param>
         /// <returns>Text to display.</returns>
-        public static string GetIdentifierSymbolUsageLabel(IdentifierSymbol symbol, HelpContext context)
+        public static string GetIdentifierSymbolUsageLabel(Command symbol)
+            => GetIdentifierSymbolUsageLabel(symbol, symbol._aliases);
+
+        /// <inheritdoc cref="GetIdentifierSymbolUsageLabel(Command)"/>
+        public static string GetIdentifierSymbolUsageLabel(Option symbol)
```
jonsequitur, line 85: Should the word "identifier" be removed from these two methods? Since the type was removed, the term might no longer have significance in the API.

src/System.CommandLine.Tests/CompletionTests.cs
adamsitnik: We have at least three ways of setting the aliases:
* what we had so far, which required the user to allocate an array in an explicit way
```cs
new Option<bool>("--help", new [] { "-h", "-?" }, "description");
```
* using params, which looks way nicer, but requires the description to be set via a property. TBH I am not sure if it's a bad thing, but it's very subjective.
```cs
new Option<bool>("--help", "-h", "-?")
{
    Description = "description"
};
```
* using collection initalizer syntax. IMO it looks great but I would expect very few C# devs to know about this duck typing.
```cs
new Option<bool>("--help", "description")
{
    Aliases = { "-h", "-?" }
};
```

@jonsequitur if you believe it's worth it, I can remove the params ctor and revert this part of the changes. My personal preference would be to merge it in the new form and eventually revisit it after API design review.

src/System.CommandLine/Help/HelpBuilder.Default.cs
adamsitnik: TBH I expect that we are going to remove it very soon and make help more customizable in V2.

adamsitnik: I've renamed it.

src/System.CommandLine.Tests/CompletionTests.cs
jonsequitur: My hesitation is that the more common case is now more verbose, and it will be a very large source-breaking change. The reasoning makes sense though. I'd be curious to hear more opinions.

adamsitnik: > I'd be curious to hear more opinions.

@KathleenDollard @Keboo @KalleOlaviNiemitalo what is your opinion on that?

jonsequitur (approved):

src/System.CommandLine.Tests/CompletionTests.cs
adamsitnik: I am going to merge the PR now, but please don't hesitate to share your feedback.

jeffhandley: In my first example above, I meant to explicitly state that there'd still be the single-string overload too. Calling that out in case it wasn't inferred.

```c#
Option(string name, string description);
Option(string[] nameAndAliases, string description);
```

PR #2080, opened 2023-03-07
Minor token improvements

adamsitnik:
While creating the API design proposal I've noticed that:

- public `Token` ctor accepts `Symbol`, but the property is internal. Some users already asked for exposing in the past (#1815)
- `Token` does not need to override the quality operators, I have most likely introduced them when I've made it a struct in the past

### DIFF
--- a/src/System.CommandLine.Tests/TokenTests.cs
+++ b/src/System.CommandLine.Tests/TokenTests.cs
@@ -1,0 +1,53 @@
+﻿using FluentAssertions;
+using System.CommandLine.Parsing;
+using Xunit;
+
+namespace System.CommandLine.Tests
+{
+    public class TokenTests
+    {
+        [Fact]
+        public void Tokens_are_equal_when_they_use_same_value_type_and_symbol()
+        {
+            Option<int> count = new ("--count");
+            Token token = new (count.Name, TokenType.Option, count);
+            Token same = new (count.Name, TokenType.Option, count);
+
+            token.Equals(same).Should().BeTrue();
+            token.Equals((object)same).Should().BeTrue();
+        }
+
+        [Fact]
+        public void Tokens_are_not_equal_when_they_do_not_use_same_value_type_and_symbol()
+        {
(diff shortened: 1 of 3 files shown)


src/System.CommandLine/Parsing/Token.cs
```diff
@@ -46,7 +46,7 @@ internal Token(string? value, TokenType type, Symbol? symbol, int position)
         /// <summary>
         /// The Symbol represented by the token (if any).
         /// </summary>
-        internal Symbol? Symbol { get; set; }
+        public Symbol? Symbol { get; internal set; }
```
jonsequitur, line 49: We should test that this is correctly reassigned when `OnlyTake` is used.

src/System.CommandLine/Parsing/Token.cs
```diff
@@ -59,21 +59,5 @@ internal Token(string? value, TokenType type, Symbol? symbol, int position)
 
         /// <inheritdoc />
         public override string ToString() => Value;
-
```
andrewimcclement, line 62: Standard guidance (https://learn.microsoft.com/en-us/dotnet/api/system.iequatable-1?view=netstandard-2.0#notes-to-implementers) does suggest we should keep the equality operators for consistency, even if we aren't using them at the moment.

From a personal perspective, I like having the equality operators available since it makes the code simpler when dealing with nullable types, and using the equality operators helps avoid (via compilation errors) accidentally making typos that result in calling `type1.Equals(type2)` (calling into `Type1.Equals(object)`, always returning false).

src/System.CommandLine.Tests/TokenTests.cs
```diff
@@ -0,0 +1,53 @@
+﻿using FluentAssertions;
+using System.CommandLine.Parsing;
+using Xunit;
+
+namespace System.CommandLine.Tests
```
Keboo, line 5: Are we leveraging file scoped namespaces?
I see at least one file was converted to use them by @jonsequitur in #2226

Issue #2082, opened 2023-03-07
Why Option exposes IsRequired while it also exposes Arity?


adamsitnik:
Currently there are two ways of saying that given `Option` is required:
* by setting `option.IsRequired` to true
* by setting `option.Arity` to an arity with `MinimumNumberOfValues` >= 1.

We also have two different errors for both cases: `OptionRequiredArgumentMissing` and `RequiredOptionWasNotProvided`.

I need to understand it better to provide good implementation for #2052

jonsequitur:
`Option.Arity` describes how many argument tokens the option requires **_if the option is provided_**. It only has significance when the option is present on the command line being parsed.

<img width="480" alt="image" src="https://user-images.githubusercontent.com/547415/223561278-079a2735-f41e-4a1b-8083-142fb5e225e9.png">

`Option.IsRequired` indicates whether an option must be provided.  Only when it's provided are the arity rules meaningful.

<img width="437" alt="image" src="https://user-images.githubusercontent.com/547415/223561179-cb6fc61c-d194-45df-9017-d8080f1d681c.png">

jonsequitur:
> @jonsequitur how does it work when it comes to Argument, which exposes Arity but no IsRequired?

Options are optional by default, whereas arguments are required by default. 

The test does look wrong. The default arity for that argument is {1,1} which means there should be an error there.

PR #2083, opened 2023-03-08
introduce ParseResult.GetValue<T>(string name)

### LINKED ISSUE #2052: Introduce ParseResult.GetValue<T>(string name)

adamsitnik:
fixes #2052

### DIFF
--- a/src/System.CommandLine.Tests/GetValueByNameParserTests.cs
+++ b/src/System.CommandLine.Tests/GetValueByNameParserTests.cs
@@ -2,7 +2,6 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using FluentAssertions;
-using System.Linq;
 using Xunit;
 using Xunit.Abstractions;
 
@@ -46,20 +45,20 @@ namespace System.CommandLine.Tests
         {
             RootCommand command = new()
             {
-                new Option<int>("opt", new[] { "-i", "--integer" }),
+                new Option<int>("--integer", "-i"),
                 new Command("inner1")
                 {
-                    new Option<int>("opt", new[] { "-i", "--integer" }),
+                    new Option<int>("--integer", "-i"),
                     new Command("inner2")
                     {
-                        new Option<int>("opt", new[] { "-i", "--integer" })
+                        new Option<int>("--integer", "-i")
                     }
                 }
             };
 
             ParseResult parseResult = command.Parse("-i 1 inner1 --integer 2 inner2 -i 3");
 
-            parseResult.GetValue<int>("opt").Should().Be(3);
+            parseResult.GetValue<int>("--integer").Should().Be(3);
         }
 
         [Fact]
@@ -67,20 +66,23 @@ namespace System.CommandLine.Tests
         {
             RootCommand command = new()
             {
-                new Option<int>("opt", new[] { "-i", "--integer" })
+                new Option<int>("--integer", "-i")
             };
 
             ParseResult parseResult = command.Parse("");
 
-            parseResult.GetValue<int>("opt").Should().Be(default);
+            parseResult.GetValue<int>("--integer").Should().Be(default);
         }
 
         [Fact]
-        public void When_argument_value_is_not_parsed_then_default_value_is_returned()
+        public void When_optional_argument_is_not_parsed_then_default_value_is_returned()
         {
             RootCommand command = new()
             {
                 new Argument<int>("arg")
+                {
+                    Arity = ArgumentArity.ZeroOrOne
+                }
             };
 
             ParseResult parseResult = command.Parse("");
@@ -93,7 +95,7 @@ namespace System.CommandLine.Tests
         {
             RootCommand command = new()
             {
-                new Option<int>("required", new[] { "-i", "--integer" })
+                new Option<int>("--required")
                 {
                     IsRequired = true
                 }
@@ -101,12 +103,12 @@ namespace System.CommandLine.Tests
 
             ParseResult parseResult = command.Parse("");
 
-            Action getRequired = () => parseResult.GetValue<int>("required");
+            Action getRequired = () => parseResult.GetValue<int>("--required");
 
             getRequired
                 .Should()
                 .Throw<InvalidOperationException>()
-                .Where(ex => ex.Message == LocalizationResources.RequiredOptionWasNotProvided("required"));
+                .Where(ex => ex.Message == LocalizationResources.RequiredOptionWasNotProvided("--required"));
         }
 
         [Fact]
@@ -116,7 +118,7 @@ namespace System.CommandLine.Tests
             {
                 new Argument<int>("required")
                 {
-                    Arity = new ArgumentArity(1, 1)
+                    Arity = ArgumentArity.ExactlyOne
                 }
             };
 
@@ -127,7 +129,20 @@ namespace System.CommandLine.Tests
             getRequired
                 .Should()
                 .Throw<InvalidOperationException>()
-                .Where(ex => ex.Message == LocalizationResources.RequiredOptionWasNotProvided("required"));
+                .Where(ex => ex.Message == LocalizationResources.RequiredArgumentMissing(parseResult.FindResultFor(command.Arguments[0])));
+        }
+
+        [Fact]
+        public void When_non_existing_name_is_used_then_exception_is_thrown()
+        {
+            ParseResult parseResult = new Command("noSymbols").Parse("");
+
+            Action getRequired = () => parseResult.GetValue<int>("required");
+
+            getRequired
+                .Should()
+                .Throw<InvalidOperationException>()
+                .Where(ex => ex.Message == LocalizationResources.RequiredArgumentMissing(parseResult.FindResultFor(command.Arguments[0])));
         }
     }
 }
--- a/src/System.CommandLine.Tests/GetValueByNameParserTests.cs
+++ b/src/System.CommandLine.Tests/GetValueByNameParserTests.cs
@@ -1,0 +1,133 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using FluentAssertions;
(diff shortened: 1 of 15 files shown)


src/System.CommandLine/Parsing/CommandResult.cs
```diff
@@ -165,15 +210,16 @@ private void ValidateArguments(bool completeValidation)
...
                     SymbolResultTree[argument] = argumentResult;
-                }
-                else if (argument.Arity.MinimumNumberOfValues > 0)
-                {
-                    AddError(LocalizationResources.RequiredArgumentMissing(this));
-                    continue;
+
+                    if (!argument.HasDefaultValue && argument.Arity.MinimumNumberOfValues > 0)
+                    {
+                        argumentResult.AddError(LocalizationResources.RequiredArgumentMissing(argumentResult));
```
adamsitnik, line 220: this fixes the issue I've mentioned in https://github.com/dotnet/command-line-api/issues/2082#issuecomment-1460476519 : when a required argument is missing, we report the error for the ArgumentResult, not for CommandResult

src/System.CommandLine/ParseResult.cs
```diff
@@ -129,6 +129,14 @@ symbol switch
         public T? GetValue<T>(Option<T> option)
             => RootCommandResult.GetValue(option);
 
+        /// <summary>
+        /// Gets the parsed or default value for the specified symbol name, in the context of parsed command.
```
adamsitnik, line 133: this is important: the implementation provides only values for the parsed command, not for the entire symbol tree.

src/System.CommandLine.Tests/GetValueByNameParserTests.cs
```diff
@@ -0,0 +1,148 @@
...
+        }
+
+        protected override T GetValue<T>(ParseResult parseResult, Option<T> option)
+            => parseResult.GetValue<T>(option.Name);
+
+        protected override T GetValue<T>(ParseResult parseResult, Argument<T> argument)
+            => parseResult.GetValue<T>(argument.Name);
+
+        [Fact]
+        public void In_case_of_argument_name_conflict_the_value_which_belongs_to_the_last_parsed_command_is_returned()
```
jonsequitur, line 23: A few additional test cases come to mind:

* What happens when an option and an argument name are the same...
  * at the same level of the tree?
  * at different levels of the tree?
* What happens when symbols with the same name exist at different levels of the tree and the inner one is not provided? Does it fall back to the outer one's value?
  * Does this vary depending on whether the inner one has a default value defined?
* How do customer parsers behave when...
  * ...parsing an inner symbol `-x` and looking up a value for `-x` when there's an outer symbol `-x`? 
  * ...parsing an outer symbol `-x` and looking up a value for `-x` when there's an inner symbol `-x`?
  * ...parsing symbol `-a` and looking up a value for symbol `-x`...
    * ...which is above it in the tree
    * ...which is below it in the tree

I have opinions on what some of these behaviors should be but I'll wait so I don't anchor anyone.

adamsitnik: > * What happens when an option and an argument name are the same at the same level of the tree?
  
We should throw.
  
> * What happens when an option and an argument name are the same at different levels of the tree?

The new methods provides only values for the parsed command (`parseResult.CommandResult`), not for the entire symbol tree.

> * What happens when symbols with the same name exist at different levels of the tree and the inner one is not provided? Does it fall back to the outer one's value?

The outer value will never be returned, only inner if provided or has default.

> How do customer parsers behave when...

Such scenario is not supported by design (nobody asked for it), the method is exposed only for `ParseResult`, not for `SymbolResult`. It simplifies the design a lot (the cache can be populated just once).

src/System.CommandLine.Tests/GetValueByNameParserTests.cs
```diff
@@ -135,14 +135,88 @@ public void When_required_argument_value_is_not_parsed_then_an_exception_is_thro
...
+            const string nonExistingName = "nonExisting";
+            Command command = new ("noSymbols");
+            ParseResult parseResult = command.Parse("");
 
-            Action getRequired = () => parseResult.GetValue<int>("required");
+            Action getRequired = () => parseResult.GetValue<int>(nonExistingName);
 
             getRequired
                 .Should()
                 .Throw<InvalidOperationException>()
```
jonsequitur, line 146: `ArgumentException` might be more appropriate here, though I could make an argument for `InvalidOperationException` if the name exists on an unparsed branch of the larger grammar, e.g. under a subcommand that wasn't present in the current result.

src/System.CommandLine.Tests/GetValueByNameParserTests.cs
```diff
@@ -135,14 +135,88 @@ public void When_required_argument_value_is_not_parsed_then_an_exception_is_thro
...
+            };
+
+            ParseResult parseResult = command.Parse("");
+
+            Action getConflicted = () => parseResult.GetValue<int>(sameName);
+
+            getConflicted
+                .Should()
+                .Throw<NotSupportedException>()
+                .Where(ex => ex.Message == $"More than one symbol uses name \"{sameName}\" for command \"{command.Name}\".");
```
jonsequitur, line 171: I feel like we should be detecting this at parser configuration time, e.g. under `ThrowIfInvalid` or earlier.

src/System.CommandLine.Tests/GetValueByNameParserTests.cs
```diff
@@ -135,14 +135,88 @@ public void When_required_argument_value_is_not_parsed_then_an_exception_is_thro
...
+                        DefaultValueFactory = (_) => 456
+                    }
+                }
+            };
+
+            ParseResult parseResult = command.Parse($"outer inner 456");
+            parseResult.GetValue<int>(sameName).Should().Be(456);
+
+            parseResult = command.Parse($"outer 123");
+            parseResult.GetValue<int>(sameName).Should().Be(123);
```
jonsequitur, line 219: Not to reopen the old late binding can of worms, but are there tests for what happens in various cases if `GetValue<T>` is called for `Option<U>`?

src/System.CommandLine/Parsing/CommandResult.cs
```diff
@@ -53,28 +53,20 @@ public sealed class CommandResult : SymbolResult
 
                 if (Command.HasArguments)
                 {
-                    for (int i = 0; i < Command.Arguments.Count; i++)
-                    {
-                        SymbolResultTree.TryGetValue(Command.Arguments[i], out SymbolResult? parsedResult);
-                        cache.Add(Command.Arguments[i].Name, parsedResult);
-                    }
+                    Populate(cache, Command.Arguments);
```
jonsequitur, line 56: If I'm understanding this correctly, a call to `Populate` could be triggered by a custom parser before the complete command line has been parsed. Will symbols parsed afterward be findable by name?

src/System.CommandLine.Tests/GetValueByNameParserTests.cs
adamsitnik: Great catch, I've added the missing tests and fixed the discovered issue.

src/System.CommandLine/Parsing/CommandResult.cs
adamsitnik: A custom parser gets access to `SymbolResult`-derived type, it does not provide a reference to `ParseResult`. The `CommandResult.GetValue<T>(string name)` is internal, so it's also impossible to call outside of S.CL before the parsing has been finished. I can move this method to `ParseResult` itself to avoid confusion.

src/System.CommandLine.Tests/GetValueByNameParserTests.cs
adamsitnik: The name is immutable now, so we should be able to throw at the moment when given symbol is being added to the Command.

src/System.CommandLine/Parsing/CommandResult.cs
jonsequitur: > it's also impossible to call outside of S.CL before the parsing has been finished

The specific call I was wondering about is calls to `ArgumentResult` from within a custom parse delegate.

adamsitnik: With my latest changes it's impossible

jonsequitur (approved):

Issue #2071, opened 2023-03-01
Proposal: CLI Actions 


jonsequitur:
This proposal describes a rethinking of the approach to command invocation that's been in place in System.CommandLine from the beginning. The convention- and reflection-based approach (`CommandHandler.Create(Func<...>)` and overloads) was removed  in past previews in favor of something suited to better performance, trimming, and NativeAOT compilation (`Command.SetHandler(Func<...>)` and overloads). This change reduced the usability of the API by increasing its verbosity. The current proposal is for a design to replace these APIs with something quite different. (For more background and discussion, see #1776.) 

## A few examples

I'll start with a few code examples before discussing the various specific goals of the design.

The core concept adds a new abstraction, roughly similar to the existing `ICommandHandler`, called a `CliAction`. The core of the interface looks like this:

```csharp
public abstract class CliAction
{
    public abstract Task<int> RunAsync(
        ParseResult parseResult, 
        CancellationToken cancellationToken); 
}
```

(The parameters to `RunAsync` assume that the `InvocationContext` class will be combined into `ParseResult`, a change currently under discussion.)

A `CliAction` can be attached to a command (or other symbol--more on that below) very much like a `CommandHandler` today:

```csharp
var rootCommand = new RootCommand();
rootCommand.SetAction(new MyRootCommandAction())
```

For convenience, the action can be set up using a delegate, much like today:

```csharp
rootCommand.SetAction((parseResult, cancellationToken) => 
{
    /* custom logic */ 
});
```

After parsing, you can run the action designated by the command line input like this:

```csharp
ParseResult result = rootCommand.Parse(args);
await result.Action.RunAsync();
```

Just like the existing `InvokeAsync` APIs, this will run the code associated with the command  the end user specified.

While there are a number of similarities to the current command handler API, it's intended to be used somewhat differently, using different `CliAction` types to represent outcomes that you can inspect and alter prior to invocation. 

The key difference can be summed up as: 

**_The `ParseResult.Action` property, set during parsing, makes the behavior designated by the parsed command line explicit, inspectable, and configurable prior to invocation._** 

### Clearly separate parsing from invocation.

The `Parse` and `Invoke` methods currently appear as sibling methods on `Parser` and `Command`, which, while convenient, has led many people to misunderstand that `Parse` is a prerequisite to `Invoke`. The `Invoke` methods internally call `Parse`. It seems helpful to separate these.

### Reduce the number of overloads associated with setting up handlers.

[...]

PR #2092, opened 2023-03-14
The handler should always return an integer

adamsitnik:
Makes progress towards #2071 and meeting BCL requirements


adamsitnik:
@jonsequitur Based on our offline discussion yesterday I know that:
* forcing the users to always return an exit code from handler might be verbose, but on the other hand there is no magic (so far invoke/invokeasync were returning integer, while handlers were returning void) which aligns well with the BCL guidelines
* the removal of `InvocationResult` should be seen as "we don't have a good proposal now, so we are removing it until we have one" rather than "everything will always return an int"


Since I want to introduce `CliAction` today and replace `ICommandHandler` I am going to merge this PR right now to avoid massive conflict.

Issue #2046, opened 2023-02-14
More design questions


adamsitnik:
We are soon going to have another internal round of design discussion for System.CommandLine. I am sharing the list of open questions here so everyone can share their feedback.

# Symbols

## Description

Symbol.Description does not need to be `virtual`: https://github.com/dotnet/command-line-api/pull/2045

Every symbol can provide the description in the `ctor`.

## HelpName

When Argument is parsed, we use its position rather than name. But when we want to refer to it, for example when displaying help, it may need a name.

Currently the type has both `Name` and `HelpName`. Why does `Argument` need both?

## Arity

Can anyone provide a better name?

## Parse(string), Parse(string[])

Both methods are used only for testing purposes in System.CommandLine. They don't provide any possibility to customize the configuraiton used for parsing. They can be easily replaced with one liner:

```cs
new CommandLineBuilder(new Command() { argument }).UseDefaults().Build().Parse($input);
```

I believe that we should remove them or made them internal.

## Validators

Currently a validator is an `Action<ArgumentResult>`. What I don't like about this design is the fact that it allows for modificaitons of the `ArgumentResult`.

Example:

```cs
argument.Validators.Add(argumentResult => argumentResult.OnlyTake(3));
```

It's hard to propose a better design, as some validators need more than just `SymbolResult.Tokens`. Example: validator may want to call `GetValueOrDefault` to validate parsed value.

## ctors

`Argument<T>` provides multiple ways of setting default value: as `T`, `Func<T>` and `Func<ArgumentResult, T>`

```cs
public Argument(Func<T> defaultValueFactory)
    
public Argument(
    string name,
    T defaultValue,
    string? description = null)

public Argument(
    string? name,
    Func<ArgumentResult, T> parse,
    bool isDefault = false,
    string? description = null)
```

What I don't like is the fact that the last ctor requires the user to define whether given `Func<ArgumentResult, T>` is a default value provider or a custom parser. It's confusing.

My understanding is that we need custom parser for complex types and default values for all kinds of types. IMO the latter is way more common than the first.

Do we really need a default value provider that uses `ArgumentResult` as an input?
Do we really need the default value to be lazily provided?

How about having just one ctor:

```cs
public Argument(
    string? name = null,
    string? description = null
    T? defaultValue = null,
    Func<ArgumentResult, T> customParser = null)
```

## SetDefaultValue, SetDefaultValueFactory

Same as above, there are 3 different ways of setting default value:

```cs
[...]

PR #2095, opened 2023-03-15
Introduce CliAction, make it possible for every Symbol to define one

adamsitnik:
Overall it was way easier than I expected.

This PR:

1. Replaces `ICommandHandler` with `CliAction`
2. Renames `Command.SetHandler` with `Command.SetAction`. 
3. Changes the exit code approach: every successful program should return 0, so to avoid verbosity command actions provided by `Command.SetAction` do not need to return an int. In case somebody wants to reuturn an int, they need to implement  their own `CliAction` type (pay for play) 
5. Extends Options and Directives with `Action` property, but not `SetAction` methods (no need to, as it should be rare to have active options and directives that don't need custom types)
6. It does not expose the new action types, I want to do it in a separate PR, where I am going to write tests for them and move the configuration to these types.

Contributes to #2071

### DIFF
--- a/src/System.CommandLine.Tests/OptionTests.cs
+++ b/src/System.CommandLine.Tests/OptionTests.cs
@@ -346,5 +346,38 @@ namespace System.CommandLine.Tests
                 .Should()
                 .BeEquivalentTo(new[] { $"Argument 'Fuschia' not recognized. Must be one of:\n\t'Red'\n\t'Green'" });
         }
+
+        [Fact]
+        public void Every_option_can_provide_a_handler_and_it_takes_precedence_over_command_handler()
+        {
+            bool optionHandlerWasCalled = false;
+            bool commandHandlerWasCalled = false;
+
+            Option<bool> option = new("--test");
+            option.SetHandler((_) =>
+            {
+                optionHandlerWasCalled = true;
+                return 100;
+            });
+            Command command = new Command("cmd")
+            {
+                option
+            };
+            command.SetHandler((_) =>
+            {
+                commandHandlerWasCalled = true;
+                return 200;
+            });
+
+            ParseResult parseResult = command.Parse("cmd --test true");
+
+            parseResult.Handler.Should().NotBeNull();
+            optionHandlerWasCalled.Should().BeFalse();
+            commandHandlerWasCalled.Should().BeFalse();
+
+            parseResult.Invoke().Should().Be(100);
+            optionHandlerWasCalled.Should().BeTrue();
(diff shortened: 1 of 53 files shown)


src/System.CommandLine.Tests/OptionTests.cs
```diff
@@ -346,5 +346,38 @@ public void Option_of_enum_can_limit_enum_members_as_valid_values()
                 .Should()
                 .BeEquivalentTo(new[] { $"Argument 'Fuschia' not recognized. Must be one of:\n\t'Red'\n\t'Green'" });
         }
+
+        [Fact]
+        public void Every_option_can_provide_a_handler_and_it_takes_precedence_over_command_handler()
```
jonsequitur, line 351: What if two different options having handlers are both present?

jonsequitur:
>  I am not sure whether it should be allowed for Arguments, currently the argument handlers are ignored.

I wasn't expecting this to be available on `Argument` and I'm not sure how to reason about it. I don't believe I've seen a CLI where I would model something this way.

src/System.CommandLine.Tests/OptionTests.cs
adamsitnik: We have two possibilities:

- stop on the first active option
- let the last one to be the effective (this is what we already do for commands)

@jonsequitur choosing the last one is the most intuitive way for me. What do you think?

adamsitnik:
> I wasn't expecting this to be available on Argument and I'm not sure how to reason about it. I don't believe I've seen a CLI where I would model something this way.

That is very true. I've applied following  changes:

- Command, Option and Directive expose `CliAction Action` property
- only Command exposes `SetAction` methods (for options and directives it's too rare to pollute the public API surface)

src/System.CommandLine.Tests/OptionTests.cs
jonsequitur: We could also make it a parse error. Of course certain options (like `-h`) ignore that and the action still runs.

tmds:
@adamsitnik in https://github.com/dotnet/command-line-api/issues/2046 you said: all handlers should return an `int`.

And here the design is:

> Command.SetAction do not need to return an int. In case somebody wants to reuturn an int, they need to implement their own CliAction type (pay for play)

Failures are often expected in command line applications.

Without the capability to return an `int`, command handler set using `SetAction` are forced to `throw` for failure.

This lead to a bad user experience.
And, throwing exceptions for expected failures is a performance anti-pattern.

PR #2099, opened 2023-03-16
Remove RegisterWithDotnetSuggest

adamsitnik:
* remove `StringBuilderPool`, it's a performance anti-pattern in case of S.CL because it was almost always used only once per process lifetime. We were paying for the pool creation but not taking any benefits from having it.
* remove `RegisterWithDotnetSuggest`, it's too expensive to start a new process for every app by default (even just once)

For now those who want to use `dotnet-suggest` they need to simply register their apps in explicit way by calling a dedicated command for that:

```cmd
dotnet-suggest register --command-path $pathToYourApp
```

This should be done as part of the app installation.

In the future we are most likely going to improve this process, but we need to have a discussion with @jonsequitur and @SteveL-MSFT about it.

cc @MichalStrehovsky who "loved" this feature and "promoted" in on Twitter multiple times


MichalStrehovsky:
> cc @MichalStrehovsky who "loved" this feature and "promoted" in on Twitter multiple times

† only [once](https://twitter.com/MStrehovsky/status/1592668141800542210). I have a notepad.exe renamed to dotnet-suggest.exe on my system PATH and see a random notepad popping up from time to time. I'll miss the charm.

adamsitnik:
@MichalStrehovsky Apologies, my memory served me wrong. Anyway the problem is gone.

BTW I wanted to spend some time looking at size of a published NativeAOT app that uses System.CommandLine. Is there a doc that explains how to track code that "roots" other code and how big it is? I am basically hoping to do something similar to memory leak analysis, but for code (identify roots, sort by size, fix).

MichalStrehovsky:
Awesome! There are some notes scattered around on how to troubleshoot size. E.g. here: https://github.com/dotnet/runtime/issues/78671 or here: https://github.com/dotnet/corert/pull/7962.

PR #2102, opened 2023-03-17
expose HelpOption and HelpAction, move help configuration to the action

adamsitnik:
We now have two ways of configuring help:

Before parsing:

```cs
command.Options.Add(new HelpOption()
{
    Action = new HelpAction()
    {
        Builder = helpBuilder
    }
});
```

Or after:

```cs
ParseResult parseResult = rootCommand.Parse("-h");

if (parseResult.Action is HelpAction helpAction)
{ 
    helpAction.Builder.CustomizeSymbol(subcommand, secondColumnText: "The custom command description");
    helpAction.Builder.CustomizeSymbol(option, secondColumnText: "The custom option description");
    helpAction.Builder.CustomizeSymbol(argument, secondColumnText: "The custom argument description");
}

parseResult.Invoke(console);
```


src/System.CommandLine/Help/HelpOption.cs
```diff
@@ -1,24 +1,19 @@
...
+    public sealed class HelpOption : Option<bool>
     {
-        internal HelpOption(string name, string[] aliases)
+        public HelpOption(string name, string[] aliases)
             : base(name, aliases, new Argument<bool>(name) { Arity = ArgumentArity.Zero })
         {
             AppliesToSelfAndChildren = true;
             Description = LocalizationResources.HelpOptionDescription();
-            Action = new HelpOptionAction();
+            Action = new HelpAction();
```
jonsequitur, line 13: Should we create this lazily? I think the ability to do so is important so we don't do up-front configuration for every action including the ones that don't end up being used for a particular invocation.

src/System.CommandLine/Help/HelpOption.cs
```diff
@@ -1,24 +1,19 @@
...
         }
 
-        internal HelpOption() : this("--help", new[] { "-h", "/h", "-?", "/?" })
+        public HelpOption() : this("--help", new[] { "-h", "/h", "-?", "/?" })
```
jonsequitur, line 16: Do you want to introduce something like `HelpOption.DefaultAliases` and use it here for discoverability, etc.?

src/System.CommandLine/Help/HelpOption.cs
adamsitnik: Very good point. I can make it possible, but it's going to require making the `CliAction` property virtual. Which should be fine?

src/System.CommandLine/Help/HelpOption.cs
adamsitnik: How would it be used by the end users? Currently to get the default aliases they can just do `new HelpOption().Aliases`

adamsitnik: IIRC Krzysztof or Jeremy started this discussion and the context was that it was not clear what aliases the `HelpOption` will use when created and added to the `Command`.

In #2107 I've made the config mutable and simply added xml docs to the `HelpOption` ctor:

https://github.com/dotnet/command-line-api/blob/2dd5b76efa1770ee230107e3d732fa9e03796e09/src/System.CommandLine/Help/HelpOption.cs#L8-L18

I am expecting that this is going to solve the problem

src/System.CommandLine/SuggestDirective.cs
```diff
@@ -4,6 +4,7 @@
 using System.CommandLine.Parsing;
 using System.Threading.Tasks;
 using System.Threading;
+using static System.Collections.Specialized.BitVector32;
```
adamsitnik, line 7: > Do these use BitVector32 for something?

Great catch, VS must have added it for me and I've missed it.

```suggestion
```

src/System.CommandLine/ParseDirective.cs
```diff
@@ -3,6 +3,7 @@
 using System.CommandLine.Parsing;
 using System.Threading;
 using System.Threading.Tasks;
+using static System.Collections.Specialized.BitVector32;
```
adamsitnik, line 6:
```suggestion
```

adamsitnik:
> If I wanted to specify the list of aliases of HelpOption (omitting "/h"), and set the maximum width

Currently it would be:

```cs
new HelpOption("--help", "-h")
{
    Action = new HelpAction()
    {
        Builder = new HelpBuilder(30)
    }
}
```

FWIW we are not done with our work on help customization. Most probably I won't have the time to work on it before we release a new preview package (beginning of April), I am afraid it will have to wait until June.

PR #2107, opened 2023-03-20
make CommandLineConfiguration mutable, remove CommandLineBuilder

### LINKED ISSUE #1907: Redesign `CommandLineConfiguration` to be mutable and favor properties over constructors
```C#
    {
        // Usability studies have shown that more junior developers get confused by the IntelliSense presentation of methods with many optional parameters.
        // Similar components use get/set properties for the optional ones, and just make the setters throw after a certain point.  For your case
```

### LINKED ISSUE #1920: Move `CommandLineBuilder` into the application framework layer
```C#
namespace System.CommandLine.Builder
{
    public class CommandLineBuilder 
    {
        public CommandLineBuilder(Command rootCommand = null);
        public Command Command { get; }
        public Parser Build()
    }
 
    // All of the things that extend CommandLineBuilder should be instance members on CommandLineBuilder.
    // If UseHelpBuilder is providing namespace-based componentization then it can make sense here, but the type name is bad for it because
    //   a) it says it's extending CommandLineBuilder, but doesn't (necessarily) and b) should have a named purpose, not just "SomethingExtensions".
    public static class CommandLineBuilderExtensions 
    {
         // Is there only one environment variable, or is it multiple?  I feel like I'd expect plurality here, but don't know why.
        public static CommandLineBuilder UseEnvironmentVariableDirective(this CommandLineBuilder builder);
 
        // What happens if this is called twice?  If it's "last writer wins", then "Set" is a better verb than "Use". If it's accumulated, then Add.  Applies to all of the Use- methods.
        public static CommandLineBuilder UseExceptionHandler(this CommandLineBuilder builder, Action<Exception,InvocationContext> onException = null, int? errorExitCode = null);
 
        // The middle overload feels like it must be doing something very different than the other two.  You either need more overloads, or to rename something, or to remove the middle one, or something.
        public static CommandLineBuilder UseHelp(this CommandLineBuilder builder, int? maxWidth = null);
        public static CommandLineBuilder UseHelp(this CommandLineBuilder builder, string[] helpAliases);
        public static CommandLineBuilder UseHelp(this CommandLineBuilder builder, Action<HelpContext> customize, int? maxWidth = null);
        
        // Generic extension methods with no constraints are prohibited, otherwise they apply to object, which is prohibited.
        public static TBuilder UseHelpBuilder<TBuilder>(this TBuilder builder, Func<BindingContext,HelpBuilder> getHelpBuilder);
 
        public static CommandLineBuilder UseLocalizationResources(this CommandLineBuilder builder, LocalizationResources validationMessages);
        public static CommandLineBuilder UseParseDirective(this CommandLineBuilder builder, int? errorExitCode = null);
        public static CommandLineBuilder UseParseErrorReporting(this CommandLineBuilder builder, int? errorExitCode = null);
        public static CommandLineBuilder UseSuggestDirective(this CommandLineBuilder builder);
        
        // Does 0 disable it?  Seems like `int TypoCorrectionLevel { get; set; }` to me; or, if 0 doesn't disable it, make it `int?`
        public static CommandLineBuilder UseTypoCorrections(this CommandLineBuilder builder, int maxLevenshteinDistance = 3);
```

### LINKED ISSUE #1922: Move appropriate `CommandLineBuilder` methods to properties on `CommandLineConfiguration`
```C#
        // Technically "cancel" is a verb, but I feel like the method is actually implying "Set" as the main verb and CancelOnProcessTermination is what the property would be called.
        // So, consider making this a property instead of a method when moving it.
        public static CommandLineBuilder CancelOnProcessTermination(this CommandLineBuilder builder);
        // These Enables also look like properties.
        public static CommandLineBuilder EnableDirectives(this CommandLineBuilder builder, bool value = True);
        public static CommandLineBuilder EnableLegacyDoubleDashBehavior(this CommandLineBuilder builder, bool value = True);
        public static CommandLineBuilder EnablePosixBundling(this CommandLineBuilder builder, bool value = True);
```

### LINKED ISSUE #1924: Consider whether `UseDefaults` is still needed
```C#
        // Wouldn't `new CommandLineBuilder()` already be using defaults? What does this method do?
        // Either needs to be renamed (Reset?) or cut.
        public static CommandLineBuilder UseDefaults(this CommandLineBuilder builder);
```

### LINKED ISSUE #1925: Consider moving `UseVersionOption` to `CommandLineConfiguration` and renaming
```C#
        // As this is basically the only time "Version" appears, I have literally no idea what this means.
        // If it's equivalent to .Add(new Option("version")), then that'd be "AddVersionOption", since the main verb is "Add".
        public static CommandLineBuilder UseVersionOption(this CommandLineBuilder builder);
        public static CommandLineBuilder UseVersionOption(this CommandLineBuilder builder, string[] aliases);
```

### LINKED ISSUE #1926: Consider moving `UseTokenReplacer` to `CommandLineConfiguration` and renaming
```csharp
        // UseTokenReplacer's parameter should obviously be "tokenReplacer", not "replaceToken"
        public static CommandLineBuilder UseTokenReplacer(this CommandLineBuilder builder, System.CommandLine.Parsing.TryReplaceToken replaceToken);
    }
}
```

### LINKED ISSUE #2034: CommandLineBuilder inconsistency?
Invoking the following program with no arguments:

```csharp
using System.CommandLine;

internal static class Program
{
    //private static Task<int> Main(string[] args) =>
    //    new CommandLineBuilder(CreateRootCommand())
    //    .Build()
    //    .InvokeAsync(args);

    private static Task<int> Main(string[] args) =>
        CreateRootCommand().InvokeAsync(args);

    private static Command CreateRootCommand()
    {
        var rootCommand = new RootCommand("Specifying a sub-command is required.");
        var fooCommand = new Command("foo");
        rootCommand.Add(fooCommand);
        return rootCommand;
    }
}
```

results in a red error message that says "Required command was not provided." followed by the program help text, and the program returns an error exit code.

However, if I replace the `Main` method above with the commented-out `Main` method and invoke the program with no arguments, then nothing appears and the program returns a successful exit code.

Is this discrepancy intentional? If so, how do I make the `CommandLineBuilder` version behave the same as the non-`CommandLineBuilder` version?

This is using System.CommandLine 2.0.0-beta4.22272.1

Thanks!

### LINKED ISSUE #2046: More design questions

### LINKED ISSUE #2057: Make CommandLineConfiguration mutable

### LINKED ISSUE #2101: New Config proposal
One of our goals is simplyfing the configuration, making it easier to discover and use.

This can be achieved by making the configuration mutable (no need for a dedicated builder) and making it an optional argument for `Command.Parse` (this has already happened).

My current idea is following:

```cs
public class CliConfiguration
{
    public CliConfiguration(Command command)
    {
        Command = command;
        Directives = new()
        {
            new SuggestDirective()
            // I am not sure whether we should enable other Directives by default
            // new EnvironmentVariablesDirective()
            // new ParseDirective()
        };
    }

    public Command Command { get; }

    public bool EnablePosixBundling { get; set; } = true;

    public bool EnableDefaultExceptionHandler {  get; set; } = true;

    public bool EnableTypoCorrections { get; set; } = false;

    public TimeSpan? ProcessTerminationTimeout { get; set; } = TimeSpan.FromSeconds(2);

    public int ParseErrorExitCode { get; set; } = 1;

    public TryReplaceToken? ResponseFileTokenReplacer { get; set; } = StringExtensions.TryReadResponseFile;

    public List<Directive> Directives { get; }
}
```

How to disable signaling and handling of process termination via a `CancellationToken`? Set `ProcessTerminationTimeout` to `null`.

How to disable default response file token replacer? Set `ResponseFileTokenReplacer` to `null`.

How to customize exception handler? Set `EnableDefaultExceptionHandler` to `false` and catch the exceptions on your own:

```cs
CliConfiguration config = new(command) { EnableDefaultExceptionHandler = false };
ParseResult parseResult = command.Parse(args, config);

try
{
    return parseResult.Invoke();
}
catch (Exception ex)
{
    // custom exception handler
}
```

The alternative for customizing the `ParseErrorExitCode` is following:

```cs
ParseResult parseResult = command.Parse(args);

if (parseResult.Errors.Any())
{
    return $customValue;
}
```

The most tricky part is how `HelpOption` and `VersionOption` should be enabled and disabled. Currently, the config builder just adds them to the provided `Command` options list:

https://github.com/dotnet/command-line-api/blob/3b5c4beb3e3d09702ba7d05b098f3594759188fd/src/System.CommandLine/Builder/CommandLineBuilderExtensions.cs#L336-L351

I don't like it, as it's a side effect. But on the other hand, I don't have an ideal alternative.

Possible solutions:

## Expect the users to add them in explicit way.

```cs
command.Options.Add(new HelpOption());
command.Options.Add(new VersionOption());
```

Advantage(s): no magic, everything is crystal clear, perf.
Disadvantage(s): help would be not enabled by default.

## Create every command with Help and Version options added by default

```cs
public Command(string name, bool addDefaultOptions = true)
{
    if (addDefaultOptions)
    {
        Options.Add(new HelpOption());
        Options.Add(new VersionOption());
    }
}
```

Advantage(s): help enabled by default.
Disadvantage(s): Hard to customize help (find, cast & customize or replace) I expect that `Version` makes sense only for the root command, it would pollute subcommands.

Perhaps it would make more sense for `RootCommand`?
Maybe the argument should not be optional, so everyone who creates a Command would need to think about it?

## Expose the options as part of Config type, add them the root Command when parsing

```cs
public class CliConfiguration
{
    public HelpOption? HelpOption { get; set; } = new ();
    
    public VersionOption? VersionOption { get; set; } = new ();
}
```

Advantage(s): quite easy to discover and configure, enabled by default.
Disadvantage(s): side effect of adding them to the root command (but so far nobody complained about it beside me?)

Do we really need `VersionOption` enabled by default?

## Add a list of default Options to Config, similarly to Directives

```cs
public class CliConfiguration
{
    public List<Option> Options { get; }
}
```

Advantage(s): quite easy to discover, consistent with Directives
Disadvantage(s): hard to configure a specific option (find HelpOption, cast it and then set the builder), side effect of adding them to the root command 

@jonsequitur @KathleenDollard @Keboo @KalleOlaviNiemitalo please provide feedback

### LINKED ISSUE #2111: CancelOnProcessTermination no longer enabled by default
Before https://github.com/dotnet/command-line-api/pull/2044, the `CancellationToken` was hooked in the default configuration.

Now, you need to explicitly call `CancelOnProcessTermination`.

@adamsitnik is this an intentional change?

### LINKED ISSUE #2117: Allow new HelpOption("--help") without aliases
Now that <https://github.com/dotnet/command-line-api/pull/2102> in System.CommandLine 2.0.0-beta4.23172.1 made System.CommandLine.Help.HelpOption public, it would be handy if I could create an instance of HelpOption by using just `new HelpOption("--help")` rather than `new HelpOption("--help", aliases: null)`. That could be implemented in any of the following ways:

 - new constructor: `public HelpOption(string name)`
 - optional parameter: `public HelpOption(string name, string[]? aliases = null)`
 - params: `public HelpOption(string name, params string[] aliases)`. That would be consistent with the Option\<T\> constructor.

I don't like using `new HelpOption("--help", null)` with an unnamed `null` argument because its purpose would not be immediately obvious to people reading the code. `new HelpOption("--help", Array.Empty<string>())` is quite clear too but more verbose than a named argument.

(System.CommandLine.Help.VersionOption is not public yet.)

adamsitnik:
fixes #2101

fixes #1907
fixes #1920
fixes #1922
fixes #1924
fixes #1925 
fixes #1926
fixes #2034
closes #2046 (the last element from the TODO list)
fixes #2057
fixes #2111
fixes #2117


src/System.CommandLine/ChildList.cs
```diff
@@ -37,7 +38,35 @@ public ChildList(Command parent)
         public void Add(T item)
         {
             item.AddParent(_parent);
+
+            if (_children.Count > 0 && _parent is RootCommand)
```
adamsitnik, line 42: `RootCommand` ctor adds `HelpOption` and `VersionOption` by default now. Users might want to customize it like that:

```cs
RootCommand rootCommand = new()
{
    new VersionOption("-v", "-version"),
};
```

In order to avoid duplicates, this method checks if a help option or version option already exists. If it does, it overrides it.

src/System.CommandLine/CommandLineConfiguration.cs
```diff
@@ -19,108 +18,76 @@ public class CommandLineConfiguration
...
+        /// Initializes a new instance of the <see cref="CommandLineConfiguration"/> class.
         /// </summary>
-        internal readonly Func<Exception, InvocationContext, int>? ExceptionHandler;
+        /// <param name="rootCommand">The root command for the parser.</param>
+        public CommandLineConfiguration(Command rootCommand)
+        {
+            RootCommand = rootCommand ?? throw new ArgumentNullException(nameof(rootCommand));
+            Directives = new()
+            {
+                new SuggestDirective()
```
adamsitnik, line 29: I was not sure which directives should be enabled by default, it's an open question.

I have not enabled all of them, as it would cause a perf hit for the users.

src/System.CommandLine.Tests/VersionOptionTests.cs
```diff
@@ -156,15 +158,17 @@ public void Version_option_is_not_added_to_subcommands()
...
-                         .UseVersionOption()
-                         .UseVersionOption()
-                         .Build();
-            configuration.Output = new StringWriter();
+            // The RootCommand ctors adds VersionOption by default.
+            // Users might not know if or just want to customize it.
+            RootCommand commandWithDuplicates = new()
+            {
+                new VersionOption(),
+                new VersionOption(),
```
jonsequitur, line 166: Shouldn't this be an error because of the symbol identifier collision?

I think the expectation with the builder is that it's idempotent, but I don't think I have the same expectation with `Command.Add` since we're enforcing name and alias uniqueness.

src/System.CommandLine/ChildList.cs
jonsequitur: I think it might be less surprising to just ask people to remove the other instance, especially given that the `HelpAction` is  also customizable and is probably the more common path to customizing help.

adamsitnik: @jonsequitur you are right, I've removed this logic and updated the tests

src/System.CommandLine.Tests/VersionOptionTests.cs
```diff
@@ -155,33 +155,17 @@ public void Version_option_is_not_added_to_subcommands()
...
         {
-            RootCommand rootCommand = new()
+            RootCommand rootCommand = new();
+
+            for (int i = 0; i < rootCommand.Options.Count; i++)
             {
-                new VersionOption("-v", "-version"),
-            };
+                if (rootCommand.Options[i] is VersionOption)
+                    rootCommand.Options[i] = new VersionOption("-v", "-version");
```
jonsequitur, line 166: This is a general comment and doesn't need to block merging.

The API through beta4 treated the arguments to `Option<T>` such that  `new VersionOption("-v", "-version")` would indicate `-version` is the description.

The beta5 API understands this as setting `Name` to `-v` and using `-version` as an alias. In general usage though, `-version` would be the name (being longer) and `-v` would be the alias.

Two observations:

* I've found this to be a gotcha in upgrading my code.
* This usage in tests, while perfectly valid, might mislead about the way in which the API is typically used.

jonsequitur (approved):

Issue #2112, opened 2023-03-21
Uncaught AggregateException when CancelOnProcessTermination


tmds:
The `ProcessTerminationHandler` is waiting with a timeout for the handler task to complete here:

https://github.com/dotnet/command-line-api/blob/02a776b901846b96cc2a7442b2032102944cafa9/src/System.CommandLine/Invocation/ProcessTerminationHandler.cs#L85

It's currently not handling the case where the handler completed with an exception.
When that happens, the `Wait` method throws `AggregateException`.
That exception should be caught and ignored.

cc @adamsitnik

adamsitnik:
@tmds great catch, would you like to send a PR with a fix?

PR #2113, opened 2023-03-22
ProcessTerminationHandler: handle case where handler completes with an exception.

tmds:
Fixes https://github.com/dotnet/command-line-api/issues/2112.

@adamsitnik ptal.


adamsitnik (approved):
LGTM, thank you @tmds!

PR #2124, opened 2023-03-29
move IValueDescriptor to System.CommandLine.NamingConventionBinder

### LINKED ISSUE #1917: Consider renaming and/or moving types out of the `Binding` namespace.

### LINKED ISSUE #1918: Consider the binding space, including whether `BinderBase<T>`, `IValueSource`, and `IValueDescriptor` are needed in the public API for usage beyond `SetHandler`
```C#
namespace System.CommandLine.Binding
{ 
    // This is the only hit of BinderBase in this proposal.  Is it needed?
    // The "Base" suffix is recommended against unless it's a very niche carve-out from a class that uses the better name.
    // Basically, if more than 30 people in the whole would would ever need to write it in their code you need a better name.
    public abstract class BinderBase<T> : IValueDescriptor<T>, IValueDescriptor, IValueSource 
    {
        protected T GetBoundValue(BindingContext bindingContext);
    }

    // In general, interfaces should only define capabilities, not actual structure.
    // If at all possible, this should be turned into an abstract class.
    // Note that interfaces can basically never change (we have never successfully added a DIM as a way to expan an interface).
    // So with this interface there's no proven way to add, e.g. T GetDefaultValue<T>(Func<T> defaultValue) in a later version
    // in a way that avoids boxing values into object.
    public interface IValueDescriptor 
    {
        bool HasDefaultValue { get; }
        string ValueName { get; }
        Type ValueType { get; }
        object GetDefaultValue();
    }
 
    // Interfaces with no members are effectively prohibited.
    public interface IValueDescriptor<out T> : IValueDescriptor 
    {
    }
 
    // This doesn't look like a capability, so a base class is very strongly preferred over an interface.
    // (We really don't like interfaces)
    public interface IValueSource 
    {
        // We really don't like ref.  Is this actually C#'s `out`?
        bool TryGetValue(IValueDescriptor valueDescriptor, BindingContext bindingContext, ref object& boundValue);
    }
```

adamsitnik:
It was way easier than I expected, mostly because at the end of the day the crucial part of the code was casting `IValueDescriptor` to `Argument` and `Symbol`:

https://github.com/dotnet/command-line-api/blob/d18f598068fad37791e9f401a05d5f6a0591fa0e/src/System.CommandLine.NamingConventionBinder/SpecificSymbolValueSource.cs#L21-L40

fixes #1917
fixes #1918


jonsequitur (approved):

src/System.CommandLine/Option.cs
```diff
@@ -12,7 +12,7 @@ namespace System.CommandLine
     /// <summary>
     /// A symbol defining a named parameter and a value for that parameter. 
     /// </summary>
-    public abstract class Option : Symbol, IValueDescriptor
+    public abstract class Option : Symbol
```
am11, line 15: @adamsitnik, do you have any recommendation what should be used here: https://github.com/dotnet/runtime/blob/f8bd9d29be973e0a9b82b87b9e50a0d3a77d1697/src/coreclr/tools/Common/CommandLineHelpers.cs#L187-L189? It (`--make-repro`) is a special facility where we create a repro package with command+arguments+artifacts (as zip). In there, we are checking if `Option`'s value is default (set by default factory). Note this is not a hot path so using reflection is an option as a last resort. (to get by the compile error for now, I used reinterpretation style cast: `(IValueDescriptor)(object)option` 😅)

ps - I am in the middle of grand upgrade in runtime and jitutils repos. There is no rush, I'm gradually rebasing and picking up latest daily versions as I go further in my branch.

am11: Found a solution:

```diff
-                    IValueDescriptor descriptor = option;
                    object val = res.CommandResult.GetValue(option);
-                    if (val is not null && !(descriptor.HasDefaultValue && descriptor.GetDefaultValue().Equals(val)))
+                    if (val is not null && !(option.DefaultValueFactory is not null && option.DefaultValueFactory.Invoke().Equals(val)))
```
It is a bit strange that `CliArgument<T>` provides `HasDefaultValue and `GetDefaultValue()` but `CliOption<T>` does not. 👀

am11: https://github.com/dotnet/runtime/pull/84229

PR #2127, opened 2023-03-29
Remove dependencies

### LINKED ISSUE #1794: System.CommandLine defaults have large dependency closure
## Repro

`dotnet publish -c Release -r win-x64 /p:PublishTrimmed=true`

```csharp
using System.CommandLine;

internal class Program
{
    private static int Main(string[] args)
    {
        var fileOption = new Option<FileInfo?>(
            name: "--file",
            description: "The file to read and display on the console.");

        var rootCommand = new RootCommand("Sample app for System.CommandLine");
        rootCommand.AddOption(fileOption);

        rootCommand.SetHandler(
            (FileInfo file) => ReadFile(file),
            fileOption);
    }

    private static void ReadFile(FileInfo file)
        => File.ReadLines(file.FullName).ToList()
            .ForEach(line => Console.WriteLine(line));
}
```

## Actual result:

The trimmed application has referenced to many types and binaries that should not be needed:
```
System.Private.Uri.dll
System.Net.Primitives.dll
System.ObjectModel.dll
System.ComponentModel.dll
System.ComponentModel.Primitives.dll
System.ComponentModel.TypeConverter.dll
System.Collections.NonGeneric.dll
System.Runtime.Serialization.Formatters.dll
System.Diagnostics.Process.dll
...
```

## Expected result:

The trimmed application should only reference types and binaries that are actually needed.

adamsitnik:
At the cost of requiring the users to provide explicit custom parser for `Uri`, `IPAddress` and `IPEndPoint` and a removal of non-generic collections support I was able to remove following dependencies:

```log
- System.Private.Uri.dll
- System.Net.Primitives.dll
- System.Collections.dll
- System.Collections.Concurrent.dll
- System.Runtime.Serialization.Formatters.dll
```

The most tricky part was removing dependency to `System.Collections.dll`, which was caused by the fact that we were using `Stack<T>` (it's generic, but not defined in `System.Collections.Generic.dll` or CoreLib...)

The only dependency that an app that uses S.CL has and a Hello World app does not, is `System.Linq` (and S.CL of course).

For following app:

<details>

```cs
using System;
using System.CommandLine;

public class Program
{
    static void Run(bool boolean, string text)
    {
        Console.WriteLine($"Bool option: {text}");
        Console.WriteLine($"String option: {boolean}");
    }

    private static int Main(string[] args)
    {
        Option<bool> boolOption = new Option<bool>("--bool", "-b") { Description = "Bool option" };
        Option<string> stringOption = new Option<string>("--string", "-s") { Description = "String option" };

        RootCommand command = new ()
        {
            boolOption,
            stringOption,
        };

        command.SetAction(ctx => Run(ctx.GetValue(boolOption), ctx.GetValue(stringOption)));

        return new CommandLineConfiguration(command).Invoke(args);
    }
}
```

</details>

And 

```cmd
dotnet publish -c Release -r win-x64 /p:PublishTrimmed=true
```

The number of files dropped from 34 to 29, size from 17.5 MB to 17.3.
For NativeAOT app the size dropped from 4.65 MB to 4.5 MB.

fixes #1794

Useful link for repeating the process in the future: https://github.com/dotnet/runtime/issues/78671#issuecomment-1370315695


adamsitnik:
Updated perf numbers for startup scenario benchmarks (https://github.com/adamsitnik/commandline-perf/tree/update):

|                  Method |           Args |      Mean |    Error |   StdDev |    Median | Ratio | RatioSD |
|------------------------ |--------------- |----------:|---------:|---------:|----------:|------:|--------:|
|                   Empty |                |  67.58 ms | 0.699 ms | 2.061 ms |  67.06 ms |  0.79 |    0.04 |
|                EmptyAOT |                |  13.84 ms | 0.461 ms | 1.358 ms |  13.53 ms |  0.16 |    0.02 |
|               Corefxlab |                |  87.27 ms | 1.174 ms | 3.461 ms |  86.53 ms |  1.02 |    0.05 |
|   SystemCommandLine2021 |                | 105.36 ms | 1.328 ms | 3.916 ms | 104.77 ms |  1.23 |    0.06 |
|   SystemCommandLine2022 |                |  88.50 ms | 0.686 ms | 2.022 ms |  88.39 ms |  1.03 |    0.04 |
|    SystemCommandLineNow |                |  85.92 ms | 1.089 ms | 3.210 ms |  85.03 ms |  1.00 |    0.00 |
| SystemCommandLineNowR2R |                |  78.78 ms | 0.480 ms | 1.414 ms |  78.61 ms |  0.92 |    0.04 |
| SystemCommandLineNowAOT |                |  14.80 ms | 0.163 ms | 0.480 ms |  14.78 ms |  0.17 |    0.01 |
|                         |                |           |          |          |           |       |         |
|                   Empty | --bool -s test |  64.90 ms | 0.341 ms | 1.005 ms |  64.85 ms |  0.74 |    0.01 |
|                EmptyAOT | --bool -s test |  13.15 ms | 0.129 ms | 0.381 ms |  13.11 ms |  0.15 |    0.00 |
|               Corefxlab | --bool -s test |  85.26 ms | 0.630 ms | 1.858 ms |  84.92 ms |  0.97 |    0.02 |
|   SystemCommandLine2021 | --bool -s test | 105.72 ms | 0.792 ms | 2.336 ms | 104.96 ms |  1.20 |    0.03 |
|   SystemCommandLine2022 | --bool -s test |  90.10 ms | 0.542 ms | 1.599 ms |  89.81 ms |  1.02 |    0.02 |
|    SystemCommandLineNow | --bool -s test |  87.96 ms | 0.399 ms | 1.176 ms |  87.72 ms |  1.00 |    0.00 |
| SystemCommandLineNowR2R | --bool -s test |  79.35 ms | 0.412 ms | 1.214 ms |  79.25 ms |  0.90 |    0.02 |
| SystemCommandLineNowAOT | --bool -s test |  14.87 ms | 0.131 ms | 0.387 ms |  14.82 ms |  0.17 |    0.00 |

jonsequitur (approved):

jonsequitur (approved):

src/System.CommandLine/Argument{T}.cs
```diff
@@ -160,5 +162,36 @@ public void AcceptLegalFileNamesOnly()
                 }
             });
         }
+
+        [UnconditionalSuppressMessage("ReflectionAnalysis", "IL3050", Justification = "https://github.com/dotnet/command-line-api/issues/1638")]
+        [UnconditionalSuppressMessage("ReflectionAnalysis", "IL2091", Justification = "https://github.com/dotnet/command-line-api/issues/1638")]
```
eerhardt, line 167: I don't understand why the referenced issue is justification for suppressing these warnings. It might be better if we wrote it out here.

I believe you could fix `IL2091` by annotating the `T` with `[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors)]`. See https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/fixing-warnings#dynamicallyaccessedmembers

src/System.CommandLine/Argument{T}.cs
```diff
@@ -160,5 +162,36 @@ public void AcceptLegalFileNamesOnly()
...
+        internal static T? CreateDefaultValue()
+        {
+            if (default(T) is null && typeof(T) != typeof(string))
+            {
+                if (typeof(T).IsArray)
+                {
+                    return (T?)(object)Array.CreateInstance(typeof(T).GetElementType()!, 0);
```
eerhardt, line 174: We should be able to use https://github.com/dotnet/runtime/issues/76478 once it is implemented in .NET 8. That will allow us to do this without a warning.

src/System.CommandLine/Argument{T}.cs
```diff
@@ -160,5 +162,36 @@ public void AcceptLegalFileNamesOnly()
...
+                }
+                else if (typeof(T).IsGenericType)
+                {
+                    var genericTypeDefinition = typeof(T).GetGenericTypeDefinition();
+
+                    if (genericTypeDefinition == typeof(IEnumerable<>) ||
+                        genericTypeDefinition == typeof(IList<>) ||
+                        genericTypeDefinition == typeof(ICollection<>))
+                    {
+                        return (T?)(object)Array.CreateInstance(typeof(T).GenericTypeArguments[0], 0);
```
eerhardt, line 184: This isn't guaranteed to work in a NativeAOT app. Just because the app uses a `IList<double>`, it doesn't mean that `double[]`'s code is going to get preserved/generated.

PR #2132, opened 2023-03-30
rename symbol types with `Cli` prefix

### LINKED ISSUE #1892: Rename `Argument`, `Option`, and `Command` to disambiguate from other single-word types
```C#
    // Consider adding a disambiguator to this type name, so it doesn't feel like "ArgumentException" is related to this type.
    public abstract class Argument : Symbol, IValueDescriptor, ICompletionSource 
```

### LINKED ISSUE #1916: Consider renaming `Symbol` to be less likely to collide and to be consistent with the renaming of its derived types.
```C#
    // There aren't any "Symbol" classes in the shared runtime, but this type name existing feels like it'll be a big conflict for
    // Roslyn, or any other sort of compiler/lexer/etc.
    // Is there a two-or-more-particle name that would be almost as good?
    public abstract class Symbol : ICompletionSource 
    {
```

Related: #1892

### LINKED ISSUE #2131: Update System.CommandLine symbols API per API review notes
The following API changes were agreed upon:

- [x] CliSymbol.IsHidden => Hidden (drop the "Is")
- [x] CliOption.AppliesToSelfAndChildren => Recursive
- [x] CliOption.IsRequred = Required
- [x] CliOption we added ValueType for symmetry with CliArgument
- [x] DiagramDirective: errorExitCode was renamed to parseErrorReturnValue, and made a get/set property instead of a ctor parameter.
- [x] We discussed the "Cli" prefix. We agree there does need to be some particle (because "Symbol", "Option", "Command", etc are too general and will probably have app collisions). No one thought it was great, but no one had a better name.

https://github.com/dotnet/runtime/issues/68578#issuecomment-1490763062

jonsequitur:
This also renames a number of properties and renames or moves a few other files so that file names match type names.

It fixes #2131. 
fixes #1892
fixes #1916


src/System.CommandLine/ParseDiagramDirective.cs
```diff
@@ -6,25 +6,27 @@ namespace System.CommandLine
     /// Enables the use of the <c>[parse]</c> directive, which when specified on the command line will short 
     /// circuit normal command handling and display a diagram explaining the parse result for the command line input.
     /// </summary>
-    public sealed class ParseDirective : Directive
+    public sealed class ParseDiagramDirective : CliDirective
```
adamsitnik, line 9: The approved name is `DiagramDirective`. I am going to send a separate PR with a rename and changing `[parse]` to `[diagram]` just to start a discussion

src/System.CommandLine/Parsing/OptionResult.cs
```diff
@@ -28,13 +28,13 @@ public sealed class OptionResult : SymbolResult
...
         /// </summary>
-        public Option Option { get; }
+        public CliOption Option { get; }
 
         /// <summary>
         /// Indicates whether the result was created implicitly and not due to the option being specified on the command line.
         /// </summary>
         /// <remarks>Implicit results commonly result from options having a default value.</remarks>
-        public bool IsImplicit => IdentifierToken is null || IdentifierToken.IsImplicit;
+        public bool Implicit => IdentifierToken is null || IdentifierToken.Implicit;
```
adamsitnik, line 37: thank you for renaming all the properties, including even these which were not officially approved yet. It's great for consistency  and should help in the next API review round

jonsequitur:
Thanks for linking the other issues this addressed. @adamsitnik!

PR #2134, opened 2023-03-31
dotnet-suggest-shim.bash: also suggest completions past the first word.

tmds:
@jonsequitur @adamsitnik ptal.


jonsequitur (approved):

jonsequitur:
Thanks, @tmds!

PR #2141, opened 2023-04-01
Rename files to match their class name changes.

tmds:
For the renames to CliConfiguration and CliParser.

@adamsitnik @jonsequitur ptal.


adamsitnik (approved):
Thank you @tmds!

PR #2160, opened 2023-04-11
SuggestionStore::GetCompletions to check exit code of invoked applica…

### LINKED ISSUE #2137: GetCompletions should check exit code of invoked application
This code should probably check the exit code and **not** consider standard output to be completions when it is non-zero:

https://github.com/dotnet/command-line-api/blob/403234a4c4247151d09073bc07ccc13530ce62c6/src/System.CommandLine.Suggest/SuggestionStore.cs#L47-L50

cc @jonsequitur @adamsitnik

baradgur:
This fixes #2137.

### DIFF
--- a/src/System.CommandLine.Suggest/SuggestionStore.cs
+++ b/src/System.CommandLine.Suggest/SuggestionStore.cs
@@ -44,7 +44,7 @@ namespace System.CommandLine.Suggest
 
                     Task<string> readToEndTask = process.StandardOutput.ReadToEndAsync();
 
-                    if (readToEndTask.Wait(timeout))
+                    if (readToEndTask.Wait(timeout) && process.HasExited && process.ExitCode == 0)
                     {
                         result = readToEndTask.Result;
                     }


baradgur:
@dotnet-policy-service agree

jonsequitur (approved):

src/System.CommandLine.Suggest/SuggestionStore.cs
```diff
@@ -44,7 +44,7 @@ public string GetCompletions(string exeFileName, string suggestionTargetArgument
 
                     Task<string> readToEndTask = process.StandardOutput.ReadToEndAsync();
 
-                    if (readToEndTask.Wait(timeout))
+                    if (readToEndTask.Wait(timeout) && process.HasExited && process.ExitCode == 0)
```
jonsequitur, line 47: The `else` block might need some additional checks as well. It doesn't appear that we have code coverage over both of these code branches.

baradgur:
I saw the issue yesterday and that is why th PR is not marked as ready for review yet. Sorry for confusion, I will fix and test this today. Thanks for the feed back anyway:)

src/System.CommandLine.Suggest.Tests/TestsWithTestApps.cs
```diff
@@ -0,0 +1,74 @@
+using System.IO;
+using System.Linq;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace System.CommandLine.Suggest.Tests;
+
+[Collection("TestsWithTestApps")]
```
baradgur, line 8: Note: this attribute is inherited, so both SuggestionStoreTests and DotnetSuggestEndToEndTests will run in the same collection, meaning - not in parallel. This allows the gracefull cleanup of TestsApp files.

src/System.CommandLine.Suggest.Tests/WaitAndFailTestApp/Program.cs
baradgur, line 1: Not sure whether that solution with the faulted console app was the right way, honestly.

src/System.CommandLine.Suggest/SuggestionStore.cs
```diff
@@ -41,10 +42,20 @@ public string GetCompletions(string exeFileName, string suggestionTargetArgument
...
                     process.Start();
-
+                    process.BeginOutputReadLine();
+                    
+                    var stringBuilder = new StringBuilder();
+                    process.OutputDataReceived += (sender, eventArgs) =>
+                    {
+                        if (eventArgs.Data != null)
+                        {
+                            stringBuilder.AppendLine(eventArgs.Data);
```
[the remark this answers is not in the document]
baradgur: Thank you, might be the mistake is here! I hope it is my last. Otherwise I am completely at loss right now on why the test keeps faling.
Regarding Append/AppendLine: seems not the case here - I have tested it, the eventArgs.Data comes without any line terminator.

baradgur:
Can I do a diagnostic commit (e.g. add some messages) and revert it later? Is this allowed? The test is failing and I unfortunately have no means to identify the reason.

baradgur:
@jonsequitur Thanks, I saw this one :) My trouble is  - I cannot reproduce the issue on my side no matter what I try. If you can offer some thoughts about this, I would be grateful. 
My current plan is to configure gihub actions on a separate branch (or even repo, not sude wheter github allows actions on a fork) to run tests on the machine with the same configuration as the build machine (e.g [ubuntu one](https://github.com/actions/runner-images/releases/tag/ubuntu20%2F20230517.1)) and add some robust logging to diagnose the issue.

PR #2164, opened 2023-04-12
Add `OptionResult.IdentifierTokenCount` property

### LINKED ISSUE #669: Multiple flag options

jonsequitur:
This introduces a property, `OptionResult.IdentifierTokenCount`, that can be used to provide a solution to #669.

This property exposes the number of times an identifier token was encountered for a given option. This value is incremented once per token regardless of whether the option name or one of its aliases was parsed.

It also fixes an `ArgumentException` that would occur when `-v -v -v` was parsed.


adamsitnik (approved):
LGTM!

CEbbinghaus (approved):
This looks great. Exactly what I was looking for. Hopefully this gets merged ASAP

PR #2186, opened 2023-05-08
dotnet-suggest script: use appropriate line-endings.

tmds:
Use Windows line-endings when printing the PowerShell script, and Unix line-endings when printing the bash and zsh scripts.

@jonsequitur @adamsitnik ptal.


adamsitnik (approved):
LGTM, thank you for providing the fix @tmds!

adamsitnik (approved):
LGTM, thank you for providing the fix @tmds!

Issue #2143, opened 2023-04-03
bash completion script: suggestions should be filtered out that do not start with the word to complete


tmds:
`dotnet-suggest` returns completions that do not start with the word to complete.

For example `p<TAB>`, can offer suggestions like:
```
pear
--project
--expose
```

For the bash completion script, in this case, we should only return suggestions that start with `p`: `pear`.

I'm not sure why `dotnet-suggest` is proposing options that do not start with `WordToComplete`? I assume it is for the UX with another shell?

If it is not desired to filter out these suggestions in `dotnet-suggest`, we can filter them out in the bash script.

jonsequitur:
Completions use substring matching, which is intentional.

tmds:
> which is intentional.

This is for PowerShell, I assume?

We need to filter them out for bash.

I will take a stab at improving the bash competition script next week.

jonsequitur:
This isn't a shell-specific.

System.CommandLine uses substring matching, which has become more common than not in the IDE and editor world. Multi-word matching is also starting to become more common. These approaches help discoverability and ergonomics.

When you say they need to be filtered out for bash, do you think this falls into the category of preference or functional requirement?

tmds:
I looked into this and there is a bigger issue.

To know what word is being completed, the script passes `COMP_POINT`. This is the index in `COMP_LINE`.

When the directive makes a suggestion for the completion, it doesn't see `COMP_LINE` but it sees the separate arguments that got parsed.

That means the index that is passed may not reflect the argument that is being completed.

For example:
```
dotnet-shift deploy --con<TAB>
```
completes using:
```
dotnet-shift --position 25 --  dotnet-shift deploy --con
```
successfully to:
```
dotnet-shift deploy --context
```

However
```
dotnet-shift              deploy --con<TAB>
```
does not.

It tries to complete using:
```
dotnet-shift --position 38 --  dotnet-shift deploy --con
```
Note the shift in position due to the additional spaces on the command.

Because the position can not be matched to the `--con` argument being completed, all options are suggested rather than completing the option name.

tmds:
You can ignore my previous comment. It was running against a script I had modified locally.

The whole line gets passed down as a single argument. This surprises me. It means S.CL can parse full lines into their arguments with the same semantics as bash.

jonsequitur:
> The whole line gets passed down as a single argument. 

[...]

PR #2187, opened 2023-05-08
dotnet-suggest-shim.bash: filter on word to complete.

tmds:
Fixes https://github.com/dotnet/command-line-api/issues/2143.

@jonsequitur ptal.

### DIFF
--- a/src/System.CommandLine.Suggest/dotnet-suggest-shim.bash
+++ b/src/System.CommandLine.Suggest/dotnet-suggest-shim.bash
@@ -1,30 +1,30 @@
 # dotnet suggest shell complete script start
 _dotnet_bash_complete()
 {
-    local fullpath=`type -p ${COMP_WORDS[0]}`
-    local escaped_comp_line=$(echo "$COMP_LINE" | sed s/\"/'\\\"'/g)
-    local completions=`dotnet-suggest get --executable "${fullpath}" --position ${COMP_POINT} -- "${escaped_comp_line}"`
+    local fullpath=$(type -p "${COMP_WORDS[0]}")
+    local completions=$(dotnet-suggest get --executable "${fullpath}" --position "${COMP_POINT}" -- "${COMP_LINE}")
 
+    # filter suggestions by word to complete.
+    local word="${COMP_WORDS[COMP_CWORD]}"
     local IFS=$'\n'
-    local suggestions=($(compgen -W "$completions"))
+    local suggestions=($(compgen -W "$completions" -- "$word"))
 
-    if [ "${#suggestions[@]}" == "1" ]; then
-        local number="${suggestions[0]/%\ */}"
-        COMPREPLY=("$number")
-    else
-        for i in "${!suggestions[@]}"; do
-            suggestions[$i]="$(printf '%*s' "-$COLUMNS"  "${suggestions[$i]}")"
-        done
+    # format suggestions as shell input.
+    for i in "${!suggestions[@]}"; do
+        suggestions[i]="$(printf '%q' "${suggestions[$i]}")"
+    done
 
-        COMPREPLY=("${suggestions[@]}")
-    fi
+    COMPREPLY=("${suggestions[@]}")
 }
 
 _dotnet_bash_register_complete()
 {
-    local IFS=$'\n'
-    complete -F _dotnet_bash_complete `dotnet-suggest list`
+    if command -v dotnet-suggest &>/dev/null; then
+        local IFS=$'\n'
+        complete -F _dotnet_bash_complete $(dotnet-suggest list)
+    fi
 }
+
 _dotnet_bash_register_complete
-export DOTNET_SUGGEST_SCRIPT_VERSION="1.0.2"
+export DOTNET_SUGGEST_SCRIPT_VERSION="1.0.3"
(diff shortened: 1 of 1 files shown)


src/System.CommandLine.Suggest/dotnet-suggest-shim.bash
```diff
@@ -1,30 +1,30 @@
 # dotnet suggest shell complete script start
 _dotnet_bash_complete()
 {
-    local fullpath=`type -p ${COMP_WORDS[0]}`
-    local escaped_comp_line=$(echo "$COMP_LINE" | sed s/\"/'\\\"'/g)
-    local completions=`dotnet-suggest get --executable "${fullpath}" --position ${COMP_POINT} -- "${escaped_comp_line}"`
+    local fullpath=$(type -p "${COMP_WORDS[0]}")
```
tmds, line 4: I'm sure what the `sed s/\"/'\\\"'/g` is about so I left it out.

src/System.CommandLine.Suggest/dotnet-suggest-shim.bash
```diff
@@ -1,30 +1,30 @@
...
+    local completions=$(dotnet-suggest get --executable "${fullpath}" --position "${COMP_POINT}" -- "${COMP_LINE}")
 
+    # filter suggestions by word to complete.
+    local word="${COMP_WORDS[COMP_CWORD]}"
     local IFS=$'\n'
-    local suggestions=($(compgen -W "$completions"))
+    local suggestions=($(compgen -W "$completions" -- "$word"))
```
tmds, line 10: This is the fix for https://github.com/dotnet/command-line-api/issues/2143.

src/System.CommandLine.Suggest/dotnet-suggest-shim.bash
```diff
@@ -1,30 +1,30 @@
...
 
-    if [ "${#suggestions[@]}" == "1" ]; then
-        local number="${suggestions[0]/%\ */}"
-        COMPREPLY=("$number")
-    else
-        for i in "${!suggestions[@]}"; do
-            suggestions[$i]="$(printf '%*s' "-$COLUMNS"  "${suggestions[$i]}")"
-        done
+    # format suggestions as shell input.
```
tmds, line 12: Unfortunately, there are no comments trying to explain what the script was trying to do.
There seemed to be two things going on:

- `${suggestions[0]/%\ */}` some form of eating characters after whitespace when there is a single suggestion.

To deal with whitespace, I've added the `%q` formatting. For example spaces in a completion will be escaped as `\ `.


- `"$(printf '%*s' "-$COLUMNS" "${suggestions[$i]}")"` printing at line width to make each suggestion appear on a separate line.

This is an opinionated formatting, I left it out in favor of the default formatting.

src/System.CommandLine.Suggest/dotnet-suggest-shim.bash
```diff
@@ -1,30 +1,30 @@
...
-        COMPREPLY=("${suggestions[@]}")
-    fi
+    COMPREPLY=("${suggestions[@]}")
 }
 
 _dotnet_bash_register_complete()
 {
-    local IFS=$'\n'
-    complete -F _dotnet_bash_complete `dotnet-suggest list`
+    if command -v dotnet-suggest &>/dev/null; then
```
tmds, line 22: This is to avoid printing errors if the tool was uninstalled and the script is still sourced.

baronfel:
Raising a flag for myself to cross-check these changes against the .NET CLI suggestion script once this merges :)

tmds:
> Raising a flag for myself to cross-check these changes against the .NET CLI suggestion script once this merges :)

I didn't know/forgot the sdk came with a completion script too.
(Longer term) I think it would be good to merge these. I've created an issue: https://github.com/dotnet/sdk/issues/32354.

tmds:
@adamsitnik ptal.

adamsitnik:
> @adamsitnik ptal.

I am a scripting n00b, I would prefer @jonsequitur to review the PR (or somebody else who is confident in doing that)

src/System.CommandLine.Suggest/dotnet-suggest-shim.bash
baronfel: This portion makes complete sense to me:

* COMP_WORDS - array of parsed words
* COMP_CWORD - array index of the current cursor position
* ${COMP_WORDS[COMP_CWORD]} - current whole word
* `compgen -W "$completions" -- "$word"` - -W is the full list of potential completions, and specifying a `$word` argument tells compgen which words to filter on

src/System.CommandLine.Suggest/dotnet-suggest-shim.bash
baronfel: Agree with both of these changes - `%q` is explicitly for formatting a term as  shell-safe input, and leaving the input as the default allows users to customize their bash completion displays consistently across tools via the well-known `completion-` prefixed shell options.

baronfel:
@jonsequitur @adamsitnik I am completely happy with these changes - I've commented a bit about them in this PR if you'd like to review, but if it counts for anything I approve them.

baronfel (approved):

PR #2211, opened 2023-06-06
Respect custom help builder during parse error reporting

adamsitnik:
It may be the last bug for https://github.com/dotnet/sdk/pull/29131 (at leas I hope so)


adamsitnik:
> This conflicts with my changes in #2205, where I've removed the `EnableParseErrorReporting` property.

I am aware of that. This fixes solves the last bug that stops us from updating the SDK:

https://dev.azure.com/dnceng-public/public/_build/results?buildId=297080&view=ms.vss-test-web.build-test-results-tab

image

it would be really great to merge to be able to merge https://github.com/dotnet/sdk/pull/29131

jonsequitur (approved):

PR #2212, opened 2023-06-07
Fix #2210

### LINKED ISSUE #2210: Option with nullable type causes InvalidOperationException when parsing invalid value
## Problem
If I use a nullable type for an `Option<T>` (which is useful for determining whether the option was specified at all), and an invalid value is provided to that Option on the command line, the application will crash with an InvalidOperationException. It does not crash if the Option type is not nullable.

I will freely admit that using a nullable type for an Option may not be the "correct" approach; it _seemed_ like the most expedient way to identify whether an option was specified by the user. 

## Example for Reproducing
For example, invoking the following application with arguments "--test ouch" will crash:
```csharp
using System.CommandLine;

RootCommand rootCommand = new("example-app");
Option<bool?> testOption = new("--test");
rootCommand.AddOption(testOption);
rootCommand.SetHandler(rootHandler, testOption);
CommandLineBuilder builder = new(rootCommand);
return builder.Command.InvokeAsync(args).Result;

async Task<int> rootHandler(bool? test)
{
	Console.WriteLine("In handler");
	return 0;
}
```

```
Unhandled exception: System.InvalidOperationException: Cannot parse argument 'ouch' for option '--test' as expected type 'System.Boolean'.
   at System.CommandLine.Binding.ArgumentConverter.GetValueOrDefault[T](ArgumentConversionResult result)
   at System.CommandLine.Parsing.SymbolResult.GetValueForOption[T](Option`1 option)
...
```

But if I change it to `Option<bool>` instead of `Option<bool?>` (and the corresponding change to rootHandler), then there's no crash:

```
'ouch' was not matched. Did you mean one of the following?
-h
Unrecognized command or argument 'ouch'.

Description:
  example-app
...
```

## Proposed Solutions

1. Support nullable types for Option so they are handled more like "regular" types from an end user standpoint.
2. Block the use of nullable types for Option (provided there is a different/better way to determine whether an option was specified).

jonsequitur:
This fixes #2210. The issue was that `bool` parsing had a special case in place so that it doesn't consume the following token if that token can't be parsed as a bool. This lets the unparseable token by consumed by a different symbol, which is typically the intent.

The fix includes `bool?` in that special case.

### DIFF
--- a/src/System.CommandLine/ParseResult.cs
+++ b/src/System.CommandLine/ParseResult.cs
@@ -135,6 +135,7 @@ namespace System.CommandLine
         public T? GetValue<T>(string name)
         {
             var command = CommandResult.Command;
+
             if (_namedResults is null)
             {
                 // A null value means that given name exists, but was not parsed
@@ -180,12 +181,12 @@ namespace System.CommandLine
                 }
             }
 
-            static T? Convert(ArgumentConversionResult validatedResult)
+            static T Convert(ArgumentConversionResult validatedResult)
             {
                 var convertedResult = validatedResult.ConvertIfNeeded(typeof(T));
 
-                if (validatedResult.Result == ArgumentConversionResultType.Successful
-                    && convertedResult.Result == ArgumentConversionResultType.NoArgument)
+                if (validatedResult is { Result: ArgumentConversionResultType.Successful }
+                    && convertedResult is { Result: ArgumentConversionResultType.NoArgument })
                 {
                     // invalid cast has been detected, InvalidCastException will be thrown
                     return (T)validatedResult.Value!;
--- a/src/System.CommandLine/Parsing/ParseOperation.cs
+++ b/src/System.CommandLine/Parsing/ParseOperation.cs
@@ -246,9 +246,10 @@ namespace System.CommandLine.Parsing
                         break;
                     }
                 }
-                else if (argument.ValueType == typeof(bool) && 
+                else if ((argument.ValueType == typeof(bool) || argument.ValueType == typeof(bool?))  && 
                          !bool.TryParse(CurrentToken.Value, out _))
                 {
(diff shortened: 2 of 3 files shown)


Keboo (approved):

src/System.CommandLine/ParseResult.cs
```diff
@@ -180,12 +181,12 @@ public IReadOnlyList<string> UnmatchedTokens
...
             }
 
-            static T? Convert(ArgumentConversionResult validatedResult)
+            static T Convert(ArgumentConversionResult validatedResult)
             {
                 var convertedResult = validatedResult.ConvertIfNeeded(typeof(T));
 
-                if (validatedResult.Result == ArgumentConversionResultType.Successful
-                    && convertedResult.Result == ArgumentConversionResultType.NoArgument)
+                if (validatedResult is { Result: ArgumentConversionResultType.Successful }
```
adamsitnik, line 188: do you find this syntax more readable?

adamsitnik (commented):
Thank you for fixing the bug!

src/System.CommandLine/ParseResult.cs
jonsequitur: I prefer it as a best practice because the behavior of `==` can be changed by operator overloading.

src/System.CommandLine/Parsing/ParseOperation.cs
```diff
@@ -246,9 +246,10 @@ private void ParseOptionArguments(OptionResult optionResult)
                         break;
                     }
                 }
-                else if (argument.ValueType == typeof(bool) && 
+                else if ((argument.ValueType == typeof(bool) || argument.ValueType == typeof(bool?))  &&
```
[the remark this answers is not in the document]
jonsequitur: Thanks for pointing it out! I've updated to use that in my other PR.

PR #2217, opened 2023-06-11
rename filename to appropriate one

picu63:
I renamed Option{T}.cs file to new one and changed comment for future mass class renaming


jonsequitur (approved):

jonsequitur:
Thanks, @picu63! Nice catch.

Issue #2223, opened 2023-06-15
name collision behavior is inconsistent for commands, directives, and options added to the same parent


jonsequitur:
Currently, certain naming collisions between symbols under a common parent command will result in a parser configuration whose behavior might not be correct during parsing (e.g. one symbol will always shadow the other) or might throw an exception during parsing. 

* Commands and directives having the same name will throw `System.ArgumentException: An item with the same key has already been added.` even though this will not result in a misbehaving parser, since at parse time the actual directive tokens are differentiated from commands by the presence of surround `[` and `]` brackets. (Related: #2220)
  * A command with brackets in its name (`"[command]"`) could be shadowed by a directive named `command`.
* Commands and options with matching names or aliases will not currently result in an exception during parsing.

jeffhandley:
We don't think this is an issue anymore.

PR #2224, opened 2023-06-15
Move directives to `CliRootCommand`

jonsequitur:


src/System.CommandLine/ChildSymbolList{T}.cs
```diff
@@ -19,7 +19,7 @@ public ChildSymbolList(CliCommand parent)
             _parent = parent;
             _children = new();
         }
-
+
```
adamsitnik, line 22: nit: whitespaces

src/System.CommandLine/CliConfiguration.cs
```diff
@@ -26,17 +25,14 @@ public class CliConfiguration
...
-        /// Gets a mutable list of the enabled directives.
-        /// Currently only <see cref="SuggestDirective"/> is enabled by default.
-        /// </summary>
-        public List<CliDirective> Directives { get; }
+        internal bool HasDirectives =>
+            RootCommand switch
+            {
+                CliRootCommand root => root.Directives.Count > 0,
+                _ => false
+            };
```
adamsitnik, line 35: subjective: IMO a switch statement for two possibilities is not worth it

```cs
internal bool HasDirectives => RootCommand is CliRootCommand root && root.Directives.Count > 0;
```

src/System.CommandLine/Parsing/StringExtensions.cs
```diff
@@ -101,9 +101,15 @@ internal static (string? Prefix, string Alias) SplitPrefix(this string rawAlias)
                             ? arg.Substring(1, colonIndex - 1) // [name:value]
                             : arg.Substring(1, arg.Length - 2); // [name] is a legal directive
 
-                        CliDirective? directive = knownTokens.TryGetValue(directiveName, out var directiveToken)
-                            ? (CliDirective)directiveToken.Symbol!
-                            : null;
+                        CliDirective? directive;
+                        if (knownTokens.TryGetValue($"[{directiveName}]", out var directiveToken))
```
adamsitnik, line 105: So the actual parse error fix was to wrap the directive name in `[]`? 

May I ask why have you decided to move `CliDirective` from config to root command then? I am not against it, I am just curious whether overall you got to a conclusion that it's where it belongs or whether I can't see some other aspect of the actual fix.

src/System.CommandLine/Parsing/StringExtensions.cs
```diff
@@ -405,16 +410,17 @@ static IEnumerable<string> SplitLine(string line)
             }
         }
 
-        private static Dictionary<string, CliToken> ValidTokens(this CliCommand command, IReadOnlyList<CliDirective>? directives)
+        private static Dictionary<string, CliToken> ValidTokens(this CliCommand command)
         {
             Dictionary<string, CliToken> tokens = new(StringComparer.Ordinal);
 
-            if (directives is not null)
+            if (command is CliRootCommand { Directives: IList<CliDirective> directives })
```
adamsitnik, line 417: again subjective on style: what do we benefit from using this syntax here?

```cs            
if (command is CliRootCommand { Directives: IList<CliDirective> directives })
{
    for (int i = 0; i < directives.Count; i++)
```

```cs
if (command is CliRootCommand rootCommand)
{
    for (int i = 0; i < rootCommand.Directives.Count; i++)
```

adamsitnik (commented):
Overall the fix looks very good, thank you @jonsequitur !

I am curious why you have decided to move the directives from config to root command now. I am not against it, but it's going to have non-trivial impact on the future of `RootCommand` (we won't be able to delete it). Are you planning to move more things from the config to the root command?

src/System.CommandLine/Parsing/StringExtensions.cs
jonsequitur: This change isn't ideal but it's a step toward reducing the impact of a larger bug (#2223). It can't be fully fixed without incorporating directives into the symbol model, which is why moving them to `CliRootCommand` at this point was helpful. It's a design decision we discussed a few times for the sake of consistency. Tactically, this was a good time to do it since it related to this bug.

src/System.CommandLine/Parsing/StringExtensions.cs
jonsequitur: This avoids an intermediate assignment of a `CliRootCommand` variable when all we need from it is the directives:

```csharp
if (command is CliRootCommand { Directives: IList<CliDirective> directives })
```

I don't know if this makes any difference from a performance perspective but I find it (subjectively, again) to be less cluttered.

jonsequitur:
> Are you planning to move more things from the config to the root command?

The sense from a few design discussions we've had about this is that since directives are only valid at the root, and since they're symbols, it's consistent and intuitive for them to be children of `CliRootCommand` rather than `CliConfiguration`. 

The other changes we've discussed making to `CliConfiguration` involve factoring out invocation-related members to create a cleaner separation between parser and invocation concepts, and to leave `CliConfiguration` focused specifically on parsing concerns that are cross-cutting and not specific to any symbols, e.g. token replacement / response files, POSIX bundling behaviors, and so on.

src/System.CommandLine/Parsing/StringExtensions.cs
adamsitnik: >  It can't be fully fixed without incorporating directives into the symbol model

Thanks, now I know all I needed! 👍

PR #2241, opened 2023-07-21
Remove src/Directory.Build.(props/targets) from solution since they do not exist.

andrewimcclement:
(Note that the Directory.Build.(props/targets) at the root of the repository are still included under "Solution Items".)


andrewimcclement:
The versions at the repository root are already included, as per my description - see https://github.com/dotnet/command-line-api/blob/main/System.CommandLine.sln#L13-L15.

jonsequitur (approved):

jonsequitur:
Thanks, @andrewimcclement!

PR #2249, opened 2023-07-25
Fix #2226

### LINKED ISSUE #2226: ParseErrorResult creates a new HelpAction for ShowHelp
If you have a created a custom help action for your commands and there is a parse error then the if creates a new HelpAction with the default HelpBuilder.

Wondering if this shouldn't be changed to look for the HelpOption on the command and call that Action instead?

jonsequitur:
This fixes #2226.

### DIFF
--- a/src/System.CommandLine/Invocation/ParseErrorAction.cs
+++ b/src/System.CommandLine/Invocation/ParseErrorAction.cs
@@ -3,6 +3,7 @@
 
 using System.Collections.Generic;
 using System.CommandLine.Help;
+using System.CommandLine.Parsing;
 using System.Linq;
 using System.Threading;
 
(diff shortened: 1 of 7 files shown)


src/System.CommandLine/Invocation/ParseErrorAction.cs
```diff
@@ -59,7 +60,30 @@ private static void WriteErrorDetails(ParseResult parseResult)
 
     private static void WriteHelp(ParseResult parseResult)
     {
-        new HelpAction().Invoke(parseResult);
+        var availableHelpOptions =
```
adamsitnik, line 63: I have fixed this bug once: https://github.com/dotnet/command-line-api/pull/2211, how is it possible that it's back? Some merge issue?

Since `ParseResult` exposes `RootCommandResult` and `RootCommand` should be the only command having it, we can simplify it to:

```cs
HelpOption helpOption = parseResult.RootCommandResult.Command.Options.FirstOrDefault(option => option is HelpOption) as HelpOption ?? new HelpOption();
```

jonsequitur: > Since ParseResult exposes RootCommandResult and RootCommand should be the only command having it...

This isn't the case. Different help options can be defined at different levels of the hierarchy.

jonsequitur:
I've also removed the direct usage of a new `HelpAction`. Now, if there's no `HelpOption` in the tree, then parse errors will not attempt to show help.

Keboo (approved):

src/System.CommandLine/Invocation/ParseErrorAction.cs
```diff
@@ -59,7 +60,30 @@ private static void WriteErrorDetails(ParseResult parseResult)
...
+            parseResult
+                .CommandResult
+                .Command
+                .RecurseWhileNotNull(c => c.Parents.OfType<CliCommand>().FirstOrDefault())
+                .Select(c => c.Options.OfType<HelpOption>().FirstOrDefault());
```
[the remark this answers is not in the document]
jonsequitur: Thanks. I added a test and fixed this in the latest commit.

PR #2311, opened 2023-12-16
chore: Centralize Package Management

BenjaminMichaelis:


BenjaminMichaelis:
Not sure what's causing the failure for the release tests

Directory.Packages.props
```diff
@@ -0,0 +1,30 @@
...
+    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp" Version="4.0.1" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp.Scripting" Version="4.0.1" />
+    <PackageVersion Include="Microsoft.CSharp" Version="4.7.0" />
+    <PackageVersion Include="Microsoft.DotNet.PlatformAbstractions" Version="3.1.6" />
+    <PackageVersion Include="Microsoft.Extensions.Configuration.CommandLine" Version="6.0.0" />
+    <PackageVersion Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="6.0.0" />
+    <PackageVersion Include="Microsoft.Extensions.Hosting" Version="6.0.0" />
+    <PackageVersion Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.0" />
+    <PackageVersion Include="Newtonsoft.Json" Version="13.0.2" />
+    <PackageVersion Include="system.reactive.core" Version="5.0.0" />
```
ViktorHofer, line 21: It would be good to group those dependencies on where they are coming from. See aspire for an example: https://github.com/dotnet/aspire/blob/main/Directory.Packages.props

Directory.Packages.props
```diff
@@ -0,0 +1,30 @@
...
+  </ItemGroup>
+  <ItemGroup Condition="'$(DisableArcade)' == '1'">
+    <PackageVersion Include="xunit" Version="2.4.1" />
+    <PackageVersion Include="xunit.runner.visualstudio" Version="2.4.1" />
+  </ItemGroup>
```
ViktorHofer, line 26: Add a comment to keep the xunit version used in sync with the one that Arcade promotes. Otherwise you would test with different test frameworks which you probably don't want.

Directory.Packages.props
```diff
@@ -0,0 +1,30 @@
...
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+    <PackageVersion Include="System.Memory" Version="4.5.4" />
+  </ItemGroup>
```
ViktorHofer, line 29: I don't think it's recommended to condition PackageVersions based on the TargetFramework. I would just remove the condition. cc @jeffkl for opinions

NuGet.config
```diff
@@ -11,5 +11,24 @@
...
+      <package pattern="Microsoft.DotNet.Build.Tasks.Feed" />
+      <package pattern="Microsoft.DotNet.SignTool" />
+      <package pattern="Microsoft.DotNet.Tar" />
+      <package pattern="Microsoft.SymbolUploader.Build.Task" />
+      <package pattern="Microsoft.DotNet.*" />
+      <package pattern="Microsoft.DiaSymReader.*" />
+      <package pattern="Microsoft.SourceBuild.Intermediate.*" />
+      <package pattern="sn" />
+    </packageSource>
+  </packageSourceMapping>
```
ViktorHofer, line 33: Package source mapping doesn't work with Maestro internal branches (i.e. release and servicing): https://github.com/dotnet/arcade-services/issues/3002

I recommend to do disable the warning: https://github.com/dotnet/format/blob/4425369d3706effc4f5aa67ec16828b07e69c713/Directory.Packages.props#L6-L7

Directory.Packages.props
BenjaminMichaelis: This just came from here: https://github.com/dotnet/command-line-api/blob/a0782c85fd8f7aed16f46935998e9280c86ed2fb/src/System.CommandLine/System.CommandLine.csproj#L28-L30

I can remove it, but I didn't want to remove a condition that was added in this commit without reason to: https://github.com/dotnet/command-line-api/commit/83402ea1f242f2cc86825b8abed2d2bb7cac1c13

ViktorHofer: You still want to keep the condition on the PackageReference itself but you don't need it on the PackageVersion item.

Directory.Packages.props
BenjaminMichaelis: I see the sample but am not really sure on the best grouping in this case, do you have a preference?

ViktorHofer: the Microsoft.CodeAnalysis ones should be under "roslyn", the Microsoft.Extensions ones should be under "runtime". The others are probably "external".

BenjaminMichaelis: Updated!

ViktorHofer: Just FYI, the Microsoft.NETFramework.ReferenceAssemblies entry will be removed with https://github.com/dotnet/command-line-api/pull/2312/files as it isn't used anymore.

NuGet.config
BenjaminMichaelis: Looking at the builds, this looks insufficient so far, any other suggestions or changes?

Directory.Packages.props
BenjaminMichaelis: > Just FYI, the Microsoft.NETFramework.ReferenceAssemblies entry will be removed with

Removed it, will wait for your pr to merge before rebasing

NuGet.config
ViktorHofer: It is sufficient but this project incorrectly overwrites `NoWarn` globally: https://github.com/BenjaminMichaelis/command-line-api/blob/d0270ef03bd9ac9d518197d815f46d0733f35f33/src/System.CommandLine.DragonFruit.Tests/System.CommandLine.DragonFruit.Tests.csproj#L10

This should `<NoWarn>$(NoWarn);1701;1702;1591</NoWarn>` instead.

BenjaminMichaelis: Good spot, thanks

Directory.Packages.props
```diff
@@ -0,0 +1,33 @@
+<Project>
+  <PropertyGroup>
```
ViktorHofer, line 2: nit: for readability, please add an empty line before and after the property group and item groups. See https://github.com/dotnet/format/blob/main/Directory.Packages.props for an example.

BenjaminMichaelis: Done

ViktorHofer (approved):

ViktorHofer:
I approved but we need to wait for one of the authorized users.

ViktorHofer (approved):

jonsequitur (approved):

ViktorHofer:
Thank you @BenjaminMichaelis

Issue #2346, opened 2024-03-07
Create error reporting output


KathleenDollard:
This will be a good first issue for someone who has some understanding of System.CommandLine and wants to understand how the new pipeline design works. It involves finding the error reporting code in System.CommandLine extended (in [`main-powderhouse` branch](https://github.com/dotnet/command-line-api/tree/main-powderhouse)  and putting it in `ErrorReportingSubsystem.Execute`

This part is very easy. However, testing will require resolution of #2345.

If you do not want to wait, consider: 

* Use an internal method in `ErrorReportingSubsystem` that takes a collection to avoid the `ReadonlyList` on `ParseResult`
* Copy 'ParseError' to the `System.CommandLine.Subsystem` (I think that gives it precedence) or copy it and give it a recognizable fake name like `ParseError2`
* Do a draft PR because we can't take this as is. The draft PR will ensure we move the decision in #2345 to the front of the queue of things to resolve

JeanJoeris:
Taking a look at this, will put up a draft PR, but it sounds like #2345 is leaning towards `ParseError` being public so I'll write with that in mind

karshav16:
Would love to contribute!

PR #2369, opened 2024-04-01
Error Reporting Subsystem

JeanJoeris:
This PR aims to address https://github.com/dotnet/command-line-api/issues/2346 by expanding the functionality of the error reporting subsystem.

It adds unit testing around a `Report` method that preserves old reporting logic. After discussions with @KathleenDollard it was decided to make `ParseError` public, as Powderhouse is moving towards a world where we expect errors might be created by a variety of subsystems through different projects

It also extends the pattern used elsewhere in Powderhouse of renaming old tests to be suffixed with `FunctionalTests` as preserving old logic is essentially an integration test. However, these functional tests rely on help functionality not yet present so are commented out.

Would love any thoughts or feedback


src/System.CommandLine.Subsystems/ErrorReportingSubsystem.cs
```diff
@@ -1,24 +1,49 @@
...
+        ConsoleHelpers.SetTerminalForegroundRed();
+
+        foreach (var error in errors)
+        {
+            consoleHack.WriteLine(error.Message);
+        }
+        consoleHack.WriteLine();
+
+        ConsoleHelpers.ResetTerminalForegroundColor();
+    }
```
JeanJoeris, line 48: `Report` was largely pulled out to enable easier testing. It is currently public to enable testing, as IVT of subsystem to subsystem.tests, without also adding @KathleenDollard 's version subsystem work, breaks compilation

JeanJoeris:
@dotnet-policy-service agree

src/System.CommandLine.Subsystems.Tests/ErrorReportingSubsystemTests.cs
```diff
@@ -0,0 +1,70 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Reflection;
+using FluentAssertions;
+using Xunit;
+using System.CommandLine.Parsing;
+using static System.Runtime.InteropServices.JavaScript.JSType;
+
+namespace System.CommandLine.Subsystems.Tests
```
Keboo, line 10: We should file-scope the namespace on new files.
```suggestion
namespace System.CommandLine.Subsystems.Tests;
```

src/System.CommandLine.Subsystems.Tests/System.CommandLine.Subsystems.Tests.csproj
```diff
@@ -29,6 +29,8 @@
     <Compile Include="PipelineTests.cs" />
     <Compile Include="VersionFunctionalTests.cs" />
     <Compile Include="VersionSubsystemTests.cs" />
+    <Compile Include="ErrorReportingFunctionalTests.cs" />
+    <Compile Include="ErrorReportingSubsystemTests.cs" />
```
Keboo, line 33: Are these needed? I would have expected the default SDK to pick up *.cs files in the root on the project.

JeanJoeris: We're currently using `<EnableDefaultCompileItems>False</EnableDefaultCompileItems>` to default exclude files and avoid compilation errors in files that aren't working yet. These will be ripped out later but needed at the moment

src/System.CommandLine.Subsystems.Tests/ErrorReportingSubsystemTests.cs
JeanJoeris: Good catch

src/System.CommandLine.Subsystems.Tests/ErrorReportingSubsystemTests.cs
```diff
@@ -0,0 +1,68 @@
...
+        errorSubsystem.Report(consoleHack, errors);
+
+        consoleHack.GetBuffer().Trim().Should().Be("");
+    }
+
+    [Theory]
+    [InlineData("-v", false)]
+    [InlineData("-x", true)]
+    [InlineData("", false)]
+    public void GetIsActivated_tests(string input, bool result)
```
BenjaminMichaelis, line 55:
```suggestion
    public void GetIsActivated_GivenParseResult_ShouldReturnExpectedResult(string input, bool result)
```
Tried coming up with a better name, but I would consider splitting this into two tests, one with expected of true, and one of the false cases.

I would argue this also makes naming the test way easier because the expected result is clear, and when a failure occurs, the expected result is more clear making the expectation more clear.

src/System.CommandLine.Subsystems/ConsoleHack.cs
```diff
@@ -10,7 +10,7 @@ public class ConsoleHack
     private readonly StringBuilder buffer = new();
     private bool redirecting = false;
 
-    public void WriteLine(string text)
+    public void WriteLine(string text = "")
```
BenjaminMichaelis, line 13: Is empty string actually a valid default?

src/System.CommandLine.Subsystems/ErrorReportingSubsystem.cs
```diff
@@ -1,24 +1,49 @@
...
+
+        Report(pipelineContext.ConsoleHack, pipelineContext.ParseResult.Errors);
+
         return CliExit.SuccessfullyHandled(pipelineContext.ParseResult);
     }
+
+    public void Report(ConsoleHack consoleHack, IReadOnlyList<ParseError> errors)
+    {
+        ConsoleHelpers.ResetTerminalForegroundColor();
+        ConsoleHelpers.SetTerminalForegroundRed();
```
BenjaminMichaelis, line 39: Is a Reset needed if we are about to set it?
```suggestion
        ConsoleHelpers.SetTerminalForegroundRed();
```

src/System.CommandLine.Subsystems/ConsoleHack.cs
BenjaminMichaelis: I think I would re-write WriteLine to act slightly different here if this was going to stay and to behave more like Console.WriteLine, but I think ConsoleHack is going to disappear so we can ignore this

src/System.CommandLine.Subsystems.Tests/ErrorReportingSubsystemTests.cs
JeanJoeris: Good call, definitely should be two tests 👍

src/System.CommandLine.Subsystems/ConsoleHack.cs
JeanJoeris: It's valid, at least in current `main` behavior. See my other comment for the code that uses this call, but this was done to preserve the existing `WriteLine()` interface for writing an empty newline

src/System.CommandLine.Subsystems/ErrorReportingSubsystem.cs
JeanJoeris: This code is just pulled from the existing `WriteErrorDetails` method on `ParseErrorAction` from main.

https://github.com/dotnet/command-line-api/blob/963d34b1fb712c673bfb198133d7e988182c9ef4/src/System.CommandLine/Invocation/ParseErrorAction.cs#L47-L60

The idea is pulling in that existing code from `main` into `main-powderhouse` we return `main` functionality over `main-powderhouse`'s current `pipelineContext.ConsoleHack.WriteLine("You have errors!");`. I think it's quite likely that this needs to change / be updated but I'm not sure that should be done in this PR

src/System.CommandLine.Subsystems.Tests/ErrorReportingSubsystemTests.cs
JeanJoeris: Pushed up a stab at that, lemme know what you think. The tests are repeated but the test name really clarifies intent and the not-necessarily-intuitive boolean behavior. For the error subsystem, GetIsActived should
- return true when there is an invalid input, or when there _is_ an error
- return false when there is a valid input, or when there _is not_ an error

src/System.CommandLine.Subsystems/ErrorReportingSubsystem.cs
JeanJoeris: I reference this in my other comment, but `parseResult.Configuration.Error.WriteLine();` on line 54 is why I added a zero-arity version of `WriteLine` in this PR

src/System.CommandLine.Subsystems.Tests/ErrorReportingSubsystemTests.cs
```diff
@@ -0,0 +1,84 @@
...
+        var rootCommand = new CliRootCommand {new CliOption<bool>("-v")};
+        var configuration = new CliConfiguration(rootCommand);
+        var errorSubsystem = new ErrorReportingSubsystem();
+        IReadOnlyList<string> args = [""];
+        Subsystem.Initialize(errorSubsystem, configuration, args);
+
+        var parseResult = CliParser.Parse(rootCommand, input, configuration);
+        var isActive = Subsystem.GetIsActivated(errorSubsystem, parseResult);
+
+        isActive.Should().Be(result);
```
BenjaminMichaelis, line 65:
```suggestion
    [Theory]
    [InlineData("-x")]
    [InlineData("-non_existant_option")]
    public void GetIsActivated_GivenInvalidInput_SubsystemIsActive(string input,)
    {
        var rootCommand = new CliRootCommand {new CliOption<bool>("-v")};
        var configuration = new CliConfiguration(rootCommand);
        var errorSubsystem = new ErrorReportingSubsystem();
        IReadOnlyList<string> args = [""];
        Subsystem.Initialize(errorSubsystem, configuration, args);

        var parseResult = CliParser.Parse(rootCommand, input, configuration);
        var isActive = Subsystem.GetIsActivated(errorSubsystem, parseResult);

        isActive.Should().BeTrue();
```
No need for the second result parameter

Is this test name slightly more clear?

src/System.CommandLine.Subsystems.Tests/ErrorReportingSubsystemTests.cs
```diff
@@ -0,0 +1,84 @@
...
+    }
```
BenjaminMichaelis, line 83:
```suggestion
    [Theory]
    [InlineData("-v")]
    [InlineData("")]
    public void GetIsActivated_GivenValidInput_SubsystemShouldBeActive(string input)
    {
        var rootCommand = new CliRootCommand { new CliOption<bool>("-v") };
        var configuration = new CliConfiguration(rootCommand);
        var errorSubsystem = new ErrorReportingSubsystem();
        IReadOnlyList<string> args = [""];
        Subsystem.Initialize(errorSubsystem, configuration, args);

        var parseResult = CliParser.Parse(rootCommand, input, configuration);
        var isActive = Subsystem.GetIsActivated(errorSubsystem, parseResult);

        isActive.Should().BeFalse();
    }
```
No need for the second result parameter

Is this test name slightly more clear?

JeanJoeris: Thanks for catching that second arg, and I think those test names are even clearer. Really drives home what the boolean values for GetIsActivated actually mean in business logic 👍 

Appreciate your feedback

KathleenDollard (approved):

src/System.CommandLine.Subsystems/ErrorReportingSubsystem.cs
```diff
@@ -1,24 +1,49 @@
...
-    // TODO: Stash option rather than using string
     protected internal override bool GetIsActivated(ParseResult? parseResult)
         => parseResult is not null && parseResult.Errors.Any();
 
+    // TODO: properly test execute directly when parse result is usable in tests
     protected internal override CliExit Execute(PipelineContext pipelineContext)
     {
-        pipelineContext.ConsoleHack.WriteLine("You have errors!");
+        var _ = pipelineContext.ParseResult
+            ?? throw new ArgumentNullException($"{nameof(pipelineContext)}.ParseResult");
```
Keboo, line 29: It is un-usual to have the name of an argument exception not directly map to the argument name of the method. Typically when it is a nested property we just use an ArgumentException with a message (NB: ArgumentException, and ArgumentNullException do not do their constructor parameters in the same order).

Also we likely need to localize the exception message.

```suggestion
            ?? throw new ArgumentException("The parse result has not been set", nameof(pipelineContext));
```

Keboo (approved):

PR #2371, opened 2024-04-04
Powderhouse extensibility overview doc

mhutch:
The shape of some of the APIs has been superseded since this was originally written but the conceptual overview is still useful

### DIFF
--- a/docs/proposals/extensibility-overview.md
+++ b/docs/proposals/extensibility-overview.md
@@ -1,0 +1,369 @@
+# System.CommandLine Extensibility
+
+## Overview
+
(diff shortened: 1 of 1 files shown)


KathleenDollard:
I think we should consider a different file layout for proposals. 

If we create a "Powderhouse" folder in docs, or we create a "Proposals" folder under the root or in docs, it will be easier to find documents like this. I plan to switch to markdown docs for the proposals I have been putting into issues, where they seem big enough. 

Also, I think we should consider a corresponding issue for proposals where we can get feedback and record changes until the document is updated. 

I'm doing this for the proposal I am putting up today, so we can resolve the layout in discussing these PRs.

KathleenDollard:
Also, what kind of review do you want on this? I have been working with this doc and I think implemented the annotations as described, but I now have a more depth on the pipeline. So, I approve, but have not reviewed it in detail now.

KathleenDollard (approved):

KathleenDollard (approved):

PR #2378, opened 2024-04-10
Powderhouse output proposal

KathleenDollard:
Tentative proposal that will make creating help and error reporting and outputting them in multiple formats quite easy. 

For example, the gnarly code to determine the right space for column widths would occur only once, and the fact this code isn't needed in markdown would mean it would not run.


JeanJoeris (commented):
`docs/Proposals/powderhouse-output.md` should be changed to `docs/proposals/powderhouse-output.md` to match other proposal PRs

mhutch:
The `docs/Proposals` directory should be `docs/proposals` for consistency with other PRs and to avoid conflicts between case-sensitive and case-insensitive filesystems.

PR #2386, opened 2024-04-20
Removed tokens from ParseResult

KathleenDollard:
This is part of cleaning out ParseResult to provide a desired public API without cruft from parsing


JeanJoeris (commented):
No stylistic concerns or nits, looks very reasonable for temporarily pulling out tokens

mhutch (approved):

Issue #2392, opened 2024-04-23
[main] Test has incorrect errors for `-x 1 2 3 4`


KathleenDollard:
[The test I used to confirm unexpected behavior](https://github.com/dotnet/command-line-api/blob/963d34b1fb712c673bfb198133d7e988182c9ef4/src/System.CommandLine.Tests/ParserTests.cs#L1550)

The test is:

```csharp
      [Fact]
      public void When_option_arguments_are_greater_than_maximum_arity_then_an_error_is_returned()
      {
          var command = new CliCommand("the-command")
                  {
                      new CliOption<int[]>("-x") { Arity = new ArgumentArity(2, 3)}
                  };

          var parseResult = CliParser.Parse(command, "-x 1 2 3 4");
          parseResult.Errors
                     .Select(e => e.Message)
                     .Should()
                     .Contain(LocalizationResources.UnrecognizedCommandOrArgument("4"));
      }
```

I modified the test only to get an interim variable for `parseResult`,

When I do this, the error list is

![image](https://github.com/dotnet/command-line-api/assets/5844318/47b49213-b4d0-43a0-a29e-5663eb80c9a1)

I encountered this because for Powderhouse I encountered issues with options with collection types and I believe I am using the same or funcionally equivalent code - and this test passes. In exploring why my new tests were failing, I realized this test does not appear to be succeeding at what the CLI and the commandline request.

I believe this test should result in a single parse error - an unexpected argument only on the 4.

Keboo:
I think this may be a bug (even on main). 

First, the test input is bad. The [successful test a couple up](https://github.com/dotnet/command-line-api/blob/963d34b1fb712c673bfb198133d7e988182c9ef4/src/System.CommandLine.Tests/ParserTests.cs#L1499) shows how an option's argument with a minimum arity greater than 1 should be tested. Specifically the test input should be:
`command.Parse("-x 1 -x 2 -x 3 -x 4")`.

Now with that, said, even once that is corrected that will only solve your first three "Unrecognized command or argument" errors from your screenshot. You are then left with two "Option '-x' expects a single argument but 4 were provided." errors. I would propose both the error message is incorrect, as well as the fact that there are two errors (I agree there really should only be one). This is not something new with powerhouse, this is existing behavior on main.

elgonzo:
For this test to succeed, it would be required to set the option's `AllowMultipleArgumentsPerToken` property to `true`:
[...]

PR #2402, opened 2024-04-25
Fixing issue #2392

Keboo:
There were two problems addressed here.
First, the call to Diagram was causing the second error to be added to the ParseResult.Errors collection. The diagram call, was being invoked by the debugger as part of the ParseResult.ToString() method. Causing the number of errors to change when inspecting a parse result.

The second issue was the error message text. It was implying that the option expected a single value, when its arity allowed for more.

### DIFF
--- a/src/System.CommandLine/LocalizationResources.cs
+++ b/src/System.CommandLine/LocalizationResources.cs
@@ -14,10 +14,22 @@ namespace System.CommandLine
     internal static class LocalizationResources
     {
         /// <summary>
-        ///   Interpolates values into a localized string similar to Command &apos;{0}&apos; expects a single argument but {1} were provided.
+        ///   Interpolates values into a localized string similar to Option &apos;{0}&apos; expects a single argument but {1} were provided.
         /// </summary>
         internal static string ExpectsOneArgument(OptionResult optionResult)
             => GetResourceString(Properties.Resources.OptionExpectsOneArgument, GetOptionName(optionResult), optionResult.Tokens.Count);
+
+        /// <summary>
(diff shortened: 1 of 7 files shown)


src/System.CommandLine/LocalizationResources.cs
```diff
@@ -14,11 +14,23 @@ namespace System.CommandLine
     internal static class LocalizationResources
     {
         /// <summary>
-        ///   Interpolates values into a localized string similar to Command &apos;{0}&apos; expects a single argument but {1} were provided.
+        ///   Interpolates values into a localized string similar to Option &apos;{0}&apos; expects a single argument but {1} were provided.
```
KathleenDollard, line 17: Isn't this used by both options and commands?

src/System.CommandLine/Parsing/CommandResult.cs
```diff
@@ -125,7 +125,7 @@ private void ValidateOptions(bool completeValidation)
                 // When_there_is_an_arity_error_then_further_errors_are_not_reported
                 if (!ArgumentArity.Validate(argumentResult, out var error))
                 {
-                    optionResult.AddError(error.ErrorMessage!);
+                    argumentResult.AddError(error.ErrorMessage!);
```
KathleenDollard, line 128: I do not understand this change. If this is on options, I thought we got the errors from the option. Do we get them from the option's argument?

KathleenDollard: I want to discuss the design here. I do not think the Diagram subsystem should know that `CliOption` treats its argument as a `CliArgument`. My intuition is that this problem will appear in numerous places. 

I'd like to discuss this prior to merging.

PR #2405, opened 2024-04-28
Make ArgumentResult, OptionResult and CommandResult `internal`

KathleenDollard:
ArgumentResult and OptionResult are being replaced with the more API friendly ValueResult. 

CommandResult is being replaced with the more API friendly class that has the temporary name of CommandValueResult.

### DIFF
--- a/src/System.CommandLine.Subsystems/Directives/DiagramSubsystem.cs
+++ b/src/System.CommandLine.Subsystems/Directives/DiagramSubsystem.cs
@@ -35,7 +35,7 @@ public class DiagramSubsystem( IAnnotationProvider? annotationProvider = null)
         var builder = new StringBuilder(100);
 
 
-        Diagram(builder, parseResult.RootCommandResult, parseResult);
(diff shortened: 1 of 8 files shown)


src/System.CommandLine.Subsystems/Directives/DiagramSubsystem.cs
```diff
@@ -35,7 +35,7 @@ internal static StringBuilder Diagram(ParseResult parseResult)
         var builder = new StringBuilder(100);
 
 
-        Diagram(builder, parseResult.RootCommandResult, parseResult);
+        //Diagram(builder, parseResult.RootCommandResult, parseResult);
```
JeanJoeris, line 38: Discussion from meeting clarified this is being commented out due to it being broken.  Can we get a todo on fixing this to capture this?

mhutch (approved):

JeanJoeris (commented):
Makes sense and looks groovy

PR #2406, opened 2024-04-28
Renamed `Location.Start` to `Location.Index`

KathleenDollard:
Start may imply position in string, and this is position in the args array or response file.


JeanJoeris (commented):
Makes sense, straight forward low-risk rename 🔥

mhutch (approved):
Definitely makes things clearer, easy to confuse `Start` and `Offset` otherwise. Would be even nicer to have doc comments on `Location` ;)

PR #2407, opened 2024-04-28
Changed Pipeline creation from constructors to factory methods

KathleenDollard:
I think we want really good discoverability on available pipelines. Constructors are not great on this, and I think the ergonomics of factory methods is better. If you know you want a pipeline, you can see what ones are available.

There are several options:

* Make the more common create with standard features `Create` and the empty one `CreateEmpty`
* Make the empty one `Create` because it is empty and make the other something like `CreateStandard`
* Make neither `Create` and use `CreateEmpty` and `CreateStandard`
* Make them both `Create` and use parameters (make a suggestion, I didn't like what I tried here)

However we declare it, should empty look like standard with null instead of the default subsystems to aid discoverability, or should we use properties?

Is it bad to allow the subsystems to change after construction. _Interesting_ things could happen. 

This PR has the first option, and I kept an extra commit to show where I made mistakes, and the fixes show that the property requires two lines.

Looking forward to opinions.


mhutch (approved):
I like this pattern :)

PR #2413, opened 2024-05-02
Value subsystem and core changes

KathleenDollard:
Apologies for the big PR. 

As I worked on the ValueSubsystem, I encountered unexpected issues in the core classes and had to fix them.

I also had a few things -scope changes and pipeline cleanup that I thought I could pull out and a conflict prohibited.

Looking forward to feedback.


KathleenDollard:
Just changed to draft when I discovered whitespace issues, particularly in ParseResult. Need to walk dog, then will try to fix tonight.

src/System.CommandLine.Subsystems.Tests/ValueSubsystemTests.cs
```diff
@@ -0,0 +1,94 @@
...
+        var isActive = Subsystem.GetIsActivated(subsystem, parseResult);
+
+        isActive.Should().BeTrue();
+    }
+
+    [Fact]
+    public void ValueSubsystem_returns_values_that_are_entered()
+    {
+        var consoleHack = new ConsoleHack().RedirectToBuffer(true);
+        CliOption<int> option1 = new CliOption<int>("--intValue");
```
BenjaminMichaelis, line 37:
```suggestion
        CliOption<int> option = new CliOption<int>("--intValue");
```
There seems to be only one `option` variable, can this drop the suffix?

This exists multiple places in this file

src/System.CommandLine.Subsystems.Tests/ValueSubsystemTests.cs
```diff
@@ -0,0 +1,94 @@
...
+        CliRootCommand rootCommand = [
+            new CliCommand("x")
+            {
+                option1
+            }];
+        var configuration = new CliConfiguration(rootCommand);
+        var pipeline = Pipeline.CreateEmpty();
+        pipeline.Value = new ValueSubsystem();
+        const int expected1 = 42;
```
BenjaminMichaelis, line 46:
```suggestion
        const int expected = 42;
```
There seems to be only one `expected` variable, can this drop the suffix? (Great choice of numbers btw)

This exists multiple places in this file

src/System.CommandLine.Subsystems.Tests/ValueSubsystemTests.cs
```diff
@@ -0,0 +1,94 @@
...
+        CliRootCommand rootCommand = [option1];
+        var configuration = new CliConfiguration(rootCommand);
+        var pipeline = Pipeline.CreateEmpty();
+        pipeline.Value = new ValueSubsystem();
+        pipeline.Value.DefaultValue.Set(option1, 43);
+        const int expected1 = 43;
+        var input = $"";
+
+        var parseResult = pipeline.Parse(configuration, input); // assigned for debugging
+        pipeline.Execute(configuration, input, consoleHack);
```
BenjaminMichaelis, line 69:
```suggestion

        pipeline.Execute(configuration, "", consoleHack);
```
Removing lines that exist just for debugging and unnecessary assignment

This exists multiple places in this file

src/System.CommandLine.Subsystems/Subsystems/Annotations/AnnotationAccessor.cs
```diff
@@ -8,12 +8,33 @@ namespace System.CommandLine.Subsystems.Annotations;
 /// <summary>
 /// Allows associating an annotation with a <see cref="CliSymbol"/>. The annotation will be stored by the accessor's owner <see cref="CliSubsystem"/>.
 /// </summary>
+/// <remarks>
+/// The annotation will be stored by the accessor's owner <see cref="CliSubsystem"/>. 
+/// </summary>
+/// <typeparam name="TValue">The type of value to be stored</typeparam>
+/// <param name="owner">The subsystem that this annotation store data for.</param>
+/// <param name="id">The identifier for this annotation, since subsystems may have multiple annotations.</param>
 public struct AnnotationAccessor<TValue>(CliSubsystem owner, AnnotationId<TValue> id)
```
BenjaminMichaelis, line 17: Just as a heads up, this will have conflicts with https://github.com/dotnet/command-line-api/pull/2412

src/System.CommandLine.Tests/ParseResultValueTests.cs
```diff
@@ -0,0 +1,55 @@
...
+                        option1,
+                        option2
+                    };
+
+        var parseResult = CliParser.Parse(rootCommand, "--opt1 Kirk");
+
+        var symbol1 = parseResult.GetSymbolByName("--opt1");
+        var symbol2 = parseResult.GetSymbolByName("--opt2");
+        symbol1.Should().Be(option1);
+        symbol2.Should().Be(option2);
```
BenjaminMichaelis, line 30:
```suggestion
        using (new AssertionScope())
        {
            symbol1.Should().Be(option1);
            symbol2.Should().Be(option2);
        }
```
This could be a nice place to use an assertion scope if we are going to have multiple assertions in one test?

src/System.CommandLine.Subsystems/Subsystems/Annotations/AnnotationAccessor.cs
KathleenDollard: Did you also add a XML Comment? I can remove mine if you did.

src/System.CommandLine.Subsystems.Tests/ValueSubsystemTests.cs
KathleenDollard: Thanks!

src/System.CommandLine.Subsystems.Tests/ValueSubsystemTests.cs
KathleenDollard: Thanks!

src/System.CommandLine.Tests/ParseResultValueTests.cs
KathleenDollard: Great idea. I think this case warrants to two options as one state.

src/System.CommandLine.Subsystems/Subsystems/Annotations/AnnotationAccessor.cs
BenjaminMichaelis: I did not, but am happy to bring these comments over to my pr if you want to drop them to make it easier since I should add some for the new parameter I add and the new `Get` method anyways

src/System.CommandLine.Subsystems.Tests/ValueSubsystemTests.cs
BenjaminMichaelis: Quite the onion! This may be one of those situations where cutting into it causes more tears than may be worth it at this point in time :)

One minor thing we could consider doing if it is possible (and I am not sure without looking into parseResult), is if that can be converted into some easily readable string format, we could put that as the because parameter in the assertion so we can get some contextual information at the least without having to actually just run the test in debug mode? Not sure if we can get a nice string with some good information, but that is my most simple idea so far to potentially take a step in the right direction without solving the issue entirely.

src/System.CommandLine/ParseResult.cs
```diff
@@ -162,17 +178,41 @@ public IReadOnlyList<string> UnmatchedTokens
...
+        public ValueResult? GetValueResult(CliOption option)
+            => GetValueResultInternal(option);
+
+        /// <summary>
+        /// Gets the result, if any, for the specified argument.
+        /// </summary>
+        /// <param name="argument">The argument for which to find a result.</param>
+        /// <returns>A result for the specified argument, or <see langword="default"/> if it was not entered by the user.</returns>
+        public ValueResult? GetValueResult(CliArgument argument)
+            => GetValueResultInternal(argument);
```
BenjaminMichaelis, line 211: Is it intentional that in the comments on these, one uses `langword="default"` and one `langword="null"` when it seems that both call the same method that currently returns null if no result?

src/System.CommandLine/Parsing/SymbolLookupByName.cs
```diff
@@ -0,0 +1,175 @@
...
+            return cache;
+        }
+        for (int i = 1; i < cache.Count; i++) // we tested for 0 earlier
+        {
+            if (cache[i].Command == currentCommand)
+            {
+                return cache.Skip(i);
+            }
+        }
+        return null;
```
BenjaminMichaelis, line 173:
```suggestion
        int index = cache.FindIndex(c => c.Command == currentCommand);
        if (index == -1)
        {
            return null;
        }
        return cache.Skip(index);
```
I haven't found where this is used Could we do something with `.FindIndex` here instead of a for loop? It seems like it may be slightly easier to read and a bit simpler, and I'm not sure the short circuit at the top is saving much in terms of execution time, but if we think it is we could still keep that?

src/System.CommandLine/Parsing/SymbolLookupByName.cs
```diff
@@ -0,0 +1,175 @@
...
+    /// <summary>
+    /// Gets the symbol with the requested name that appears nearest to the starting command, which defaults to the current or leaf command.
+    /// </summary>
+    /// <param name="name">The name to search for</param>
+    /// <param name="startCommand">The command to start searching up from, which defaults to the current command.</param>
+    /// <param name="skipAncestors">If true, only the starting command and no ancestors are searched.</param>
+    /// <param name="valuesOnly">If true, commands are ignored and only options and arguments are found.</param>
+    /// <returns>A tuple of the found symbol and its parent command. Throws if the name is not found.</returns>
+    /// <exception cref="InvalidOperationException">Thrown if the name is not found.</exception>
+    public (CliSymbol symbol, CliCommand parent) GetSymbolAndParent(string name, CliCommand? startCommand = null, bool skipAncestors = false, bool valuesOnly = false)
```
BenjaminMichaelis, line 141: Where does this get used? Should it be tested?

src/System.CommandLine/ParseResult.cs
```diff
@@ -15,48 +15,50 @@ namespace System.CommandLine
     public sealed class ParseResult
     {
         private readonly IReadOnlyDictionary<CliSymbol, ValueResult> valueResultDictionary = new Dictionary<CliSymbol, ValueResult>();
+        private SymbolLookupByName? symbolLookupByName = null;
 
-        private readonly CommandResult _rootCommandResult;
-// TODO: unmatched tokens, invocation, completion
-/*
+        private readonly CommandResult? _rootCommandResult;
```
BenjaminMichaelis, line 20: Why make this nullable? It seems like in it's assignment in the constructor we don't expect it to be null

src/System.CommandLine.Tests/ParserTests.cs
```diff
@@ -1874,6 +1874,28 @@ public void Location_offset_correct_when_colon_or_equal_used()
...
+            var expectedLocation2 = new Location("Spock", Location.User, 3, expectedOuterLocation);
+            var expectedLocation3 = new Location("Uhura", Location.User, 4, expectedOuterLocation);
+
+            var parseResult = CliParser.Parse(rootCommand, "subcommand --opt1 Kirk Spock Uhura");
+
+            var commandResult = parseResult.CommandResult;
+            parseResult.GetValue(option1);
+            var result1 = parseResult.GetValueResult(option1);
+            result1.Locations.Should().BeEquivalentTo([expectedLocation1, expectedLocation2, expectedLocation3]);
+        }
```
BenjaminMichaelis, line 1897:
```suggestion
            var option = new CliOption<string[]>("--opt1");
            option.AllowMultipleArgumentsPerToken = true;
            var rootCommand = new CliRootCommand
                    {
                        option
                    };
            var expectedOuterLocation = new Location("testhost", Location.User, -1, null);
            var expectedLocation1 = new Location("Kirk", Location.User, 2, expectedOuterLocation);
            var expectedLocation2 = new Location("Spock", Location.User, 3, expectedOuterLocation);
            var expectedLocation3 = new Location("Uhura", Location.User, 4, expectedOuterLocation);

            var parseResult = CliParser.Parse(rootCommand, "subcommand --opt1 Kirk Spock Uhura");

            var commandResult = parseResult.CommandResult;
            parseResult.GetValue(option);
            var result = parseResult.GetValueResult(option);
            result.Locations.Should().BeEquivalentTo([expectedLocation1, expectedLocation2, expectedLocation3]);
```
Changing `option1` to `option` and `result1` to `result`.

- Having `expectedLocation1` (and 2 and 3) have index's of 2,3,4 is a bit weird for some reason, is there a reason for this? Just trying to understand what is happening here.
- Is the assignment to `commandResult` for just debugging? And another onion issue here for debugging as with the `valueSubSystemTests`?

src/System.CommandLine.Subsystems/Subsystems/Annotations/AnnotationAccessor.cs
KathleenDollard: Let's do it that way. Copy away and let me know when I should delete them.

BenjaminMichaelis: Copying complete :)

https://github.com/dotnet/command-line-api/pull/2412/commits/ff932a49e1b7d5fa03e40d645bc90da038a83bf0

src/System.CommandLine.Subsystems.Tests/ValueSubsystemTests.cs
```diff
@@ -0,0 +1,94 @@
...
+using System.CommandLine.Directives;
+using System.CommandLine.Parsing;
+using Xunit;
+
+namespace System.CommandLine.Subsystems.Tests;
+
+public class ValueSubsystemTests
+{
+    [Fact]
+    public void Value_is_always_activated()
```
KathleenDollard, line 14:
```suggestion
    public void ValueSubsystem_is_activated_by_default()
```

src/System.CommandLine.Subsystems.Tests/ValueSubsystemTests.cs
```diff
@@ -0,0 +1,94 @@
...
+        var consoleHack = new ConsoleHack().RedirectToBuffer(true);
+        CliOption<int> option = new CliOption<int>("--intValue");
+        CliRootCommand rootCommand = [option];
+        var configuration = new CliConfiguration(rootCommand);
+        var pipeline = Pipeline.CreateEmpty();
+        pipeline.Value = new ValueSubsystem();
+        var x = 42;
+        pipeline.Value.DefaultValueCalculation.Set(option, () => x + 2);
+        const int expected = 44;
+        var input = $"";
```
Keboo, line 87: nit:
```suggestion
        var input = "";
```

src/System.CommandLine.Subsystems/Pipeline.cs
```diff
@@ -131,11 +81,13 @@ protected virtual void ExecuteErrorReporting(PipelineContext context)
...
-        InitializeHelp(context);
-        InitializeVersion(context);
-        InitializeCompletion(context);
-        InitializeDiagram(context);
-        InitializeErrorReporting(context);
+        foreach (var subsystem in Subsystems)
+        {
+            if ( subsystem is not null)
+            {
+            subsystem.Initialize(context);
```
Keboo, line 88: whitespace. It looks like this needs indentation

src/System.CommandLine.Subsystems/Pipeline.cs
```diff
@@ -131,11 +81,13 @@ protected virtual void ExecuteErrorReporting(PipelineContext context)
...
     protected virtual void InitializeSubsystems(InitializationContext context)
     {
-        InitializeHelp(context);
-        InitializeVersion(context);
-        InitializeCompletion(context);
-        InitializeDiagram(context);
-        InitializeErrorReporting(context);
+        foreach (var subsystem in Subsystems)
+        {
+            if ( subsystem is not null)
```
Keboo, line 86:
```suggestion
            if (subsystem is not null)
```

src/System.CommandLine.Subsystems/Pipeline.cs
```diff
@@ -4,11 +4,16 @@
 using System.CommandLine.Directives;
 using System.CommandLine.Parsing;
 using System.CommandLine.Subsystems;
+using System.Reflection.PortableExecutable;
```
KathleenDollard, line 7: Why? Remove

src/System.CommandLine.Subsystems/Subsystems/Annotations/ValueAnnotationAccessor.cs
```diff
@@ -0,0 +1,63 @@
...
+/// <summary>
+/// Associates an annotation with a <see cref="CliSymbol"/>. The symbol must be an option or argument and the value must be of the same type as the symbol./>. 
+/// </summary>
+/// <remarks>
+/// The annotation will be stored by the accessor's owner <see cref="CliSubsystem"/>. 
+/// </remarks>
+/// <typeparam name="TValue">The type of value to be stored</typeparam>
+/// <param name="owner">The subsystem that this annotation store data for.</param>
+/// <param name="id">The identifier for this annotation, since subsystems may have multiple annotations.</param>
+public struct ValueAnnotationAccessor<TValue>(CliSubsystem owner, AnnotationId<TValue> id)
```
KathleenDollard, line 17: Add extension method that parallels the With extension method.

for

```
symbol.With(help.Description, "description");
```

src/System.CommandLine.Subsystems/Subsystems/Annotations/ValueAnnotationAccessor.cs
```diff
@@ -0,0 +1,63 @@
...
+    public readonly bool TryGet<TSymbolValue>(CliSymbol symbol, [NotNullWhen(true)] out TValue? value)
+        where TSymbolValue : TValue
+        => TryGetInternal<TSymbolValue>(symbol, out value);
+
+    private readonly bool TryGetInternal<TSymbolValue>(CliSymbol symbol, [NotNullWhen(true)] out TValue? value)
+        where TSymbolValue : TValue
+    {
+        if (owner.TryGetAnnotation(symbol, id, out var storedValue))
+        {
+            value = (TSymbolValue)storedValue;
```
KathleenDollard, line 57: Consider the right behavior

Throw a more specific exception
or possibly return false and default

Open question

mhutch (approved):

PR #2416, opened 2024-05-04
Fix duplicate errors appended

### LINKED ISSUE #2392: [main] Test has incorrect errors for `-x 1 2 3 4`

Keboo:
Fixes #2392.
This is a port from the fix on main. However due to additional work that is currently being done on Powerderhouse it is forcing the creation of the ArgumentConversionResult as part of parsing. This is due to the invocation SymbolResultTree.GetValueResultDictionary() which forces the evaluation of all ArgumentConversionResults. This is adding an additional test to make sure that the behavior is not re-introduced. This also fixes a few NRT issues and a comment typo.

### DIFF
--- a/src/System.CommandLine.Tests/ParserTests.cs
+++ b/src/System.CommandLine.Tests/ParserTests.cs
@@ -4,13 +4,11 @@
 using System.Collections.Generic;
 using System.CommandLine.Parsing;
 using System.CommandLine.Tests.Utility;
-using System.IO;
 using FluentAssertions;
 using FluentAssertions.Equivalency;
 using System.Linq;
 using FluentAssertions.Common;
 using Xunit;
-using Xunit.Abstractions;
 
 namespace System.CommandLine.Tests
 {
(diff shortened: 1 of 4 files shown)


KathleenDollard:
I'd like to wait to approve this PR until we get the `ValueSubsystem` PR in, as it seems easier to reapply this one. 

Also, I do not understand the change to `ValueResult`. The double call to `GetArgumentConversionResult` should not have caused any extra work as the result appears cached in a field and used by `GetArgumentConversionResult`. 

I do not object to these changes, just want to understand this. I find this corner of the existing code confusing, at least in naming, and hope we can unwind some of that complexity after we have all the tests reinstated. 

The other overall comment: I anticipated more changes in tests. Did we change the test where we discovered this problem in a different PR? (The test that had the false positive in relation to its name)

KathleenDollard (approved):

PR #2421, opened 2024-05-06
PipelineContext replaced with PipelineResult

KathleenDollard:
Previously, `PipelineContext` was ephemeral, subsystems carried input specific caches (oops), was often recreated, and the return was `CliExit`.

Now, `PipelineResult` carries all three purposes and a single instance is used for the life of the pipeline run (per call to `Execute`)

~~_This is currently in draft state because it will have merge conflicts with #2413, includes all the commits from #2413, and because it needs a cleanup/my review pass_~~ Edit: Now up to date with those PRs

### DIFF
--- a/src/System.CommandLine.Subsystems.Tests/AlternateSubsystems.cs
+++ b/src/System.CommandLine.Subsystems.Tests/AlternateSubsystems.cs
@@ -10,7 +10,7 @@ namespace System.CommandLine.Subsystems.Tests
     {
         internal class AlternateVersion : VersionSubsystem
         {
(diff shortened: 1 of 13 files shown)


mhutch (approved):
Changes look ok. I do wonder whether we might want the pipeline to return a result type that wraps the pipeline state object, is immutable and doesn't have all the members available to the subsystems. In that case the name `PipelineResult` might be better reserved for that wrapper type. But that would be easy to change anytime before we freeze the API, so let's merge the PR as-is.

PR #2431, opened 2024-05-27
Powderhouse removed CliExit and fluent style for pipeline execution

KathleenDollard:
* Removed CLIExit (now using PipelineResult)
  * PipelineResult is now returned from methods that create it and do other things.
* Renamed Handled to AlreadyHandled


mhutch (approved):

src/System.CommandLine.Subsystems/Subsystems/CliSubsystem.cs
```diff
@@ -66,9 +66,9 @@ protected internal bool TryGetAnnotation<TValue>(CliSymbol symbol, AnnotationId<
     /// Executes the behavior of the subsystem. For example, help would write information to the console.
     /// </summary>
     /// <param name="pipelineResult">The context contains data like the ParseResult, and allows setting of values like whether execution was handled and the CLI should terminate </param>
-    /// <returns>A CliExit object with information such as whether the CLI should terminate</returns>
-    protected internal virtual CliExit Execute(PipelineResult pipelineResult)
-        => CliExit.NotRun(pipelineResult.ParseResult);
+    /// <returns>A PipelineResult object with information such as whether the CLI should terminate</returns>
```
Keboo, line 69:
```suggestion
```

src/System.CommandLine.Subsystems/Subsystems/PipelineResult.cs
```diff
@@ -13,4 +15,14 @@ public class PipelineResult(ParseResult? parseResult, string rawInput, Pipeline?
     public bool AlreadyHandled { get; set; }
     public int ExitCode { get; set; }
 
+    public void NotRun(ParseResult? parseResult)
+    {
+        // no op because defaults are false and 0
+    }
```
Keboo, line 21: If this is going to be a no-op should we simply remove the method as not calling it does the same thing?

KathleenDollard: Created an issue on this: #2443

Issue #2343, opened 2024-03-07
Apply help to all symbols in the CLI declaration


KathleenDollard:
We are currently planning to remove global options. We think they are rarely used and add complexity to the parser. 

Initialize for the subsystems gives us a great place to put the help option on every symbol in the tree. Altering HelpSubsystem.Initialize to add it across the tree, not just to the root. Also, tests are needed.

Note: This assumes that no other subsystem will add a command that will need help. If there is an order dependency here, we assume the CLI author will be aware of it and can use custom ordering by overriding Pipeline and updating InitializeSubsystems.

KathleenDollard:
The plan is a utility method that would add the option to all of your tree.

Would there be issues with that approach for you? Definitely a good example of the need to have an option on all commands of a CLI.

PR #2442, opened 2024-06-13
Apply help to all symbols in the CLI declaration

hasnik:
This PR aims to address https://github.com/dotnet/command-line-api/issues/2343 by adding help options to alll comands in the tree and includes tests.

Would love any thoughts or feedback as this is my first open-source PR.

### DIFF
--- a/src/System.CommandLine.Subsystems.Tests/AlternateSubsystems.cs
+++ b/src/System.CommandLine.Subsystems.Tests/AlternateSubsystems.cs
@@ -17,6 +17,15 @@ namespace System.CommandLine.Subsystems.Tests
             }
         }
 
+        internal class AlternateHelp : HelpSubsystem
+        {
+            protected override void Execute(PipelineResult pipelineResult)
+            {
+                pipelineResult.ConsoleHack.WriteLine($"***Help me!***");
+                pipelineResult.SetSuccess();
+            }
+        }
+
         internal class VersionThatUsesHelpData : VersionSubsystem
         {
             // for testing, this class accepts a symbol and accesses its description
--- a/src/System.CommandLine.Subsystems.Tests/HelpSubsystemTests.cs
+++ b/src/System.CommandLine.Subsystems.Tests/HelpSubsystemTests.cs
@@ -1,0 +1,89 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using FluentAssertions;
+using Xunit;
+using System.CommandLine.Parsing;
+
+namespace System.CommandLine.Subsystems.Tests
+{
+    public class HelpSubsystemTests
+    {
+        [Fact]
+        public void When_help_subsystem_is_used_the_help_option_is_added_to_each_command_in_the_tree()
+        {
+            var rootCommand = new CliRootCommand
+                {
+                    new CliOption<bool>("-x"), // add option that is expected for the test data used here
+                    new CliCommand("a")
+                };
+            var configuration = new CliConfiguration(rootCommand);
+
+            var pipeline = Pipeline.CreateEmpty();
+            pipeline.Help = new HelpSubsystem();
+
+            // Parse is used because directly calling Initialize would be unusual
+            var result = pipeline.Parse(configuration, "");
+
+            rootCommand.Options.Should().NotBeNull();
+            rootCommand.Options
+                .Count(x => x.Name == "--help")
+                .Should()
+                .Be(1);
+            var subcommand = rootCommand.Subcommands.First();
+            subcommand.Options.Should().NotBeNull();
+            subcommand.Options
+                .Count(x => x.Name == "--help")
+                .Should()
+                .Be(1);
+        }
+
+        [Theory]
+        [ClassData(typeof(TestData.Help))]
+        public void Help_is_activated_only_when_requested(string input, bool result)
+        {
+            CliRootCommand rootCommand = [new CliOption<bool>("-x")]; // add random option as empty CLIs are rare
+            var configuration = new CliConfiguration(rootCommand);
+            var helpSubsystem = new HelpSubsystem();
+            var args = CliParser.SplitCommandLine(input).ToList().AsReadOnly();
+
+            Subsystem.Initialize(helpSubsystem, configuration, args);
+
+            var parseResult = CliParser.Parse(rootCommand, input, configuration);
+            var isActive = Subsystem.GetIsActivated(helpSubsystem, parseResult);
+
+            isActive.Should().Be(result);
+        }
+
+        [Fact]
+        public void Outputs_help_message()
+        {
+            var consoleHack = new ConsoleHack().RedirectToBuffer(true);
+            var helpSubsystem = new HelpSubsystem();
+            Subsystem.Execute(helpSubsystem, new PipelineResult(null, "", null, consoleHack));
+            consoleHack.GetBuffer().Trim().Should().Be("Help me!");
+        }
+
+        [Fact]
+        public void Custom_help_subsystem_can_be_used()
+        {
+            var consoleHack = new ConsoleHack().RedirectToBuffer(true);
+            var pipeline = Pipeline.CreateEmpty();
+            pipeline.Help = new AlternateSubsystems.AlternateHelp();
+
+            pipeline.Execute(new CliConfiguration(new CliRootCommand()), "-h", consoleHack);
+            consoleHack.GetBuffer().Trim().Should().Be($"***Help me!***");
+        }
+
+        [Fact]
+        public void Custom_help_subsystem_can_replace_standard()
+        {
+            var consoleHack = new ConsoleHack().RedirectToBuffer(true);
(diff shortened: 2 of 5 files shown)


hasnik:
@dotnet-policy-service agree

src/System.CommandLine.Subsystems/HelpSubsystem.cs
```diff
@@ -24,8 +24,20 @@ public class HelpSubsystem(IAnnotationProvider? annotationProvider = null)
         Arity = ArgumentArity.Zero
     };
 
-    protected internal override void Initialize(InitializationContext context) 
-        => context.Configuration.RootCommand.Add(HelpOption);
+    protected internal override void Initialize(InitializationContext context)
+    {
+        AddOptionRecursively(context.Configuration.RootCommand, HelpOption);
+
+        static void AddOptionRecursively(CliCommand command, CliOption option)
```
hasnik, line 31: I imagine this method should belong to CliCommand or some extensions method, but I didn't want to change too much in this PR without consulting you first.

src/System.CommandLine.Subsystems/HelpSubsystem.cs
```diff
@@ -24,8 +24,20 @@ public class HelpSubsystem(IAnnotationProvider? annotationProvider = null)
...
+        {
+            command.Add(option);
+
+            foreach (var subcommand in command.Subcommands)
+            {
+                AddOptionRecursively(subcommand, option);
+            }
+        }
```
hasnik, line 39: I imagine this method should belong to CliCommand or some extensions method, but I didn't want to change too much in this PR without consulting you first.

src/System.CommandLine.Subsystems.Tests/HelpSubsystemTests.cs
```diff
@@ -0,0 +1,89 @@
...
+
+        [Fact]
+        public void Custom_help_subsystem_can_be_used()
+        {
+            var consoleHack = new ConsoleHack().RedirectToBuffer(true);
+            var pipeline = Pipeline.CreateEmpty();
+            pipeline.Help = new AlternateSubsystems.AlternateHelp();
+
+            pipeline.Execute(new CliConfiguration(new CliRootCommand()), "-h", consoleHack);
+            consoleHack.GetBuffer().Trim().Should().Be($"***Help me!***");
```
JeanJoeris, line 75: I might be misssing something but it seems like `Custom_help_subsystem_can_be_used` and `Custom_help_subsystem_can _replace_standard` are the same test. If so I think we might want to consolidate them

src/System.CommandLine.Subsystems.Tests/TestData.cs
```diff
@@ -93,4 +93,23 @@ internal class Value : IEnumerable<object[]>
 
         IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
     }
+
+    internal class Help : IEnumerable<object[]>
+    {
+        // This data only works if the CLI has a --version with a -v alias and also has a -x option
```
KathleenDollard, line 99: I do not understand this comment.

src/System.CommandLine.Subsystems/HelpSubsystem.cs
KathleenDollard: I am torn between having this be an extension method on CliOption or CliCommand. Regardless, I think it belongs in the core (System.CommandLine) and appreciate you being conservative on this. Let's get some more thoughts on the location. 

I think we should do a check so that if it is added twice we don't have problems. I think we should do that by name. We are in process of adding a `GetSymbolName` method, although I need to check that it is available on a `CliCommand`.

We will need to decide if first or last should win. My first intuition is to have the first win, so calling this method is "add the option to any commands that do not already have an option by this name". If you wanted custom help on one option (or another option), I think it would be most natural to add that first. Similarly, if you wanted one branch of commands in a complex tree to do help differently (`dotnet new` for example), you'd just have to call in order most specialized to least.

The other thing nagging on this is that our previous approach is that the CLI tree needs to be created before this is called. Another approach (although I like it less) is to mark the command and add the option to each command immediately before parsing. I am concerned about bad behavior if the tree is reused and a different help system is set (unless we copy the tree).

Sorry for not having a clear answer today on this. It's summer and we have some folks out, so it may be a bit before we have a clear answer on where and how to do the check.

src/System.CommandLine.Subsystems.Tests/HelpSubsystemTests.cs
```diff
@@ -0,0 +1,89 @@
...
+            var configuration = new CliConfiguration(rootCommand);
+
+            var pipeline = Pipeline.CreateEmpty();
+            pipeline.Help = new HelpSubsystem();
+
+            // Parse is used because directly calling Initialize would be unusual
+            var result = pipeline.Parse(configuration, "");
+
+            rootCommand.Options.Should().NotBeNull();
+            rootCommand.Options
```
KathleenDollard, line 29: We have a `SymbolByName` method in flight and need to ensure that it will work in this context.

KathleenDollard (commented):
Welcome! Thanks for this and including great tests!

In the long comment, this opens a couple of design questions and summer absences may result in a delay, but we should definitely be able to take this with tweaks.

src/System.CommandLine.Subsystems.Tests/TestData.cs
JeanJoeris: I believe this is copied from prior test data in this file, e.g. https://github.com/dotnet/command-line-api/blob/240817350b55a9734c0e3d46d1465777c03c8c5b/src/System.CommandLine.Subsystems.Tests/TestData.cs#L17-L23

src/System.CommandLine.Subsystems/HelpSubsystem.cs
hasnik: I'm happy to let it wait for the right time and update the code whenever the questions you pointed out have answers. Thank you sharing all the context, I do appreciate it very much!

> We will need to decide if first or last should win. My first intuition is to have the first win, so calling this method is "add the option to any commands that do not already have an option by this name". If you wanted custom help on one option (or another option), I think it would be most natural to add that first. Similarly, if you wanted one branch of commands in a complex tree to do help differently (dotnet new for example), you'd just have to call in order most specialized to least.

Haven't really thought about it before, but I'd say your first intuition makes a lot of sense to me.

> The other thing nagging on this is that our previous approach is that the CLI tree needs to be created before this is called. Another approach (although I like it less) is to mark the command and add the option to each command immediately before parsing. I am concerned about bad behavior if the tree is reused and a different help system is set (unless we copy the tree).

I believe there's much smarter individuals to discuss this with, but my intuition would be to stick with 'frozen' CLI tree - it just sounds right, unless there's any good reason not to use it. Is there any appropriate plugging point to add help 'higher' in the chain? Whatever the decision will be, do let me know - I'd love to have my small addition in the .NET ecosystem :)

src/System.CommandLine.Subsystems.Tests/HelpSubsystemTests.cs
hasnik: I've copied these shamelessly from [VersionSubsystemTests.cs](https://github.com/dotnet/command-line-api/blob/main-powderhouse/src/System.CommandLine.Subsystems.Tests/VersionSubsystemTests.cs#L96)

Couldn't really tell why there are two tests with identical body there, but I assumed I don't understand something and chose to add it. Removing it in the upcoming commit. Should I also remove the duplicate tests from `VersionSubsystemTests.cs` - specifically `Custom_version_subsystem_can_replace_standard`?

src/System.CommandLine.Subsystems.Tests/TestData.cs
hasnik: @JeanJoeris is exactly right. Removing this comment in the upcoming commit.

src/System.CommandLine.Subsystems.Tests/HelpSubsystemTests.cs
hasnik: Didn't knew about this one, will add it in the upcoming commit.

src/System.CommandLine.Subsystems.Tests/HelpSubsystemTests.cs
Keboo, line 75: nit:
```suggestion
            consoleHack.GetBuffer().Trim().Should().Be("***Help me!***");
```

src/System.CommandLine.Subsystems.Tests/AlternateSubsystems.cs
```diff
@@ -17,6 +17,15 @@ protected override void Execute(PipelineResult pipelineResult)
             }
         }
 
+        internal class AlternateHelp : HelpSubsystem
+        {
+            protected override void Execute(PipelineResult pipelineResult)
+            {
+                pipelineResult.ConsoleHack.WriteLine($"***Help me!***");
```
Keboo, line 24: nit:
```suggestion
                pipelineResult.ConsoleHack.WriteLine("***Help me!***");
```

PR #2446, opened 2024-06-21
Update ApprovalTests dependency

adamsitnik:
Required to fix the warnings we get from security tooling regarding https://msrc.microsoft.com/update-guide/vulnerability/CVE-2024-0056

Please keep in mind that the change affects only the tests projects, the product was never affected.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -8,6 +8,7 @@
     <NoWarn>$(NoWarn);NU5125;CS0618</NoWarn>
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
     <LangVersion>10.0</LangVersion>
+    <NetFrameworkMinimum>net472</NetFrameworkMinimum>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(DotNetBuildFromSource)' == 'true' or '$(DotNetBuildSourceOnly)' == 'true'">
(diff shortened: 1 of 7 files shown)


Directory.Build.props
```diff
@@ -8,6 +8,7 @@
     <NoWarn>$(NoWarn);NU5125;CS0618</NoWarn>
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
     <LangVersion>10.0</LangVersion>
+    <NetFrameworkMinimum>net472</NetFrameworkMinimum>
```
ViktorHofer, line 11: It's better to not overwrite a property that is defined by the Arcade Sdk for consistency: https://github.com/dotnet/arcade/blob/00d6decc59f5030c2399a64fd3e4f6e8e11bacca/src/Microsoft.DotNet.Arcade.Sdk/tools/TargetFrameworkDefaults.props#L29

ViktorHofer (commented):
IMO tests should either target the version that the libraries they test target (net462) or the latest version (net481). I think latest makes most sense here. Arcade defines a property for that: `NetFrameworkCurrent`.

adamsitnik:
> IMO tests should either target the version that the libraries they test target (net462) or the latest version (net481). I think latest makes sense here. Arcade defines a property for that: NetFrameworkCurrent.

I've updated the projects to use `NetFrameworkCurrent`. One thing that I don't like about it is that it's pure magic for contributors who are not familiar with Arcade. But assuming the contributors have 4.8.1 SDK installed, it should not be a problem.

ViktorHofer:
Yeah I get your point. I think NetFrameworkCurrent is mostly self-descriptive. It means the current version of .NET Framework which 4.8.1.

ViktorHofer (approved):

Issue #2390, opened 2024-04-21
How to Prevent `IHost` from Starting on Command-Line


cschuchardt88:
How to do you prevent `IHost` from starting when executing a command on the command-line? It starts no matter what when using `UseHost`. What's the best way to allow all `Command` class `ICommandHandler` interfaces to execute without starting the middleware that runs `IHostBuilder.StartAsync`? But still have all the `IHost` objects; example `IConfiguration` available `DI` in the `ICommandHandler` interface?

Running as:
```bash
> myApp.exe run
```

`Program.cs`
---
```csharp
public sealed partial class Program
{
    static async Task<int> Main(string[] args)
    {
        var rootCommand = new DefaultRootCommand();
        var parser = new CommandLineBuilder(rootCommand)
            .UseHost(DefaultNeoHostBuilderFactory, builder =>
            {
                    builder.ConfigureServices((builder, services) =>
                    {
                        services.AddSingleton<PromptSystemHostedService>();
                    });
                builder.UseCommandHandler<DefaultRootCommand, DefaultRootCommand.Handler>();
            })
            .UseDefaults()
            .Build();

        return await parser.InvokeAsync(args);
    }
}
```
`DefaultRootCommand.cs`
---
```csharp
internal sealed class DefaultRootCommand: Command
{
    public RunCommand() : base("run")
    {
        IsHidden = true;
    }

    public new sealed class Handler : ICommandHandler
    {
        public Handler(IConfiguration config) // want this from IHost
        {
        }

        public async Task<int> InvokeAsync(InvocationContext context)
        {
            var host = context.GetHost();
            // What code do I write to prevent 'host' from StartAsync

            return await RunConsolePrompt(ref context);
        }

        public int Invoke(InvocationContext context)
        {
            throw new NotImplementedException();
        }

        public static Task<int> RunConsolePrompt(ref InvocationContext context)
        {
            while (true)
            {
                lock (context.Console.Out)
                {
                    context.Console.Clear();
                    context.Console.SetTerminalForegroundColor(ConsoleColor.DarkBlue);

                    context.Console.Write("Prompt> ");
                    var line = context.Console.ReadLine();

                    context.Console.ResetTerminalForegroundColor();

                    return await PromptCommandsInvokeAsync(line);
                }
            }
        }
    }
}
```

`Console Output`
---
```bash
Prompt>
[...]

PR #2450, opened 2024-07-07
Refactor Hosting extensions

fredrikhr:
- [x] Remove `InvocationLifetime`
  Originally, `System.CommandLine.Hosting` was designed to depend only on `Microsoft.Extensions.Hosting.Abstractions` which made the `InvocationLifetime` class necessary. Currently however, the library depends on the full `Microsoft.Extensions.Hosting` library, so `InvocationLifetime` is no longer required as the functionality of the `ConsoleLifetime` class fully (and more extensively) covers its functionality.
  We should however provide guidance that you want set the process termination timeout of the `CliConfiguration` to `null` when using `System.CommandLine.Hosting` (the `UseHost` extension method does that as part of its implementation).
- [x] Introduce `HostedServiceAction`, a `CliAction` that using a .NET Generic Host hosted service as the execution for a CLI action.
  - Originally, one of the issues with using a Hosted Service as the Action for a CLI Command was that the .NET Generic Host does not automatically shutdown when the hosted service completes its execution task (that is by design, as Hosted services generally run in the background for the entire liftetime of the application).
    `HostedServiceAction` will automatically shut down the .NET Generic Host when the Hosted Service completes its work, i.e. the command finishes.
  - Add `CliHostedService`, a simple derivison from `BackgroundService`, which provides a `CliAsyncAction`-similar abstract `InvokeAsync` method to override which also allows returning back an integer result value. `HostedServiceAction` will use the return value from `CliHostedService` to return the result integer for the invocation.
    This class also avoids confusion regarding the name `BackgroundService` as the main handler for a CLI Command.
- [x] Rename `HostingAction` to `HostingWrapperAction`
  - Create a new base class `HostingAction` and make both `HostingWrapperAction` and `HostedServiceAction` derive from it.
  - Move the logic of creating the Host and starting it part of the base class, the `HostingWrapperAction` simply wraps the underlying action, while `HostedServiceAction` simply deals with waiting for the hosted service to complete.
- [x] Prepare extension points for using the modern generic host builder introduced in .NET 8.
  .NET 8 introduces the `HostApplicationBuilder` class as an alternative to the `HostBuilder` and `IHostBuilder` interfaces. Annoyingly, the `HostApplicationBuilder` does not implement the `IHostBuilder` interface necessary for the original implementation of the `System.CommandLine.Hosting` extensions.
  - Add an internal very limited wrapper around `HostApplicationBuilder` that (partially) implements `IHostBuilder`.
  - Add a generic type argument to the `HostingAction` base class that supports multiple different Host builders.
  - **Adding this functionality will require a new TFM for the `System.CommandLine.Hosting` library.** Current proposed implementation uses `#ifdef` set up for multi-targeting.
- [x] Add a publically settable `ConfigureHost` property on the `HostingAction` base class.
  - Since the new `HostedServiceAction` is publically exposed and constructable is is useful to be able to allow for per-command separate ConfigureHost methods.
  - `UseHost` will now combine its `configureHost` argument with already existing `ConfigureHost` delegates that have  previously been set on `HostingActions` set for Cli commands in the CLI configuration.
- [ ] Provide more examples for a complex multi-command set-up in `HostedPlayground`
  - Show how multiple `HostedServiceActions` can have their individual command-specific `ConfigureHost` delegates with a shared whole app `ConfigureHost` delegate.
  - Show differences between using just a regular `CommandHandler` with a method with hosting vs. using a `CliHostedService`.
- Add tests for new introduced features in this PR
  - [ ] Tests for combining `ConfigureHost` from `UseHost` and `HostingAction`
  - [ ] Tests for using `HostApplicationBuilder`
  - [ ] Tests for `HostedServiceAction`


fredrikhr:
https://github.com/dotnet/command-line-api/issues/2390 shows that there might be a use-case for an `UseHost` extension method on `BindingHandler` that would allow to create per-command Host configurations instead of the current `UseHost` extension that is applied for the entire configuration.

PR #2456, opened 2024-07-17
Updated TFM

KathleenDollard:
.NET 7 is out of support and not on my machine, lets update!

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -15,7 +15,7 @@
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(DotNetBuildFromSource)' != 'true' and '$(DotNetBuildSourceOnly)' != 'true'">
-    <TargetFrameworkForNETSDK>net8</TargetFrameworkForNETSDK>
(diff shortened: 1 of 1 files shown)


mhutch:
Nit but although net8 works, IMO net8.0 is more idiomatic.

Directory.Build.props
```diff
@@ -15,7 +15,7 @@
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(DotNetBuildFromSource)' != 'true' and '$(DotNetBuildSourceOnly)' != 'true'">
-    <TargetFrameworkForNETSDK>net7.0</TargetFrameworkForNETSDK>
+    <TargetFrameworkForNETSDK>net8</TargetFrameworkForNETSDK>
```
BenjaminMichaelis, line 18:
```suggestion
    <TargetFrameworkForNETSDK>net8.0</TargetFrameworkForNETSDK>
```

mhutch (approved):

PR #2466, opened 2024-07-30
Validation subsystem

KathleenDollard:
EDIT: Adding some context

The validation subsystem needs to connect _things set on the symbol about its data_ with a validation action. Completions needs to do the same thing so this is a general problem.

### What shall we call _things set on the symbol about its data_?

Things set on a symbol are called annotations in System.CommandLine. 

There are a couple of things different about things set about the symbol's data - the data for a data symbol or a set of data for commands. The most essential difference is that things about data have no meaning without 'ParseResults'. Things about the symbol do have meaning.

The current name in the code is `Trait`. That is not intended as the real name, probably. I thought of `Condition` after I submitted this PR, and am not going to switch without discussion. There is one implementation approach, not the current one, that lets us generally call these annotations, and just use the specialized name as a base class not visible outside a few subsystem classes. We still might not call it `Trait`, but that name is used in this comment to clarify the current code.

### Lookup approach

Each validator is of a specific type.

We need to find the intersection of traits and validators.  We could do this by looking for each validator and see if there are any traits indicating it needs to validate. Another approach is to go through the available traits and see if there is a validator. We will probably decide that not finding a validator, at least for many traits, itself creates an error or exception. 

Consider a CLI with 100 symbols (a bit smaller than the .NET CLI), with a trait on half of the symbols and 20 available validators. Looking up based on validators is 2,000 dictionary lookups (20 * 100). Looking up on traits, is 50 lookups (.5*20). Also, we must iterate the traits on the symbol if we raise an error or exception if a validator is missing.

Missing validators concern me. The CLI author must independently create code to add the trait and adding the validator. This means they may make a mistake and I would like to avoid this.

For these two reasons, this implementation searches by symbol and trait and looks up the validator.

One thing that may need extra work to accommodate is "prime validators". These are validators that if they fail, no other validation errors should be reported. There are a couple of ways we might handle this. 

### Implementation

There is not currently a mechanism to retrieve all of the annotations for a symbol. Therefore, I added an annotation which is a `List<Trait>`. Yes, a list, in a dictionary entry in another dictionary entry. I think the two dictionaries is a very good way to handle the strongly typed weak table's performance characteristics. I think including a list that generally as 0-3 items is a questionable, even if optimized for the zero case.

This is the reasons this PR is marked as draft.  I'd like to resolve this, which I think naturally leads to resolving the name.

#### Desired implementation

I would like a method, perhaps `IEnumerable<object> AnnotationStorageExtensions.GetAnnotations(this CliSymbol symbol)`., that returns all the annotations for the object. Because this list will be small, we can iterate the list, LINQ or directly, and find the things with the base class for traits. 

This will reduce the number of allocations as well as at least one dictionary lookup and some other code when the trait is assigned to a symbol. That will happen on all code paths, so optimization makes sense. 

### Why annotation ids are still important 

Subtitle: Another problem with the implementation in this draft PR

We want users to be able to reset aspects of symbols. They can do this on most things on symbol, and in `main` and in a future version using extension types, these are properties so clearly indicate that they are overwritten. This would be messy to fix in the implementation in this draft PR.

### Consider Range

There should be just one range. When multiple traits are added, they are all used for validation. This means that two ranges that match is extra work (we will probably have a `Distinct` somewhere to avoid duplicate reporting) and two ranges that do not overlap means no data will ever be valid. This is unusual.

The user may desire multiple validators. For example, there may be a validator we supply for simple validations. It might even include ranges that are relative to another data symbol's value or the current date. But it is not going to include validating (or tab completing) relative to the company's country specific calendar or workdays. It is not only custom, but it also needs an extra piece of data (the country). We can make it easy to create that validator and include it in the list of validators.

But how do we find it? The current plan is that it would have a separate type (for the extra data) and would be selected because it is the validator for that trait type and symbol type. And it could have a separate extension method. All cool.

But, what if the user inadvertently enters:

```csharp
var opt = new Option<Date>("projectedShipBy");
opt.SetRange(DateTime.Today, 6);
opt.SetRange(DateTime.Today, 6, "US");
```

If the extension method only sets the trait type, we may call two validators. Of course, the trait type and the extension methods could manage this and always use the new type. That is problematic for two reasons: first it's just extra messy code. Second, it undermines the "let me try a new subsystem and see what happens" capability (or at least it does in my head).

I think it is simplest if for basic traits the annotationId is used to ensure there is exactly one of the thing, and if there should be more than one (such as an "or"), it is an aggregate trait. Thus the annotationId for the range case would be "Range" or "RangeDataCondition" or some other key that would ensure we have only one.

### Returning to the name for _things set on the symbol about its data_

`Trait` is one of those "it's a name but means nothing" names, and thus cringy or icky. I think things like range and whether a file exists are conditions on the data of the symbol. Someone might think for a minute that it conditionally determines if a symbol should be included, but a quick check that it is not a bool and not a `Func` should get the user back on track. We could use `DataCondition` to alleviate that. `Characteristic` is too long and hard to spell. `Char` is taken. Open to ideas here. 

If we do the desired implementation, the only place this will appear is as the base class for the classes that hold the data for the things (upper and lower bound on range).


KathleenDollard:
Mikayla and I touched base on this and an updated PR is coming.

PR #2468, opened 2024-08-10
Remove type from annotation ID

mhutch:
The purpose of this was to ensure that accessors used the correct type for the annotation values. This originally applied to the `AnnotationAccessor<T>` helper for `symbol.With(help.Description, desc)` but also applied to later manually defined accessors.

This strong typing was not perfect, for example it was possible to (incorrectly) define two annotations with the same ID and different types, resulting in uninformative cast exceptions.

It also did not apply cleanly to more advanced cases, such as when the value was a `Func<T>` or where the annotation value type was based on the symbol value type. It is also not unreasonable to want to store more than one type in an annotation for other reasons. All these required removing the typing by using `AnnotationId<object?>`.

This makes things a bit cleaner by removing the strong typing from the annotation ID and instead adding `TryGetAnnotation` overloads that take a type and throw an informative exception on mismatch.

This only affects folks defining new annotations. It does not change anything for authors of CLIs or authors of subsystems, as they should to use the strongly typed accessors that should be defined along with every annotation ID.

### DIFF
--- a/src/System.CommandLine.Subsystems/Subsystems/Annotations/ValueAnnotations.cs
+++ b/src/System.CommandLine.Subsystems/Subsystems/Annotations/ValueAnnotations.cs
@@ -14,18 +14,20 @@ public static class ValueAnnotations
     /// Default value for an option or argument
     /// </summary>
     /// <remarks>
-    /// Although the type is <see cref="object?"/>, it must actually be the same type as the type
-    /// parameter of the <see cref="CliArgument{T}"/> or <see cref="CliOption{T}"/>.
+    /// Must be actually the same type as the type parameter of
+    /// the <see cref="CliArgument{T}"/> or <see cref="CliOption{T}"/>.
     /// </remarks>
-    public static AnnotationId<object?> DefaultValue { get; } = new(Prefix, nameof(DefaultValue));
+    public static AnnotationId DefaultValue { get; } = new(Prefix, nameof(DefaultValue));
 
     /// <summary>
     /// Default value calculation for an option or argument
     /// </summary>
     /// <remarks>
-    /// Although the type is <see cref="object?"/>, it must actually be a <see cref="Func{TResult}">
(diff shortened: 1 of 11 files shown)


src/System.CommandLine.Subsystems/Subsystems/Annotations/ValueAnnotations.cs
```diff
@@ -14,18 +14,17 @@ public static class ValueAnnotations
...
 
     /// <summary>
     /// Default value calculation for an option or argument
     /// </summary>
     /// <remarks>
-    /// Although the type is <see cref="object?"/>, it must actually be a <see cref="Func{TResult}">
-    /// with a type parameter matching the the type parameter type of the <see cref="CliArgument{T}"/>
-    /// or <see cref="CliOption{T}"/>
+    /// Must be return a <see cref="Func{TValue}"> with the same type parameter as
+    /// the <see cref="CliArgument{T}"/> or <see cref="CliOption{T}"/>.
```
KathleenDollard, line 27: Would it be better to have a "Please use the extension methods and not call this directly" type comment here? Same on line 17.

KathleenDollard (approved):

src/System.CommandLine.Subsystems/Subsystems/Annotations/ValueAnnotations.cs
mhutch: It's probably worth doing it some point, but we'd have to add it to all the annotation ID definitions, not just `ValueAnnotations`. All the methods that let you actually do anything with an `AnnotationId` already have a remark to that effect, so it'd be more of a discoverability thing than correctness thing.

KathleenDollard: That sounds fine. Let's merge and I'll rebase the next PR :smile:

PR #2470, opened 2024-08-10
Add error tracking in PipelineResult

KathleenDollard:
This lays groundwork for Validation and other error reporting.

Will be updated to CliDiagnostic when it is avialable

### DIFF
--- a/src/System.CommandLine.Subsystems/PipelineResult.cs
+++ b/src/System.CommandLine.Subsystems/PipelineResult.cs
@@ -1,10 +1,13 @@
 ﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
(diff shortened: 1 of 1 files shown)


src/System.CommandLine.Subsystems/PipelineResult.cs
```diff
@@ -13,6 +16,22 @@ public class PipelineResult(ParseResult? parseResult, string rawInput, Pipeline?
...
+        if (errors is not null)
+        {
+            this.errors.AddRange(errors);
+        }
+    }
+
+    public void AddError(ParseError error)
+        => errors.Add(error);
+
+    public IEnumerable<ParseError> GetErrors(bool excludeParseErrors = false)
```
mhutch, line 30: Returning `ParseError` but an argument called `excludeParseErrors` seems odd - are we going to rename `ParseError`?

mhutch (approved):

PR #2471, opened 2024-08-10
Added CliDataSymbol

KathleenDollard:
This acts as an interim base between `CliOption` and `CliArgument` and will allow simplifying a large number of places where they had to be separately handled, particularly in the subsystem layer.


mhutch (approved):
FWIW I would have liked to see the code that uses this change rather than just the change in isolation. I'm also curious how this affects options with multiple argument?

PR #2478, opened 2024-08-19
Update polyfills

mhutch:


src/Common/Compat/CodeAnalysis/NullabilityAttributes.cs
```diff
@@ -0,0 +1,149 @@
...
+		}
+
+		/// <summary>Gets the return value condition.</summary>
+		public bool ReturnValue { get; }
+
+		/// <summary>Gets field or property member names.</summary>
+		public string[] Members { get; }
+	}
+#endif
+}
```
KathleenDollard, line 149: Several files here lack a new line at the end of the file. No big deal, unless autoformatting adds it and results in future whitespace changes.

KathleenDollard (commented):
I like this change and will approve as soon as Kevin has looked at this or we figure out that he is not back yet - in which case he can catch up.  :-)

KathleenDollard (approved):

PR #2489, opened 2024-09-19
Centralize annotation resolution and allow multivalued annotations

mhutch:
* Replace per-subsystem annotation providers with an annotation resolver. This simplifies things and means that subsystem authors have one centralized place to get annotation values, instead of having to know which subsystem to use.
* Pass additional context to annotation providers, to allow more powerful dynamic resolution of annotation values
* Annotations may be defined as having multiple values, in which case multiple values may be added to the internal storage on the symbol, and providers may return enumerations of values. New annotation
enumeration APIs allow enumerating the values from the symbol and the providers.
* Strongly type SetRange, add GetRange for symmetry


mhutch:
Note: I didn't touch ValueAnnotationExtensions.cs as it's mostly removed in #2486 and I expect that PR to go in before this one. I'll fix it up after #2486 merges.

KathleenDollard (approved):

PR #2524, opened 2025-03-01
Fix ArgumentConverted for nullable args

hoyosjs:
This fixes cases where conversion of tokens in options of nullable scalar-bindable values with Artities that are not {1,1} like `Option<bool?>` but conversion is still well known.

Alternatively, if this is the only case that's a `ZeroOrOne` arity that we care about, it might be faster to elide branches and do what the  `{1, 1}` branch does.

### DIFF
--- a/src/System.CommandLine.Tests/Binding/TypeConversionTests.cs
+++ b/src/System.CommandLine.Tests/Binding/TypeConversionTests.cs
@@ -270,6 +270,23 @@ namespace System.CommandLine.Tests.Binding
                 .Be(null);
         }
 
+        [Theory]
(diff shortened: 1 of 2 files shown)


hoyosjs:
@adamsitnik @jonsequitur - can you please review this? It's something that broke when we bumped diagnostics to more recent versions of commandline.

mikem8361 (approved):

jonsequitur:
Do you have a test for this?

jonsequitur (approved):

adamsitnik:
@hoyosjs big thanks for your contribution!

PR #2530, opened 2025-03-14
Add ValueType to Option

MiYanni:
## Summary

When using the exposed surface of S.CL to create a JSON document that represents the dotnet CLI (via [this issue](https://github.com/dotnet/sdk/issues/46345)), I realized there was no way to get the System.Type of an Option. I looked at the S.CL class hierarchy and realized that information existed in the composed Argument within Option. (I used reflection [here](https://github.com/MiYanni/sdk/commit/b5b98619d8d71fdbb3fb56dc35a1f99512fc0ef4#diff-0e0732e037481ba444a5015df3d0c74184716f314d8b0b3db2402ef41c2bf31cR152-R153) to get at that information)

What I've done is added a `ValueType` property to `Option` that is exposed in the same way it is for `Argument`. For Options, a majority of the data is stored in the composed Argument, so to keep that design, I simply expose `_argument.ValueType` within `Option`.

### DIFF
--- a/src/System.CommandLine.ApiCompatibility.Tests/ApiCompatibilityApprovalTests.System_CommandLine_api_is_not_changed.approved.txt
+++ b/src/System.CommandLine.ApiCompatibility.Tests/ApiCompatibilityApprovalTests.System_CommandLine_api_is_not_changed.approved.txt
@@ -1,4 +1,4 @@
-System.CommandLine
+﻿System.CommandLine
   public abstract class Argument : Symbol
     public ArgumentArity Arity { get; set; }
     public System.Collections.Generic.List<System.Func<System.CommandLine.Completions.CompletionContext,System.Collections.Generic.IEnumerable<System.CommandLine.Completions.CompletionItem>>> CompletionSources { get; }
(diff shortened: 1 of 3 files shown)


jonsequitur (approved):
It makes sense for this to be at parity with `Argument` which also exposes a public `ValueType` property. I also hit an issue with this API change relative to the published beta4.

MiYanni:
@KalleOlaviNiemitalo 

> This change looks fine as such but the justification is not quite accurate.

You can get at almost **anything** via reflection, so that's cheating. 😈 I meant that via `Option` (not `Option{T}`), I can't use the public surface to get at the underlying type. I'm iterating over the `Command`s and those expose `IList<Option> Options`. And that works fine for everything except the type that Option is.

The original reflection I did was this:
```C#
var internalArgument = typeof(CliOption).GetProperties(BindingFlags.Instance | BindingFlags.NonPublic).First(pi => pi.Name == "Argument").GetValue(option) as CliArgument;
writer.WriteString(nameof(internalArgument.ValueType), internalArgument.ValueType.FullName);
```

MiYanni:
@jonsequitur What change do I need to make to make this test pass? `System_CommandLine_api_is_not_changed` Is there some kind of allowlist for API changes?

jonsequitur:
The project uses [ApprovalTests](https://github.com/approvals/ApprovalTests.Net) for this. 

After running the tests, look for *.received.txt` files here and just overwrite the existing `*.approved.txt` file with the changes.

image

MiYanni:
@jonsequitur Thank you, sir. Ezpz.

PR #2531, opened 2025-03-15
Remove illegal string manipulation

GrabYourPitchforks:
Part of the "reduce unsafe" effort across .NET.

The command line parsing code previously manipulated string instances in-place, which is not a legal operation on any .NET runtime. Removing this logic and falling back to a standard allocation.

@adamsitnik, when you get a chance, please make a **reduce-unsafe** label similar to what runtime and winforms already have, and assign that label to this PR. We're using that label to track this effort across repos.

I have not perf-tested this change, but I don't expect it to be too bad, especially since all the other perf improvements Adam introduced several years ago will remain in place. If this change does end up unacceptably impacting perf, we can work together on a safer solution if needed.

### DIFF
--- a/src/System.CommandLine/Parsing/StringExtensions.cs
+++ b/src/System.CommandLine/Parsing/StringExtensions.cs
@@ -225,44 +225,38 @@ namespace System.CommandLine.Parsing
             {
                 int tokensBefore = tokenList.Count;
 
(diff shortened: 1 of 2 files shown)


GrabYourPitchforks:
Note to reviewers: I recommend using the "hide whitespace" checkbox in the PR review tab. The majority of changes were just reducing an indentation level. :)

PR #2543, opened 2025-04-14
Bump package versions from out-of-support 6.0 to 8.0

jeffhandley:


jeffhandley:
The CI failure is possibly related to this version jump. `System.CommandLine.Tests.Invocation.CancelOnProcessTerminationTests.CancellableHandler_is_cancelled_on_process_termination` is failing where a `GracefulExitCode` of `42` is expected to be returned when the operation is canceled, but instead the `SIGINT_EXIT_CODE` of `130` is returned.

PR #2547, opened 2025-04-22
Move from FluentAssertions to AwesomeAssertions

artl93:
Fixes #2545 

Moves from FluentAssertions 6.x to AwesomeAssertions 8.x. 
Addressed subtle differences in API.

PR #2548, opened 2025-04-22
Enable Apple Silicon for dotnet suggest and test app

artl93:
Adds Apple Silicon for compilation support and test apps. 

Was required to get build to pass on Apple Silicon when working on #2547.


adamsitnik (approved):
LGTM, thank you @artl93 !

PR #2556, opened 2025-05-02
Enable ControlFlowGuard on the NativeAOT test

jeffhandley:
Lots of whitespace differences in this edit, but the only real change is adding `<ControlFlowGuard>Guard</ControlFlowGuard>` to address a binskim diagnostic.

### DIFF
--- a/src/System.CommandLine.Tests/TestApps/NativeAOT/NativeAOT.csproj
+++ b/src/System.CommandLine.Tests/TestApps/NativeAOT/NativeAOT.csproj
@@ -1,21 +1,22 @@
 <Project Sdk="Microsoft.NET.Sdk">
-	
-	<PropertyGroup>
-		<OutputType>Exe</OutputType>
-		<TargetFramework>$(TargetFrameworkForNETSDK)</TargetFramework>
-		<!-- producing more detailed log output -->
-		<TrimmerSingleWarn>false</TrimmerSingleWarn>
-    <PublishAot>true</PublishAot>
-	</PropertyGroup>
 
-	<PropertyGroup>
-		<SystemCommandLineDllPath Condition="'$(SystemCommandLineDllPath)'==''">..\..\..\System.CommandLine\bin\Release\$(TargetFrameworkForNETSDK)\System.CommandLine.dll</SystemCommandLineDllPath>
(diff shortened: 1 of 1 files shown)


eerhardt:
> adding <ControlFlowGuard>Guard</ControlFlowGuard> to address a binskim diagnostic.

I didn't know we ran these on tests.

src/System.CommandLine.Tests/TestApps/NativeAOT/NativeAOT.csproj
```diff
@@ -1,21 +1,22 @@
...
-	<PropertyGroup>
-		<SystemCommandLineDllPath Condition="'$(SystemCommandLineDllPath)'==''">..\..\..\System.CommandLine\bin\Release\$(TargetFrameworkForNETSDK)\System.CommandLine.dll</SystemCommandLineDllPath>
-	</PropertyGroup>
+    <PropertyGroup>
+        <OutputType>Exe</OutputType>
+        <TargetFramework>$(TargetFrameworkForNETSDK)</TargetFramework>
+        <!-- producing more detailed log output -->
+        <TrimmerSingleWarn>false</TrimmerSingleWarn>
+        <PublishAot>true</PublishAot>
+        <ControlFlowGuard>Guard</ControlFlowGuard>
```
eerhardt, line 9: @agocke @MichalStrehovsky - is there a reason why this isn't enabled by default with `PublishAot`?

eerhardt (approved):

jeffhandley:
> I didn't know we ran these on tests.
It can be configured to include/exclude paths as desired, and this repo's config has it including the test outputs. I kind that it caught the issue here, because it prompted your question for why it's not the default.

src/System.CommandLine.Tests/TestApps/NativeAOT/NativeAOT.csproj
MichalStrehovsky: > @agocke @MichalStrehovsky - is there a reason why this isn't enabled by default with `PublishAot`?

We follow platform conventions - Control Flow Guard requires an explicit opt-in with VC++ too.

PR #2559, opened 2025-05-14
some ergonomic improvements

jonsequitur:
This restores a few of the fluent-friendly method signatures for better source code compatibility with beta4.


adamsitnik (approved):
LGTM, apologies for the delay in the review.

PR #2565, opened 2025-06-03
Add new SetAction overload to avoid fire & forget silent error

### LINKED ISSUE #2562: SetAction overloads should have Task options without CancellationToken
Folks might accidentally pass an async lambda that doesn't accept a cancellation token and the compiler will treat it as async void, which will in turn be treated as a synchronous action by System.Commandline.  Many places in our API we treat CancellationToken as optional, we should do so here.  Please add overloads that accept Func returning Task without a cancellation token.

adamsitnik:
fixes #2562

cc @ericstj

### DIFF
--- a/src/System.CommandLine.Tests/Invocation/InvocationTests.cs
+++ b/src/System.CommandLine.Tests/Invocation/InvocationTests.cs
@@ -188,7 +188,42 @@ namespace System.CommandLine.Tests.Invocation
 
             (await rootCommand.Parse("").InvokeAsync()).Should().Be(123);
         }
-        
+
+        [Fact]
+        public async Task Anonymous_async_action_is_not_mapped_into_sync_void_with_fire_and_forget()
+        {
+            RootCommand rootCommand = new();
+            CancellationTokenSource cts = new();
+            bool wasCancelled = false;
+
+            Task cancellableDelay = Task.Delay(TimeSpan.FromHours(1), cts.Token);
+
+            // We are not using the overload that takes a CancellationToken to ensure that
+            // we reproduce the behavior described in https://github.com/dotnet/command-line-api/issues/2562:
+            // having an "async void" action passed as Action<ParseResult> to SetAction
+            // and running into "fire and forget".
+            rootCommand.SetAction(async parseResult => 
+            {
+                try
+                {
+                    await cancellableDelay;
+                }
+                catch (TaskCanceledException)
+                {
+                    wasCancelled = true;
+                }
+                
+            });
+
+            Task started = rootCommand.Parse("").InvokeAsync();
+            started.IsCompleted.Should().BeFalse("It should return a Task that has not finished yet.");
+
+            cts.Cancel();
+            await started;
+            wasCancelled.Should().BeTrue("It should be faulted because the delay was cancelled.");
+            started.IsCompleted.Should().BeTrue("It should complete after the cancellation.");
+        }
+
         [Fact]
         public void Terminating_option_action_short_circuits_command_action()
         {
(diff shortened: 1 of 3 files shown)


src/System.CommandLine/Command.cs
```diff
@@ -149,6 +149,30 @@ public void SetAction(Func<ParseResult, CancellationToken, Task> action)
...
+        /// <summary>
+        /// Sets an asynchronous action to be run when the command is invoked.
+        /// </summary>
+        /// <remarks>
+        /// When possible, prefer using the <see cref="SetAction(Func{ParseResult, CancellationToken, Task})"/> overload
+        /// and passing the <see cref="CancellationToken"/> parameter to the async method(s) called by the action.
+        /// </remarks>
+        // Hide from intellisense, it's public to avoid the compiler choosing a sync overload
+        // for an async action (and fire and forget issue described in https://github.com/dotnet/command-line-api/issues/2562).
+        [EditorBrowsable(EditorBrowsableState.Never)]
```
ericstj, line 161: If you wanted to force folks to think about the `CancellationToken` you could also mark it obsolete -- not sure how hardcore you are about that.

adamsitnik: It was my first idea, but then I realized that we should most likely not release new public `[Obsolete]` methods. But I agree, it would be even better as it would allow us emit a compiler warning/error with a clear message.

@jonsequitur thoughts?

src/System.CommandLine.Tests/Invocation/InvocationTests.cs
```diff
@@ -188,7 +188,42 @@ public async Task Anonymous_RootCommand_int_returning_Action_can_set_custom_resu
...
+                }
+                catch (TaskCanceledException)
+                {
+                    wasCancelled = true;
+                }
+                
+            });
+
+            Task started = rootCommand.Parse("").InvokeAsync();
+            started.IsCompleted.Should().BeFalse("It should return a Task that has not finished yet.");
```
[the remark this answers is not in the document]
adamsitnik: I need to prove that it's fire and forget without the new overload. I've simplified the test to only have an action that is supposed to wait for an hour.  `Barrier` is sync, I am not able to implement a test that shows the bug without getting into deadlock when the bug reproduces.

@jonsequitur is the new simplified version OK?

src/System.CommandLine/Command.cs
adamsitnik: > Would it be simpler to differentiate the methods by name rather than using overloads, e.g. `SetAction` and `SetAsyncAction`?

I like the `SetAsyncAction` name idea (we never discussed actions at the API review), but users could still run into this particular bug and it would cause another wave of breaking changes in SDK/VMR etc. Let's discuss it offline, I think it's now or never.

KalleOlaviNiemitalo:
I feel this deserves a more general Roslyn analyzer, which would warn whenever an async lambda expression as an argument of a call is implicitly converted to a void-returning delegate type.  If the developer really wants an async void function, they can replace the lambda expression with a named local function so that the `void` is explicit.

The analyzer could abstain from warning if there is no overload that would accept some sort of task-type-returning delegate instead.  However, I suspect there wouldn't be too many warnings even without this check.

adamsitnik:
> I feel this deserves a more general Roslyn analyzer, which would warn whenever an async lambda expression as an argument of a call is implicitly converted to a void-returning delegate type. If the developer really wants an async void function, they can replace the lambda expression with a named local function so that the `void` is explicit.
> 
> The analyzer could abstain from warning if there is no overload that would accept some sort of task-type-returning delegate instead. However, I suspect there wouldn't be too many warnings even without this check.

It's great idea, would you like to open an issue in the analyzers repo?

KalleOlaviNiemitalo:
> It's great idea, would you like to open an issue in the analyzers repo?

I found some pre-existing feature requests for warning about async void:

- <https://github.com/dotnet/roslyn-analyzers/issues/1576> rejected; they don't want to add analysers for async.  The recommendation was to use analysers from vs-threading instead.
- <https://github.com/dotnet/roslyn-analyzers/issues/4944>, but it is scoped to only analyse custom analysers and thus would not apply to programs using System.CommandLine.

At this time then, it does not seem useful to file a new request there.  But I'll be looking into whether I can start using the vs-threading analysers.

PR #2566, opened 2025-06-05
Update to .NET 8.0

zbalkan:
Since .NET 7.0 EOL, it's better to stick to LTS version that is till supported. I used LTS instead of LTS to minimize friction between new versions.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -1,12 +1,10 @@
 <Project>
-
   <PropertyGroup>
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
     <CentralPackageTransitivePinningEnabled>false</CentralPackageTransitivePinningEnabled>
     <!-- Using multiple feeds isn't supported by Maestro: https://github.com/dotnet/arcade/issues/14155. -->
     <NoWarn>$(NoWarn);NU1507</NoWarn>
   </PropertyGroup>
-
   <ItemGroup>
     <!-- Roslyn dependencies -->
     <PackageVersion Include="Microsoft.CodeAnalysis" Version="4.0.1" />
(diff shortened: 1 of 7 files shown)


src/System.CommandLine/ArgumentValidation.cs
```diff
@@ -171,7 +171,7 @@ private static void FileOrDirectoryExists<T>(ArgumentResult result)
 
                 if (checkFile && checkDirectory)
                 {
-#if NET7_0_OR_GREATER
+#if NET8_0_OR_GREATER
```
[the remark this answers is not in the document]
zbalkan: https://github.com/dotnet/command-line-api/pull/2566/commits/ffea4e6a2a765706975677208d1c2c6f40cebb49

Directory.Packages.props
```diff
@@ -1,12 +1,10 @@
 <Project>
-
```
ViktorHofer, line 2: Revert the removal of these empty lines please. We intentionally put them between property / items groups and after the project start and before the project closing tag for readability.

ViktorHofer (commented):
@jeffhandley @adamsitnik I think it would be good to consider this change for the next possible nuget.org release.

Directory.Packages.props
zbalkan: This is done by the linter I guess. I'll fix it.

zbalkan:
5 commits for this small change may look a bit messy. Should I squash and force-push or do you prefer a merge with squash after it is approved?

jonsequitur:
Other than the package version change for System.Memory, these changes only affect tests. The System.CommandLine library already targets `net8.0` and `netstandard2.0`.

ViktorHofer:
Right but we shouldn't be depending on old / obsolete package dependencies, specifically System.Memory/4.5.4 and Newtonsoft.Json/13.0.2 in upcoming releases.

ViktorHofer (approved):

Frulfump:
From PR description
>LTS instead of LTS

?!
PR title is a bit confusing since the focus is the .NET version used in tests, and the other changes bump some packages not necessarily tied to .NET 8? And it show up as just that in the GitHub release description.

adamsitnik:
> From PR description
> 
> > LTS instead of LTS
> 
> ?! PR title is a bit confusing since the focus is the .NET version used in tests, and the other changes bump some packages not necessarily tied to .NET 8? And it show up as just that in the GitHub release description.

Good point, I've renamed it to "Update test projects to use "net8.0" as a moniker for testing "

Issue #2592, opened 2025-06-24
Option/Argument.GetRequiredValue throws when DefaultValueFactory is set


jonsequitur:
When an option is marked `Required` and a `DefaultValueFactory` is set, calling `Option.GetRequiredValue` still throws. 

Program code:

```csharp
using System.CommandLine;

var rootCommand = new RootCommand("This is the description")
{
    new Option<string>("--required") 
    {
        Required = true,
        DefaultValueFactory = _ => "hello"
    }
};

var parseResult = rootCommand.Parse("-h");

Console.WriteLine(parseResult.GetRequiredValue<string>("--required"));


```

Exception: 

```console
Unhandled exception. System.InvalidOperationException: --required is required but was not provided.
   at System.CommandLine.Parsing.SymbolResult.GetRequiredValue[T](String name)
   at System.CommandLine.ParseResult.GetRequiredValue[T](String name)
   at Program.<Main>$(String[] args) in C:\temp\command-name-issue\Program.cs:line 14
```

![Image](https://github.com/user-attachments/assets/6eb9886f-b03e-4bab-ac27-c3955c268560)

PR #2601, opened 2025-06-27
Fix some help and `GetRequiredValue` issues

### LINKED ISSUE #2573: GetRequiredValue throws for implicit values
Currently, `Argument.GetRequiredValue` and `Option.GetRequiredValue` will throw `InvalidOperationException` when the argument or option wasn't actually parsed from command line input. For example:

```csharp
var option = new Option<bool>("-x");

var command = new RootCommand
{
    option
};

// the following throws: -x is required but was not provided
var value = command.Parse("").GetRequiredValue(option);
```

If `Option.DefaultValueFactory` is set, then the exception is not thrown. 

I believe the design of `GetValue` and `GetRequiredValue` should focus on whether the value for the symbol can be bound, including implicit and default values, regardless of whether a matching token was found on the parsed commandline. 

This could be the wrong intuition though. For `bool` it seems intuitive, and perhaps for `int` and other numeric types that default to `0`, but for more complex structs such as `DateTime`, as well as for reference types, this kind of behavior seems nonsensical.

What are people's thoughts on this?

### LINKED ISSUE #2582: When displaying default values do not display [] when value is null
_v2.0.0-beta5.25306.1_

I have some default values that _can_ be sourced from environment variables if they're present.

For example,

```c#
        var passwordOption = new Option<string?>("--password". "-p", "/p")
        {
            Description = "The password to use.",
            DefaultValueFactory = defaultValue =>
            {
                if (string.IsNullOrWhiteSpace(Environment.GetEnvironmentVariable("_Password")))
                {
                    return null;
                }

                return Environment.GetEnvironmentVariable("_Password");
            }
        };

```

Currently the help text would display 

```
-p, --password          The password to use. []
```

It would, to my mind, be cleaner if, when the `DefaultValueFactory` returns null that nothing is displayed as the default, because in this case there is no default.

e.g.

```
-p, --password          The password to use.
```

jonsequitur:
This fixes the following issues:

* #2592
* #2582
* #2573

The approach I took to fix #2573 makes `Argument<bool>` and `Option<bool>` have an implicit `DefaultValueFactory`. This change drove changes to the `HelpBuilder` so I also fixed #2582 while I was in there.

It's not obvious to me that this is the right fix, so I thought that the PR would help clarify the discussion.

For example, should other types than `bool` have this behavior of an implicit default value?

### DIFF
--- a/src/System.CommandLine.Tests/OptionTests.cs
+++ b/src/System.CommandLine.Tests/OptionTests.cs
@@ -4,6 +4,7 @@
 using System.CommandLine.Parsing;
 using FluentAssertions;
 using System.Linq;
+using FluentAssertions.Execution;
 using Xunit;
 
 namespace System.CommandLine.Tests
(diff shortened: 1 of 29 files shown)


src/System.CommandLine/Argument{T}.cs
```diff
@@ -76,6 +93,11 @@ public Argument(string name) : base(name)
         {
             if (DefaultValueFactory is null)
             {
+                if (IsBoolean())
+                {
+                    return false;
+                }
```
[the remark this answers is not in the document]
jonsequitur: This is dead code in any case. The actual current behavior matches that expectation.

src/System.CommandLine.Tests/System.CommandLine.Tests.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>$(TargetFrameworkForNETSDK);$(NetFrameworkCurrent)</TargetFrameworks>
+    <TargetFrameworks>$(TargetFrameworkForNETSDK)</TargetFrameworks>
```
jonsequitur, line 4: @ViktorHofer I was getting the error `error CS0518: Predefined type 'System.Range' is not defined or imported` for usage of index and range syntax for `net481` within this project until I did this. I tried adding the classes directly as well as referencing Microsoft.Bcl.Memory and neither approach worked. 

Is this target framework needed for the tests? (It's not included for the main library.)

ViktorHofer: The issue is in ApprovalTests: https://github.com/approvals/ApprovalTests.Net/issues/768

adamsitnik (approved):
LGTM, thank you for fixing all the issues @jonsequitur !

src/System.CommandLine.Tests/System.CommandLine.Tests.csproj
ViktorHofer: We should definitely keep the .NET Framework tests. That validates that the library works correctly on modern .NET and .NET Framework.

jonsequitur: That would be great. Thanks!

src/System.CommandLine/Argument{T}.cs
```diff
@@ -27,7 +28,23 @@ public Argument(string name) : base(name)
...
+        public Func<ArgumentResult, T>? DefaultValueFactory
+        {
+            get
+            {
+                if (_defaultValueFactory is null)
+                {
+                    switch (this)
+                    {
+                        case Argument<bool> boolArgument:
+                            boolArgument.DefaultValueFactory = _ => false;
```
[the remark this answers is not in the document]
jonsequitur: Agreed, `if` is cleaner. I was considering whether other types should have similar treatment.

src/System.CommandLine.Tests/OptionTests.cs
```diff
@@ -405,6 +469,8 @@ public void Multiple_identifier_token_instances_without_argument_tokens_can_be_p
 
             var result = root.Parse("-v -v -v");
 
+            using var _ = new AssertionScope();
```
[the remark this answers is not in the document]
adamsitnik: > This changes the behavior of the assertions so that all will be evaluated and reported even if the first one fails. Multiple assertions in a test is often considered an antipattern because they can fail independently and failing the test on the first assertion failure loses useful information.
> 
> The `AssertionScope` approach provides a way to make sure all assertion results are reported.

Thanks a lot! It would be great if we could enable it project-wide without a code change (like a csproj setting)

src/System.CommandLine.Tests/System.CommandLine.Tests.csproj
ViktorHofer: Done

PR #2613, opened 2025-07-04
thoughts on implicit usings?

SimonCropp:


Frulfump:
My thoughts on that is captured in this issue https://github.com/dotnet/sdk/issues/49182 where the community at least in that issue so far is against it and Microsoft employees are for it so they will do it like the MS employees want...

PR #2616, opened 2025-07-05
move to Verify

SimonCropp:


jonsequitur:
It looks like the files documenting the API surface would no longer be part of the version controlled artifacts in the repo?

These are useful to have and a key reason why ApprovalTests was used here. Does Verify have an analogous approach?

SimonCropp:
@jonsequitur my mistake. that should now be fixed

PR #2624, opened 2025-07-08
make the non-generic Option and Argument ctors non-private

jonsequitur:
This allows these untyped `Option` and `Argument` classes to be inherited in code referencing System.CommandLine. This pattern was used by `HelpOption` and `VersionOption` in order to implement flag-style option behavior (with zero arity and no parseable return types). This same capability is useful beyond the scope of System.CommandLine's internals.


anowarulbb:
src/System.CommandLine/Option{T}.cs

anowarulbb:
src/System.CommandLine/Option{T}.cs

PR #2644, opened 2025-07-30
fix #2591

### LINKED ISSUE #2591: --version does not work when you configure an argument on the root command
I have a configuration like the following

```c#
var statusArgument = new Argument<string>("status")
{
    Description = "What you are doing right now?"
};

var allowRepliesOption = new Option<bool>("--allowReplies", "-r", "/r")
{
    Description = "Allow replies to your status update.",
    DefaultValueFactory = defaultValue => true
};

var rootCommand = new RootCommand("Post a quick status update on Bluesky")
{
    statusArgument,
    allowRepliesOption
};
```

etc.

`-h` works, but `--version` takes the argument into account, so I get

```
Required argument missing for command: 'bloop'.
```

Edited: 

More detailed repro steps below at https://github.com/dotnet/command-line-api/issues/2591#issuecomment-3001683393.

jonsequitur:
This fixes #2591.

The approach I've taken to this fix generalizes how certain actions interact with parse errors, adding the (currently internal) `CommandLineAction.IgnoresParseErrors` property and removing the special case for help.

### DIFF
--- a/src/System.CommandLine/Invocation/CommandLineAction.cs
+++ b/src/System.CommandLine/Invocation/CommandLineAction.cs
@@ -17,5 +17,9 @@ public abstract class CommandLineAction
     /// </summary>
     public bool Terminating { get; protected init; } = true;
 
+    /// <summary>
+    /// Indicates that the action clears any parse errors associated with symbols other than one that owns the <see cref="CommandLineAction"/>.
+    /// </summary>
+    /// <remarks>This property is ignored when <see cref="Terminating"/> is set to <see langword="false"/>.</remarks>
     public virtual bool ClearsParseErrors => false;
 }


KalleOlaviNiemitalo:
I have two programs that customize help by searching for the automatically-added HelpOption and assigning an instance of a custom action class to its Action property.

This PR looks like it would no longer suppress syntax errors when such customized help is invoked.

As HelpAction is a sealed class, I wouldn't be able to fix this by deriving my custom action class from HelpAction, either.

jonsequitur:
Maybe `HelpAction` should not be `sealed`. I could also imagine making `IgnoresParseErrors` public but would need to think through and test whether this introduces other potential issues.

KalleOlaviNiemitalo:
Unsealing HelpAction would allow custom help to ignores parse errors again, but only as a synchronous action, not asynchronous.

I wonder whether there is any reasonable scenario that would need an asynchronous action for custom `--help` or `--version`.  Perhaps in a command that downloads a plugin over the network (similar in concept to `dotnet tool exec` <https://github.com/dotnet/sdk/issues/31103>) and merges the plugin options with the host options (like `dotnet new` does to template options).  However the same network I/O would then be needed in command-line completion too, and Symbol.GetCompletions does not currently support async, so perhaps it is not a reasonable scenario.

For `--version`, a network check for available updates could perhaps make sense.

baronfel (approved):
I'm gonna approve this, but I'd ideally like @jeffhandley's thoughts as well.

I'm having a hard time finding a solution that:

* enables replacing the built-in HelpAction entirely (like we've been enabling for the past several changes around helpbuilder, etc)
* doesn't rely on type-testing a known set of types
* doesn't rely on subclassing HelpAction itself (because then folks wouldn't be able to subclass)
* _isn't_ some kind of publically-visible property on CommandLineAction

I think if we can add a test to pin the invariant that the value of this property on the _primary_ action (the one to be invoked) is the only one that changes the behavior of the pipeline, this is good to go.

PR #2647, opened 2025-07-31
Helper property to easily check if tokens are provided for a symbol

xIceFox:
I want to propose a clearer way to check if a SymbolResult has any provided tokens with this small pull request.

This should help in custom validators, and leads to better readability.

It's really only a convenience property, the name is just a suggestion, it could also be named "HasTokens" or something.

Example validator:

```
var option1Result = x.GetResult(_option1);
var option2Result = x.GetResult(_option2);

// null checks 
[..]

// before
if(option1Result.Tokens.Count > 0 && option2Result.Tokens.Count > 0)
{
    x.AddError($"{_option1.Name} and {_option2} can not be set together.");
}

// after
if(option1Result.IsProvided && option2Result.IsProvided)
{
    x.AddError($"{_option1.Name} and {_option2} can not be set together.");
}
```

### DIFF
--- a/src/System.CommandLine.ApiCompatibility.Tests/ApiCompatibilityApprovalTests.System_CommandLine_api_is_not_changed.approved.txt
+++ b/src/System.CommandLine.ApiCompatibility.Tests/ApiCompatibilityApprovalTests.System_CommandLine_api_is_not_changed.approved.txt
@@ -233,6 +233,7 @@ System.CommandLine.Parsing
     public System.String ToString()
   public abstract class SymbolResult
     public System.Collections.Generic.IEnumerable<ParseError> Errors { get; }
+    public System.Boolean IsProvided { get; }
     public SymbolResult Parent { get; }
     public System.Collections.Generic.IReadOnlyList<Token> Tokens { get; }
     public System.Void AddError(System.String errorMessage)
--- a/src/System.CommandLine.Tests/ParserTests.cs
+++ b/src/System.CommandLine.Tests/ParserTests.cs
@@ -1686,5 +1686,45 @@ namespace System.CommandLine.Tests
 
             result.GetValue(option).Should().BeEmpty();
         }
+        
+        [Fact]
+        public void Parsed_result_is_provided_should_be_true_for_provided_tokens()
+        {
+            const string optionName = "--test-option";
+
+            var option = new Option<string>(optionName);
+
+            var rootCommand = new RootCommand
+            {
+                option
+            };
+
+            var result = rootCommand.Parse([optionName, "test-value", optionName, "test-value-2"]);
+
+            var parseResult = result.GetResult(option);
+
+            parseResult.Should().NotBeNull();
+            parseResult!.IsProvided.Should().BeTrue();
+        }
+
+        [Fact]
+        public void Parsed_result_is_provided_should_be_false_for_not_provided_tokens()
+        {
+            const string optionName = "--test-option";
+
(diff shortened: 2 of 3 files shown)


baronfel:
@xIceFox are you trying to answer the question "did a user provide some input that led to this value being created?"? If so, I think the `Implicit` property on the `OptionResult` does what you want there?

xIceFox:
@baronfel Not really, I wrote a custom validator that checks if a user specified two specific options that can't be specified together.
Implicit is something else in this case. Because my options are explicitly provided by the user.
Like: If option1 provided, you cant provide option2. And the other way around.
Currently I check them with: OptionResult.Tokens.Count > 0 or OptionResult.Tokens.Any(). 
This convenience property is just an attempt to make the code more straightforward and readable. It just says, hey the user provided some input for this option.

baronfel:
That's part of what I'm getting at - your code works for options that require an argument, but it doesn't work for options that are just a flag - for example some `Option<bool> Verbose = new Option<bool>("--verbose") { Arity = ArgumentArity.Zero }`. Being able to cover this kind of option is why the [`Implicit`](https://github.com/dotnet/command-line-api/blob/main/src/System.CommandLine/Parsing/OptionResult.cs#L36) property exists.

xIceFox:
Yeah understood, what I initially overlooked before pulling out this pull request was the fact that there is no OptionResult in the ParseTree, when the user does not provide a specific parameter, is this assumption right?
In that case my real solution to my problem is just checking if both OptionResults are not null at the same time.

baronfel:
There _can_ be an OptionResult _if_ the `DefaultValueFactory` of the Option was triggered - this can happen if you explicitly create one, but S.CL will also create some DefaultValueFactories implicitly - for example, in the `Option<bool>` example above, if you _don't_ specify `--verbose`, the DefaultValueFactory for the option will run and give you a value of `false`.

So I think the full logic you'd want would be something like what we have in the [dotnet CLI](https://github.com/dotnet/sdk/blob/4284ea102618605f83abf8cb4bf30b6acf6d811c/src/Cli/dotnet/Extensions/ParseResultExtensions.cs#L285-L309):

```csharp
    /// <summary>
    /// Checks if the option is present and not implicit (i.e. not set by default).
    /// This is useful for checking if the user has explicitly set an option, as opposed to it being set by default.
    /// </summary>
    public static bool HasOption(this ParseResult parseResult, Option option) => parseResult.GetResult(option) is OptionResult or && !or.Implicit;

    /// <summary>
    /// Checks if the option with given name is present and not implicit (i.e. not set by default).
    /// This is useful for checking if the user has explicitly set an option, as opposed to it being set by default.
    /// </summary>
    public static bool HasOption(this ParseResult parseResult, string name)
        => parseResult.GetResult(name) is OptionResult or && !or.Implicit;

    /// <summary>
    /// Checks if the option is present and not implicit (i.e. not set by default).
    /// This is useful for checking if the user has explicitly set an option, as opposed to it being set by default.
    /// </summary>
    public static bool HasOption(this SymbolResult symbolResult, Option option) => symbolResult.GetResult(option) is OptionResult or && !or.Implicit;

    /// <summary>
    /// Checks if the option with given name is present and not implicit (i.e. not set by default).
    /// This is useful for checking if the user has explicitly set an option, as opposed to it being set by default.
    /// </summary>
    public static bool HasOption(this SymbolResult symbolResult, string name)
        => symbolResult.GetResult(name) is OptionResult or && !or.Implicit;
```

PR #2656, opened 2025-08-08
Fix #2128

### LINKED ISSUE #2128: --help no longer displays the option argument if Option.HelpName is not set
In System.CommandLine 2.0.0-beta4.22272.1 (commit 209b724a3c843253d3071e8348c353b297b0b8b5), if I set up an option like this

```csharp
            this.jsonDirectoryOption = new Option<DirectoryInfo>(
                name: "--json-directory",
                description: "Writes API JSON responses to the specified directory.");
            this.jsonDirectoryOption.ExistingOnly();
            this.Add(this.jsonDirectoryOption);
```

then `--help` generated a default description for the argument of the option:

```
  --json-directory <json-directory>   Writes API JSON responses to the specified directory.
```

In System.CommandLine 2.0.0-beta4.23165.2 (commit 4d1e6ac49a1fd0606862364aa5a2c6be2152ac16), if I similarly set up an option

```csharp
            this.jsonDirectoryOption = new Option<DirectoryInfo>("--json-directory")
            {
                Description = "Writes API JSON responses to the specified directory.",
            };
            this.jsonDirectoryOption.AcceptExistingOnly();
            this.Options.Add(this.jsonDirectoryOption);
```

then `--help` does not show that the option requires an argument:

```
  --json-directory                    Writes API JSON responses to the specified directory.
```

I can correct the `--help` output by setting Option.HelpName in my application, but this is quite a surprising change and should at least be documented in the upcoming release notes.

jonsequitur:
This fixes #2128. When `Option.HelpName` isn't supplied, the `Option.Name` (with prefixes removed) is used.


src/System.CommandLine/Help/HelpBuilder.Default.cs
```diff
@@ -75,28 +75,44 @@ public static string GetArgumentUsageLabel(Symbol parameter)
...
+                    return null;
+                }
 
-                    if (!string.IsNullOrEmpty(joined))
+                if (completionSources.Count <= 0)
+                {
+                    if (symbol is Option)
                     {
-                        return $"<{joined}>";
+                        return $"<{symbol.Name.Trim('-', '/')}>";
```
adamsitnik, line 101: nit: should we use remove both the leading and trailing chars?

```suggestion
                        return $"<{symbol.Name.TrimStart('-', '/')}>";
```

adamsitnik (approved):
Overall the change LGTM, as usual with great tests from @jonsequitur 👍 

I was about to mention that it will most likely break plenty of SDK tests (I remember I had to add `HelpName` here and there when updating), but then I've realized that the change was made in `HelpBuilder.Default.cs` and since the SDK has its own copy, it should not be affected.

Thanks!

PR #2657, opened 2025-08-08
Fix #2574 

### LINKED ISSUE #2574: add `Uri` to `ArgumentConverter.StringConverters`
Right now, if you have:
```cs
public class UrlOption : Option<Uri>
{
  public UrlOption() : base("--url")
  {
    Required = true;
  }
}
```

an `InvalidOperationException` is thrown, with the message:
```
Cannot parse argument 'http://example.com' for option '--url' as expected type 'System.Uri'.
```

jonsequitur:
This fixes #2574.

### DIFF
--- a/src/System.CommandLine.Tests/Binding/TypeConversionTests.cs
+++ b/src/System.CommandLine.Tests/Binding/TypeConversionTests.cs
@@ -28,13 +28,30 @@ namespace System.CommandLine.Tests.Binding
         }
 
         [Fact]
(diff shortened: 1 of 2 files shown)


KalleOlaviNiemitalo:
In one program, I'm specifically parsing with UriKind.Relative.  (It's a tool for replaying an HTTP request to a web API hosted in-process.)

But I suppose this PR won't make that kind of use any harder, as I can just keep using a custom parser like now.

Issue #2671, opened 2025-08-27
OperationCanceledException Not Thrown without Config in Async Apps


mpatkisson:
# OperationCanceledException Not Thrown without Config in Async Apps

[This documentation (published 02 AUG 25) on Learn](https://learn.microsoft.com/en-us/dotnet/standard/commandline/how-to-parse-and-invoke) implies that `System.CommandLine` will, without additional configuration, throw `OperationCanceledException` when `CTRL+C` is pressed allowing library consumers to gracefully handle any necessary cleanup or shutdown messaging.

Versions [2.0.0-beta5](https://github.com/dotnet/command-line-api/commits/v2.0.0-beta5.25306.1) and [2.0.0-beta6](https://github.com/dotnet/command-line-api/commits/v2.0.0-beta6.25358.103) both supported this behavior, [but 2.0.0-beta7](https://github.com/dotnet/command-line-api/commits/v2.0.0-beta7.25380.108) and [beyond](https://github.com/dotnet/command-line-api/commits/main) break it.

## Steps to reproduce

1. Clone and run <https://github.com/mpatkisson/culbertson> with .NET 9

   ```sh
   > git clone https://github.com/mpatkisson/culbertson.git
   > cd culbertson
   > dotnet run --project ./src/Culbertson
   ```

2. Once the project is running, press `CTRL+C`.

### Expected Behavior

[This routine](https://github.com/mpatkisson/culbertson/blob/c0571eaf82ba6b36738ac32ab952b5086af55c9b/src/Culbertson/Program.cs#L13) should be handled when `CTRL+C` is pressed, thus printing these message along with a slight delay for effect,

```sh
Termination requested. Cleaning up...
# 1 second delay
Cleanup complete. Exiting.
```

### Actual Behavior

The program _does_ exit when `CTRL+C` is pressed, but `OperationCanceledException` is never thrown, so cleanup messaging is _not_ displayed.

## Further Analysis

My guess is `OperationCanceledException` _should_ be thrown in Beta 7 as it is in Betas 5 and 6.  If this was a design decision, then apologies (but the docs / code commentary are confusing if so).

[...]

PR #2672, opened 2025-08-27
Re-add default value on ProcessTerminationTimeout (#2671)

mpatkisson:
This update re-introduces a default 2s value to `InvocationConfiguration.ProcessTerminationTimeout` to ensure that `OperationCanceledException` is thrown on `Ctrl+C` in apps using async commands. See detailed bug report #2671 for analysis and reproduction steps.


adamsitnik (approved):
Great catch! Big thanks for providing a detailed bug report and the fix @mpatkisson !

PR #2673, opened 2025-08-28
Backport bug fix to the 10.0-rc1 branch

adamsitnik:
let's include #2672 in the next release


adamsitnik:
According to a feedback shared offline with me, RC1 is already done.

PR #2680, opened 2025-09-10
fix #2213

### LINKED ISSUE #2213: `init` accessor of CliAction.Terminating cannot be called in C# 7
<https://github.com/dotnet/command-line-api/pull/2205> changed the CliAction.Terminating property from `{ get; protected set; }` to `{ get; protected init; }` here:

<https://github.com/dotnet/command-line-api/blob/104bb34e3a934cde1fd5b13086de441021def23a/src/System.CommandLine/Invocation/CliAction.cs#L15-L18>

The `init` accessor cannot be directly called in C# 7; it requires at least C# 9.  That makes the property difficult to initialize in these cases:

- When targeting .NET Framework, where [Microsoft does not support C# 8](https://github.com/dotnet/docs/blob/c802a0c56bccdc87908365c0788b3c5ecacea2b3/docs/csharp/language-reference/configure-language-version.md?plain=1#L12).
- When the application must remain buildable in Visual Studio 2017 ([whose support ends in April 2027](https://learn.microsoft.com/lifecycle/products/visual-studio-2017)).

<https://github.com/dotnet/command-line-api/issues/1586> was a similar issue about the `init` accessor of Option.Arity, fixed in <https://github.com/dotnet/command-line-api/pull/1595> by changing to `set`.

jonsequitur:


baronfel:
This would need Tactics approval, so I'd suggest the VMR change instead and have it flow back to here. This ensures that we do the change + compensation in SDK at the same time.

jeffhandley:
This change is being applied via [System.CommandLine - Change Terminating API to be callable from C# 7 (dotnet/dotnet#2359)](https://github.com/dotnet/dotnet/pull/2359) and the change will flow into this repo.

PR #2687, opened 2025-09-26
Augment and correct the Option.HelpName API docs

### LINKED ISSUE #2594: Documentation for Option.HelpName inaccurate and mysterious
https://github.com/dotnet/command-line-api/blob/28ad8f704a0d954079f08130d49b24f0c55765c2/src/System.CommandLine/Option.cs#L38-L49

The documentation comments for `Option.HelpName` inaccurately state that it sets the name for the option for the help. That is not true, as it sets the (placeholder) name for the option argument; it doesn't alter the option name in the help.

Also, it features a mysterious remark _"Useful for localization"_. How?

I'd like to submit a PR for this, rephrasing _"name of the Option"_ to _"name of the Option argument"_.

However, what about this _"Useful for localization"_ remark? Is that a leftover from the older code base and can i remove this? Or is this something that will be addressed and become clear with the eventual proper release? Because as it currently stands, it's rather unclear what this remark is supposed to mean.

jeffhandley:
Fixes #2594

We entertained renaming the member but did not arrive an a name improvement that justified making a breaking change at this stage. Instead of renaming, this PR augments and corrects the API docs (and therefore IntelliSense) to make the member's purpose clearer.

We intend to backport this fix into `release/10.0`.

### DIFF
--- a/src/System.CommandLine/Option.cs
+++ b/src/System.CommandLine/Option.cs
@@ -41,12 +41,22 @@ namespace System.CommandLine
         public bool HasDefaultValue => Argument.HasDefaultValue;
 
         /// <summary>
-        /// Gets or sets the name of the Option when displayed in help.
+        /// Gets or sets the placeholder name shown in usage help for the option's value.
+        /// The value will be wrapped in angle brackets (<c>&lt;</c> and <c>&gt;</c>).
         /// </summary>
+        /// <remarks>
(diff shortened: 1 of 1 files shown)


adamsitnik (commented):
It looks great, but could you please apply similar changes to `Argument.HelpName` as well?

https://github.com/dotnet/command-line-api/blob/2464766dba93d0f1f8ff5f3fc16116abc3f908ae/src/System.CommandLine/Argument.cs#L47-L50

PR #2714, opened 2025-10-30
Update global.json to include SDK settings

ericstj:
The `tools` section is proprietary and only used by arcade.

The `sdk` section is used by dotnet to ensure this SDK version is used, and also what the publicly available AzDo build tasks use when installing dotnet.

@jeffhandley


jeffhandley (approved):
I see; thanks! /cc @krwq

PR #2715, opened 2025-11-02
Translate French resource strings in Resources.fr.xlf

CyberSinh:


CyberSinh:
@dotnet-policy-service agree

PR #2718, opened 2025-11-07
Improvements to [suggest] directive

rvuistin:
When using the [suggest] directive with the exact name of a command that has subcommands, you would expect the completions to include those subcommands. However, this does not happen due to how the completion logic currently determines the symbol to complete and the command line context.

Unfortunately, I am unable to run the tests on my machine.


rvuistin:
I'm closing this PR — I'm not satisfied with the changes and will open a replacement PR (#2719) using a different approach.

PR #2719, opened 2025-11-08
Improve command-line completion handling.

rvuistin:
If you specify the complete name of a command that contains subcommands and options, the expected behavior for completions is:
•	Subcommands: All direct subcommands of the specified command should be suggested as completions.
•	Options: All options available for that command (including inherited/global options, if any) should also be suggested as completions.

For example, if you have a command structure like:
```
root
└── parentcmd
    ├── subcmd1
    ├── subcmd2
    └── --option1
```

When you type:
```
parentcmd [TAB]
```
or use the `[suggest]` directive with `parentcmd`, the completion list should include:
```
subcmd1
subcmd2
--option1
```

**Changes:**

- enhanced completion logic to handle cases where the text to match is the command name or is `null`, ensuring all relevant completions (subcommands, options, arguments) are suggested.

- updated `CompletionAction` in `CompletionAction.cs` to process multiple tokens for command-line completion, improving support for multi-token inputs.

### DIFF
--- a/src/System.CommandLine/Command.cs
+++ b/src/System.CommandLine/Command.cs
@@ -265,6 +265,13 @@ namespace System.CommandLine
 
             if (context.WordToComplete is { } textToMatch)
             {
+                // If the text to match is the command name, we should suggest all possible completions (subcommands, options, arguments).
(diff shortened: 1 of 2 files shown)


rvuistin:
I am closing this pull request because the proposed change is not necessary.
After further review, I realized that suggestions for subcommands only appear after typing an additional space, which is the expected behavior.
Therefore, the motivation for this PR is not justified.
Thank you for your understanding.

PR #2742, opened 2025-11-20
Fix XML documentation for POSIX bundling property

gewarren:
Updated XML documentation to clarify value parameter.


gewarren:
@jonsequitur Can you merge it? I don't have perms.

PR #2770, opened 2026-02-11
Fix XML comment formatting in Range.cs

gewarren:
Showed up in https://github.com/dotnet/dotnet-api-docs/pull/12311.


gewarren:
@jonsequitur Can you merge it for me?

Issue #2587, opened 2025-06-23
Missing support for Finnish language


KalleOlaviNiemitalo:
Please add Finnish translations of System.CommandLine strings like "Options:" and "Required command was not provided."

An application would have been able to provide the translations to System.CommandLine 2.0.0-beta4.22272.1, but this is no longer possible in 2.0.0-beta5.25306.1, as the UseLocalizationResources method and the LocalizationResources class were removed from the public API in <https://github.com/dotnet/command-line-api/pull/2041>.  The [migration guide](https://learn.microsoft.com/dotnet/standard/commandline/migration-guide-2.0.0-beta5#configuration) asked me to file an issue.

KalleOlaviNiemitalo:
Please reopen this.

<https://nuget.info/packages/System.CommandLine/2.0.0-rc.1.25451.107> shows that the Finnish satellite assembly is not included in the System.CommandLine package, even though the strings exist at <https://github.com/dotnet/dotnet/blob/2db1f5ee2bdda2e8d873769325fabede32e420e0/src/command-line-api/src/System.CommandLine/Properties/xlf/Resources.fi.xlf>.

Perhaps System.CommandLine.csproj needs to set the `XlfLanguages` property described at <https://github.com/dotnet/dotnet/blob/2db1f5ee2bdda2e8d873769325fabede32e420e0/src/arcade/src/Microsoft.DotNet.XliffTasks/README.md#project-properties>.

adamsitnik:
@ViktorHofer do you have any idea why adding the Finnish resource file gets a different treatment than other languages? Is it the `XlfLanguages` suggested by @KalleOlaviNiemitalo ?

jeffhandley:
We need to check that we can add Finnish translation support through our localization and compliance processes. Moving to Future.

KalleOlaviNiemitalo:
I wonder though if I could just build a resource assembly myself, fake sign it with [PublicSign](https://learn.microsoft.com/dotnet/csharp/language-reference/compiler-options/security#publicsign), and reference it alongside the System.CommandLine package.

ViktorHofer:
> @ViktorHofer do you have any idea why adding the Finnish resource file gets a different treatment than other languages? Is it the XlfLanguages suggested by @KalleOlaviNiemitalo ?

I believe so, yes. Looks like Finish isn't part of the default language set used by xliff-tasks.

PR #2773, opened 2026-02-14
Add Hebrew translations file and enable Finnish .xlf

Avid29:
Adds an xlf file with Hebrew localization resources

### DIFF
--- a/src/System.CommandLine.Tests/LocalizationTests.cs
+++ b/src/System.CommandLine.Tests/LocalizationTests.cs
@@ -7,10 +7,11 @@ namespace System.CommandLine.Tests
     public class LocalizationTests
     {
         private const string CommandName = "the-command";
+        private const string ArgumentName = "arg";
 
         [Theory]
-        [InlineData("es", $"Falta el argumento requerido para el comando: '{CommandName}'.")]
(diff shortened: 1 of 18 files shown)


KalleOlaviNiemitalo:
I'm curious about how the maintainers will handle these translations.  Previously, Finnish translations for <https://github.com/dotnet/command-line-api/issues/2587> were merged in, and they are still in this repository, but they aren't being shipped in the NuGet package.

Avid29:
> I'm curious about how the maintainers will handle these translations. Previously, Finnish translations for #2587 were merged in, and they are still in this repository, but they aren't being shipped in the NuGet package.

Yeah. I had noticed that... 

I believe this is because the Multilingual App Toolkit is out of service. The best approach is probably just to convert all the localization files to .resx and just manage it directly

Avid29:
@KalleOlaviNiemitalo Found the solution actually
https://github.com/dotnet-x-core/dotnet-xliff-tasks?tab=readme-ov-file#project-properties

KalleOlaviNiemitalo:
Yeah I had suggested `XlfLanguages` in <https://github.com/dotnet/command-line-api/issues/2587#issuecomment-3299181725> but apparently there are some _"localization and compliance processes"_ that make this more difficult.

PR #2780, opened 2026-03-10
Fix #1726 - allow `FromAmong` to be case-insensitive

### LINKED ISSUE #1726: Is it possible for the FromAmong option extension to be case-insensitive with strings?
Hey everyone! Just wanted to start by saying I love this API, it has made building a command line for my app so much faster and easier!

I'm using the OptionExtensions.FromAmong() method to allow only certain strings for a "--name" option in my app. The app doesn't care if the strings are uppercase or lowercase, just that the word matches. However, the array of names that are allowed is potentially long (10-20 names). Here is what I"m talking about when I run the command:

```
PS C:\Users\masteroftofu\Documents\cliproject\bin\Debug\net6.0> .\cli run --name name1
Argument 'name1' not recognized. Must be one of:
        'NAME1'
```

Is there any way to accomplish this in System.Commandline without repeating the names in lowercase? It seems that parseArgument isn't invoked if the validation fails, and it also looks like tokens can't be modified in the validators (which makes sense). I've also tried a few other approaches with little success.

I apologize if this is an obvious question, but I am new to C# development and I've dug around a lot for a solution and haven't been able to find one yet.

Thanks, and cheers!

jonsequitur:
This is a proposed fix for #1726. I would appreciate feedback or alternative ideas for the API.

### DIFF
--- a/src/System.CommandLine.Tests/ArgumentTests.cs
+++ b/src/System.CommandLine.Tests/ArgumentTests.cs
@@ -121,11 +121,14 @@ public class ArgumentTests
 
         var result = command.Parse("run name1");
 
+        using var _ = new AssertionScope();
+
         result.Errors.Should().BeEmpty();
+        result.GetValue(argument).Should().Be("name1");
     }
 
     [Fact]
-    public void AcceptOnlyFromAmong_with_comparer_rejects_invalid_values()
+    public void AcceptOnlyFromAmong_with_case_insensitive_comparer_rejects_values_outside_the_accepted_set()
     {
         var argument = new Argument<string>("name");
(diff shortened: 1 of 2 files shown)


src/System.CommandLine/ArgumentValidation.cs
```diff
@@ -149,7 +164,7 @@ void UnrecognizedArgumentError(ArgumentResult argumentResult)
 
                     if (token.Symbol is null || token.Symbol == argument)
                     {
-                        if (Array.IndexOf(values, token.Value) < 0)
+                        if (!values.Contains(token.Value, comparer))
```
jeffhandley: Very nice/clean

PR #2782, opened 2026-03-11
fix #2771 and #2772 - Option action fixes

### LINKED ISSUE #2771: Option actions are only invoked if the parent Command has an action
If you set an `Action` on an `Option` object, it is not automatically invoked.
It is only invoked if the `Command` object also has an action.

The following reproduces the problem, as the first test fails, but the second one passes:

```c#
using System.CommandLine;

namespace CommandLineTest
{
    [TestClass]
    public sealed class CommandLineTests
    {
        [TestMethod]
        public void OptionActionIsInvokedWithoutCommandAction()
        {
            bool actionWasInvoked = false;
            RootCommand root = new("CommandLine Test");

            Option<bool> option = new("--arg");
            option.Action = new CustomAction(_ =>
            {
                actionWasInvoked = true;
                return 0;
            });
            root.Options.Add(option);

            ParseResult parsedArgs = root.Parse(["--arg"]);
            int result = parsedArgs.Invoke();

            Assert.IsTrue(actionWasInvoked);
        }

        [TestMethod]
        public void OptionActionIsInvokedWithCommandAction()
        {
            bool actionWasInvoked = false;
            RootCommand root = new("CommandLine Test");

            Option<bool> option = new("--arg");
            option.Action = new CustomAction(_ =>
            {
                actionWasInvoked = true;
                return 0;
            });
            root.Options.Add(option);
            root.SetAction(r => { });

            ParseResult parsedArgs = root.Parse(["--arg"]);
            int result = parsedArgs.Invoke();

            Assert.IsTrue(actionWasInvoked);
        }

        class CustomAction : SynchronousCommandLineAction
        {
            private readonly Func<ParseResult, int> _callback;

            internal CustomAction(Func<ParseResult, int> callback) => _callback = callback;
            public override int Invoke(ParseResult parseResult) => _callback(parseResult);
            public override bool Terminating => false;
        }
    }
}

```

### LINKED ISSUE #2772: The return value from Option actions is discarded
If you set an `Action` on an `Option` object, it is required to return an integer, but the return value is silently discarded
The problem is on [this line](https://github.com/dotnet/command-line-api/blob/702b2bbdd795ec978c3f2bb395a1a37442c13c8c/src/System.CommandLine/Invocation/InvocationPipeline.cs#L113), where the return value is discarded.

The following reproduces the problem

```
using System.CommandLine;

namespace CommandLineTest
{
    [TestClass]
    public sealed class CommandLineTests
    {
        [TestMethod]
        public void OptionActionReturnValuePropagates()
        {
            RootCommand root = new("CommandLine Test");

            Option<bool> option = new("--arg");
            option.Action = new CustomAction(_ => 42);
            root.Options.Add(option);
            root.SetAction(r => { });

            ParseResult parsedArgs = root.Parse(["--arg"]);
            int result = parsedArgs.Invoke();

            Assert.AreEqual(42, result);
        }

        class CustomAction : SynchronousCommandLineAction
        {
            private readonly Func<ParseResult, int> _callback;

            internal CustomAction(Func<ParseResult, int> callback) => _callback = callback;
            public override int Invoke(ParseResult parseResult) => _callback(parseResult);
            public override bool Terminating => false;
        }
    }
}

```

jonsequitur:
Fixes #2771 
Fixes #2772 

This pull request improves how non-terminating option actions are handled in the command-line invocation pipeline, ensuring their return values are correctly propagated and prioritized over command actions when appropriate. It also adds comprehensive tests for these scenarios and refactors the test action helpers to support custom return values.

* Refactored `InvocationPipeline.Invoke` and `InvokeAsync` to propagate the return value from non-terminating pre-actions (such as option actions) and ensure that if both a pre-action and a command action return nonzero exit codes, the pre-action's value takes precedence. If no command action is present, the pre-action's return value is used; otherwise, the command action's return value is used only if all pre-actions return zero.

### DIFF
--- a/src/System.CommandLine/Invocation/InvocationPipeline.cs
+++ b/src/System.CommandLine/Invocation/InvocationPipeline.cs
@@ -55,14 +55,14 @@ namespace System.CommandLine.Invocation
                         break;
 
                     case AsynchronousCommandLineAction asyncAction:
-                        var startedInvocation = asyncAction.InvokeAsync(parseResult, cts.Token);
-
                         var timeout = parseResult.InvocationConfiguration.ProcessTerminationTimeout;
 
                         if (timeout.HasValue)
                         {
-                            terminationHandler = new(cts, startedInvocation, timeout.Value);
+                            terminationHandler = new(cts, timeout.Value);
                         }
+
+                        var startedInvocation = asyncAction.InvokeAsync(parseResult, cts.Token);
 
                         if (terminationHandler is null)
                         {
@@ -70,6 +70,7 @@ namespace System.CommandLine.Invocation
                         }
                         else
                         {
+                            terminationHandler.StartedHandler = startedInvocation;
                             // Handlers may not implement cancellation.
                             // In such cases, when CancelOnProcessTermination is configured and user presses Ctrl+C,
                             // ProcessTerminationCompletionSource completes first, with the result equal to native exit code for given signal.
--- a/src/System.CommandLine.Tests/CustomParsingTests.cs
+++ b/src/System.CommandLine.Tests/CustomParsingTests.cs
@@ -561,6 +561,23 @@ public class CustomParsingTests
         i.Should().Be(2);
     }
 
+    [Fact] // https://github.com/dotnet/command-line-api/issues/2743
+    public void Setting_CustomParser_to_null_reverts_to_default_parsing()
+    {
+        Argument<int> argument = new("int")
+        {
+            CustomParser = (_) => 0
+        };
+
+        argument.CustomParser = null;
+
+        var command = new RootCommand { argument };
+
+        var result = command.Parse("123");
+
+        result.GetValue(argument).Should().Be(123);
+    }
+
     [Fact]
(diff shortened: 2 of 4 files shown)


src/System.CommandLine/Invocation/InvocationPipeline.cs
```diff
@@ -10,38 +10,52 @@ internal static class InvocationPipeline
...
-            {
-                return ReturnCodeForMissingAction(parseResult);
-            }
-
             ProcessTerminationHandler? terminationHandler = null;
             using CancellationTokenSource cts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
 
             try
             {
+                int exitCode = 0;
```
jeffhandley: It wasn't obvious to me at first glance that `exitCode` represents the preaction exit code. I suggest renaming to `preActionExitCode` since we don't set it to represent anything else.

jeffhandley (approved):
Just one suggestion, but looks good. I also updated the PR description to add more context.

PR #2783, opened 2026-03-13
fix #2760 - provide parser override via new `Argument.CaptureRemainingTokens` property

### LINKED ISSUE #2760: Make it possible for commands like `dnx <toolname>` to 'accrue' all tokens after `<argname>` for forwarding, not option-processing
Today in the .NET CLI experience, we have a feature that allows executing .NET Tools - .NET applications that are delivered via NuGet packages that the .NET CLI tooling can acquire on-demand or before execution. The command-line experience for these commands is structured something like the following:

```
Command("dnx")
  - Option<string[]>("--source") // for NuGet sources to get the packages from
  - Option<bool>("--interactive") // if nuget auth/other interactive experiences can be used in this interaction
  - Argument<string>("toolPackageName") // name of the NuGet package containing the tool command to run
  - Argument<string[]>("toolArgs") // arguments to forward to the tool command when run
```

and users use this command like the following:

```
dnx --source https://api.nuget.org/v3/index.json --interactive myTool -a 1 -b 2 -c --help
```

Today, as System.CommandLine processes each Token during Parsing, it classifies the `--help` token as belonging to the `--help` option that is provided by default as a `recursive` option, in this case coming from the `dotnet` RootCommand that hosts the implementation of the `dnx` command.

This is confusing to users, as they interpret the above command as

| Token | Option/Argument it belongs to |
| - | - |
| `(dotnet) dnx` | `dnx` Command under the `Dotnet` RootCommand |
| `--source` | `--source` Option | 
| `https://api.nuget.org/v2/index.json` | Argument of the `--source` Option |
| `--interactive` | `--interactive` Option |
| `myTool` | `toolPackageName` Argument |
| `-a` | `toolArgs` Argument, position 0 |
| `1` | `toolArgs` Argument, position 1 |
| `-b` | `toolArgs` Argument, position 2 | 
| `2` | `toolArgs` Argument, position 3 |
| `-c` | `toolArgs` Argument, position 4 |
| `--help` | `toolArgs` Argument, position 5 |

Users can force this behavior today if they invoke the above command like so:

```
dnx --source https://api.nuget.org/v3/index.json --interactive myTool -- -a 1 -b 2 -c --help
```

But to many users this is foreign - why can't the parser detect that anything _after_ `myTool` should be treated as an argument to the tool?

### Proposal

It should be possible to mark an Argument as _greedy_, or _final_, or some other semantic, and after the parser processes this argument every subsequent token, regardless of `TokenType`, would be aggregated into an `IEnumerable<string>` Argument (or similar collection-typed Argument) as if `--` were inserted into the token stream.

jonsequitur:
This fixes #2760 by adding a new property currently named `Argument.CaptureRemainingTokens`. 

I'm not sure this is the right name but the behavior is that this argument will capture everything following it regardless of whether it happens to match some well-known symbol. A few notes on this implementation:

* It's intended to be used with enumerable types but this is not enforced.
* It stops capturing when it reaches the arity limit for the argument. For this reason, a name like `Greedy` might be more accurate. Can we assume this term is intuitive enough?
* The `--` token has no special meaning when captured by an `Argument` with this property set to `true`. It gets captured as one of the tokens. 

@baronfel  @tmds Thoughts?

### DIFF
--- a/src/System.CommandLine.ApiCompatibility.Tests/ApiCompatibilityApprovalTests.System_CommandLine_api_is_not_changed.approved.txt
+++ b/src/System.CommandLine.ApiCompatibility.Tests/ApiCompatibilityApprovalTests.System_CommandLine_api_is_not_changed.approved.txt
@@ -1,6 +1,7 @@
 ﻿System.CommandLine
   public abstract class Argument : Symbol
     public ArgumentArity Arity { get; set; }
+    public System.Boolean CaptureRemainingTokens { get; set; }
     public System.Collections.Generic.List<System.Func<System.CommandLine.Completions.CompletionContext,System.Collections.Generic.IEnumerable<System.CommandLine.Completions.CompletionItem>>> CompletionSources { get; }
     public System.Boolean HasDefaultValue { get; }
     public System.String HelpName { get; set; }
--- a/src/System.CommandLine.Tests/ParserTests.CaptureRemainingTokens.cs
+++ b/src/System.CommandLine.Tests/ParserTests.CaptureRemainingTokens.cs
@@ -1,0 +1,261 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.CommandLine.Tests.Utility;
+using FluentAssertions;
+using FluentAssertions.Execution;
+using Xunit;
+
+namespace System.CommandLine.Tests
+{
+    public partial class ParserTests
+    {
+        public class CaptureRemainingTokens
+        {
+            [Fact]
+            public void Option_like_tokens_after_capturing_argument_are_captured()
+            {
+                var option = new Option<string>("--source");
+                var toolName = new Argument<string>("toolName");
+                var toolArgs = new Argument<string[]>("toolArgs") { CaptureRemainingTokens = true };
+                var command = new RootCommand
+                {
+                    option,
+                    toolName,
+                    toolArgs
+                };
+
+                var result = command.Parse("--source https://nuget.org myTool -a 1 --help");
+
+                using var _ = new AssertionScope();
+                result.GetValue(option).Should().Be("https://nuget.org");
+                result.GetValue(toolName).Should().Be("myTool");
+                result.GetValue(toolArgs).Should().BeEquivalentSequenceTo("-a", "1", "--help");
+                result.Errors.Should().BeEmpty();
+                result.UnmatchedTokens.Should().BeEmpty();
+            }
+
+            [Fact]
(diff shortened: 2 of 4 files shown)


KalleOlaviNiemitalo:
> The -- token has no special meaning when captured by an Argument with this property set to true. It gets captured as one of the tokens.

That is what I would expect.

KalleOlaviNiemitalo:
> It's intended to be used with enumerable types but this is not enforced.

I imagine a custom parser could very well transform this token sequence to a non-enumerable type.

jeffhandley (commented):
Since this is an implicit `--`, I started entertaining if we could take a different approach where we encapsulate the current double-dash implementation in a special-purpose `Argument<string[]>` that always captures all remaining tokens. That argument could be explicitly added to the CLI definition (as the last argument) to make the `--` implicit, and if it's not already added to the CLI then we would add it with the "--" name.

An approach like that could prevent us from having 2 different mechanism for capturing the remaining tokens, and we could disallow that special argument from having a max capacity that can get filled up.

**BUT** after some feasibility research, I don't think it would come in very cleanly for a few reasons:

1. Trying to encapsulate `--` as an argument or option would affect tokenizing logic and how the remaining tokens are treated/exposed
2. It would be challenging to make that change in a backwards-compatible manner for the existing `UnmatchedTokens`
3. Would it be an Option or an Argument? Kind of both. Kind of neither.

I'm not convinced on the `CaptureRemainingTokens` name either, and I'm anxious about the complexity of filling to capacity and needing to spill over and that being potentially unintuitive. But this does plug in pretty cleanly overall.

I'll come back to this PR after thinking about it some more.

jonsequitur:
Yeah, the name `CaptureRemainingTokens` doesn't entirely convey one implication of the design, which is that an argument with this property set to true can still be limited by its arity. Another name I was weighing was `Greedy` which I think is reasonable as well. It conveys the sense that while other parsing rules (e.g. the names of options, or the `--` token) might apply, this argument will override them.

jonsequitur:
> Since this is an implicit --, I started entertaining if we could take a different approach where we encapsulate the current double-dash implementation in a special-purpose Argument<string[]> that always captures all remaining tokens.

We considered this design when we were changing the way `--` worked. If I recall the conversation correctly, we decided that the `--` is an escape mechanism and a feature of the grammar, and not a symbol in and of itself. Like POSIX bundling, it's there to meet end users' expectations and provide ecosystem consistency, not as an abstraction for developers.

It is appealing as a way to reduce concept count though.

baronfel:
This looks like a reasonable approach! So for us in the dotnet CLI, we'd set the `toolArgs` property to `CaptureRemainingTokens` and be good to go!

baronfel (approved):

PR #2784, opened 2026-03-17
Fix #2743 - allow `Argument<T>.CustomParser` to be set to null

### LINKED ISSUE #2743: CustomParser of Argument<T> cannot be reset by setting to null
When I set the CustomParser property of an Argument&lt;T&gt; to a custom parser, and then set it back to null, the custom parser is still being used, instead of the default parser.

Code to reproduce:
```csharp
#:package System.CommandLine@2.0.0

using System.CommandLine;

Argument<int> argument = new("int")
{
    CustomParser = (_) =>
    {
        Console.WriteLine("Using Custom Parser");
        return 0;
    }
};

argument.CustomParser = null;

RootCommand command = new()
{
    argument
};

var result = command.Parse("123");

Console.WriteLine(result.GetValue(argument));
```
Expected output:
```
123
```
Actual output:
```
Using Custom Parser
0
```

jonsequitur:
Fixes #2743

The PR also includes timeout handling changes in the tests where CI was tripping up often enough to bundle that fix into this PR to ensure it could be green.

### DIFF
--- a/src/System.CommandLine/System.Runtime.CompilerServices/Range.cs
+++ b/src/System.CommandLine/System.Runtime.CompilerServices/Range.cs
@@ -10,7 +10,7 @@ namespace System
 {
     /// <summary>Represents a type that can be used to index a collection either from the beginning or the end.</summary>
     /// <remarks>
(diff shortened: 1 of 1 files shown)


KalleOlaviNiemitalo:
Contains changes from <https://github.com/dotnet/command-line-api/pull/2783>, too.

PR #2789, opened 2026-04-01
Fix FirstArgumentIsRootCommand falsely matching option values

chris-peterson:
### Problem

  `FirstArgumentIsRootCommand` used `Path.GetFileName(args[0])` to extract a potential command name from the first argument. Since `Path.GetFileName`
  returns the last segment after `/`, option values like `/p:DockerImage=registry.example.com/project/dotnet` would extract `dotnet` — falsely matching a
  root command named `dotnet`.

  This caused the parser to skip `args[0]` (treating it as the root command invocation), losing the option value entirely.

  ### Approach

  Replaced the `Path.GetFileName` heuristic with a direct `rootCommand.EqualsNameOrAlias(args[0])` check. Only an exact name or alias match should identify
  `args[0]` as the root command. The existing `inferRootCommand` + `ExecutablePath` check for unsplit string parsing is unchanged.

  ### FirstArgumentIsRootCommand behavior

  | `args[0]` | Before | After | Notes |
  |---|---|---|---|
  | `dotnet` | ✅ match | ✅ match | Exact name match, unchanged |
  | `ExecutablePath` (unsplit string) | ✅ match | ✅ match | `inferRootCommand` path, unchanged |
  | `/p:DockerImage=.../dotnet` | 🐛 match | ✅ no match | `Path.GetFileName` extracted `dotnet` |
  | `/p:Key=something/myapp` | 🐛 match | ✅ no match | `Path.GetFileName` extracted `myapp` |
  | `-p:DockerImage=.../dotnet` | 🐛 match | ✅ no match | Same extraction issue with `-p:` prefix |
  | `./myapp` | ✅ match | ⚠️ no match | Narrow edge case; .NET `Main(args)` never includes exe path |

  ### Testing

  Added regression tests covering:
  - Option value containing the command name after `/` (e.g., `/p:Key=something/myapp`)
  - Three property option formats (`/p:`, `-p:`, `--property:`) with docker image paths ending in the command name

### DIFF
--- a/src/System.CommandLine.Tests/ParserTests.RootCommandAndArg0.cs
+++ b/src/System.CommandLine.Tests/ParserTests.RootCommandAndArg0.cs
@@ -101,6 +101,39 @@ public partial class ParserTests
             result3.RootCommandResult.Command.Should().BeSameAs(result1.RootCommandResult.Command);
         }
 
+        [Fact]
+        public void When_parsing_a_string_array_option_values_containing_command_name_after_slash_are_not_treated_as_root_command()
+        {
+            // Path.GetFileName("/p:Key=something/myapp") returns "myapp",
+            // which should not be matched as the root command name.
+            var option = new Option<string>("--property", "/p") { Arity = ArgumentArity.ExactlyOne };
+            var command = new Command("myapp");
+            command.Options.Add(option);
+
+            var result = command.Parse(Split("/p:Key=something/myapp"));
+
+            result.Errors.Should().BeEmpty();
+            result.GetResult(option).Should().NotBeNull();
+        }
+
+        [Theory]
+        [InlineData("/p:DockerImage=registry.example.com/project/dotnet")]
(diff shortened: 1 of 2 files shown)


jonsequitur:
>Narrow edge case; .NET Main(args) never includes exe path

One thing to be aware of here is that an important goal of the parser is to be able to parse the values returned by `Environment.CommandLine` as well as `Environment.GetCommandLineArgs()`, the first arg of which will be a path.

I think your fix is still safe though because I don't think that either `Environment.CommandLine` or `Environment.GetCommandLineArgs()[0]` ever returns a *relative* path to the executable.

@adamsitnik Thoughts?



image

baronfel:
I agree with

> I think your fix is still safe though because I don't think that either Environment.CommandLine or Environment.GetCommandLineArgs()[0] ever returns a relative path to the executable.

argc is always absolute by spec and in practice from what I've seen.

baronfel (approved):
Love it, thanks for the contribution!

jonsequitur (approved):

PR #2799, opened 2026-04-16
Add System.CommandLine.StaticCompletions project

slang25:
## Summary

Imports the `System.CommandLine.StaticCompletions` library and its test project from `dotnet/sdk` into this repo. The library generates static shell completion scripts (bash, zsh, pwsh, fish, nushell) from a `System.CommandLine` `Command` tree. In-repo, the `System.CommandLine` dependency becomes a `ProjectReference`; tests run against a real build via `Verify.Xunit` snapshots (replacing the SDK's Helix-specific test harness).

This is step 1 of a 3-step migration: land here first, wait for Maestro to flow a package, then remove the sources from `dotnet/sdk`.

## Test plan
- [ ] CI builds the full solution without warnings
- [ ] `dotnet test src/System.CommandLine.StaticCompletions.Tests` passes all 15 snapshot tests


slang25:
@baronfel what do you think to this, does this need a bit more coordination?

src/System.CommandLine.StaticCompletions/shells/BashShellProvider.cs
```diff
@@ -0,0 +1,226 @@
...
+        }
+
+        return completions.ToArray();
+    }
+
+    /// <summary>
+    /// Generates a call to <code>dotnet complete &lt;string&gt; --position &lt;int&gt;</code> for dynamic completions where necessary, but in a more generic way
+    /// </summary>
+    /// <returns></returns>
+    /// <remarks>TODO: this is currently bound to the .NET CLI's 'dotnet complete' command - this should be definable/injectable per-host instead.</remarks>
```
baronfel:
```suggestion
    /// <summary>
    /// Generates a call to <code>&lt;appname&gt; complete &lt;string&gt; --position &lt;int&gt;</code> for dynamic completions where necessary - for example when completing NuGet package identifiers.
    /// </summary>
```

src/System.CommandLine.StaticCompletions/shells/FishShellProvider.cs
```diff
@@ -0,0 +1,28 @@
...
+    public string Extension => "fish";
+
+    public string HelpDescription => Strings.FishShellProvider_HelpDescription;
+
+    // override the ToString method to return the argument name so that CLI help is cleaner for 'default' values
+    public override string ToString() => ArgumentName;
+
+    private static readonly string _dynamicCompletionScript =
+        """
+        # fish parameter completion for the dotnet CLI
```
baronfel: I think we have an open PR for fish completions in the SDK right now that would need syncing here.

src/System.CommandLine.StaticCompletions/shells/PowershellShellProvider.cs
```diff
@@ -0,0 +1,246 @@
...
+    /// Generate a call into `dotnet complete` for dynamic argument completions, then binds the returned values as CompletionResults.
+    /// </summary>
+    /// <remarks>TODO: this is currently bound to the .NET CLI's 'dotnet complete' command - this should be definable/injectable per-host instead.</remarks>
+    private static void GenerateDynamicCompletionsCall(IndentedTextWriter writer)
+    {
+        writer.WriteLine("$text = $commandAst.ToString()");
+        writer.WriteLine("$dotnetCompleteResults = @(dotnet complete --position $cursorPosition \"$text\") | Where-Object { $_ -NotMatch \"^-|^/\" }");
+        writer.WriteLine("$dynamicCompletions = $dotnetCompleteResults | Foreach-Object { [CompletionResult]::new($_, $_, [CompletionResultType]::ParameterValue, $_) }");
+        writer.WriteLine("$completions += $dynamicCompletions");
+    }
```
baronfel, line 235: This will need to be changed to not rely on `dotnet`.

src/System.CommandLine.StaticCompletions/shells/ZshShellProvider.cs
```diff
@@ -0,0 +1,377 @@
...
+        writer.WriteLine("esac");
+    }
+
+    private static void GenerateDynamicCompleter(IndentedTextWriter writer)
+    {
+        writer.WriteLine("(dotnet_dynamic_complete)");
+        writer.Indent++;
+        writer.WriteLine("local completions=()");
+        // TODO: we're directly calling dotnet complete here - we need something pluggable.
+        writer.WriteLine("local result=$(dotnet complete -- \"${original_args[@]}\")");
```
baronfel: We also need to change this invocation to not rely on `dotnet`.

baronfel (commented):
First pass on some things I _suspected_ would need changing before making part of this repo.

src/System.CommandLine.StaticCompletions/shells/BashShellProvider.cs
KalleOlaviNiemitalo: I feel it should use a directive rather than a subcommand.

baronfel:
@slang25 I wouldn't use the suggest directive for this, and I'm not sure about using directives at all for the feature - the suggest directive requires that the tool support the directive, and have dotnet-suggest installed, which will almost certainly not be the case for AOT tools meant for non .NET audiences. The most resilient thing is for the app itself to also implement completion.

jonsequitur:
Agreed that following the design of `dotnet-suggest` or being compatible with it isn't a goal. Why not use a directive though?

src/System.CommandLine.StaticCompletions/CompletionInvocation.cs
```diff
@@ -0,0 +1,32 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace System.CommandLine.StaticCompletions;
+
+/// <summary>
+/// Describes how generated shell completion scripts invoke the target application
+/// to resolve dynamic completions at runtime.
+/// </summary>
+public abstract record CompletionInvocation
```
baronfel: Good thinking on the strategy approach here - it has me thinking about the 'protocol' of each of these modes.

* `dotnet suggest` has a known calling 'API'  expressed by its CLI grammar
* 'Subcommand' would be defining a new calling protocol that happens to match what the `dotnet` CLI has today. Is that API sufficient for a more general library? I _suspect_ not because it only provides the name/label of the completion item today - for example NuGet package identifier completions don't provide any deeper integration into the various shell providers.
* would these protocols need to be extensible by consumers of this library? If so, how would the shell providers in this library know about the new providers? It feels coupled today in a way that is hard to eject out of from a consumer's point of view.

jonsequitur: It's worth mentioning that the `[suggest]` directive (which no longer has any dependency on the `dotnet-suggest` tool) is:

* Enabled by default today
* Extensible in terms of its protocol without needing breaking changes

So the subcommand strategy isn't needed to achieve protocol changes, and the System.CommandLine completions model supports a richer set of properties on completions, inspired by LSP, than the simple `dotnet` CLI model.

baronfel: Great point-out - the only thing I think we need to worry about with the `[suggest]` invocation is that the suggest directive must be enabled on the Command. It _is_ by default as we discussed, but if a user disables it then these completion scripts would only be partially correct.

jonsequitur: Right. People would have to deliberately remove it from their parser configuration though, and I suspect most don't even know it's there. So it sounds like the choice is between:

* Dynamic completions work if you add this specific subcommand, versus
* Dynamic completions break if you remove this specific directive.

My sense is that the latter will result in broader adoption.

src/System.CommandLine.StaticCompletions/shells/FishShellProvider.cs
slang25: Now synced — I've ported the hybrid static+dynamic fish generator from dotnet/sdk#53716, with the dynamic call adapted to the `[suggest]` directive (which also resolves the TODO there about the call being bound to the .NET CLI's `complete` command). Snapshot tests are ported too and match the SDK's output apart from the dynamic invocation. While doing this I also picked up dotnet/sdk#54743, which replaced the outdated nushell script after this PR was opened.

src/System.CommandLine.StaticCompletions/CompletionInvocation.cs
slang25: Agreed — that's the direction this branch has taken: the shell providers now emit `[suggest]` directive invocations directly, and the configurable invocation strategy is gone. To address the partial-correctness concern about apps that disable the directive, `completions script` now emits a warning when the command tree contains dynamic completions but the root command has no `[suggest]` directive, so the failure mode is visible at generation time rather than silent at completion time.

src/System.CommandLine.StaticCompletions/shells/PowershellShellProvider.cs
slang25: Done — all generated scripts now call back into the application's own binary rather than `dotnet`.

src/System.CommandLine.StaticCompletions/shells/ZshShellProvider.cs
slang25: Done — all generated scripts now call back into the application's own binary rather than `dotnet`.

src/System.CommandLine.StaticCompletions/System.CommandLine.StaticCompletions.csproj
```diff
@@ -0,0 +1,33 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
```
baronfel, line 2: @ViktorHofer if we wanted to publish this package publically, we'd need to add `IsShipping` right?

baronfel (approved):

src/System.CommandLine.StaticCompletions/System.CommandLine.StaticCompletions.csproj
ViktorHofer: yes

baronfel: @slang25 if you can add `IsShipping` `true` to this project that'd be the last thing. I don't want to do it myself because it'll blow up the ability for me to approve/merge this.

ViktorHofer: Sorry, my fault. IsShipping defaults to true so no need to set it in this project: https://github.com/dotnet/arcade/blob/b8b677cf7d0e22a28f3a42b89e02527aa150a47b/src/Microsoft.DotNet.Arcade.Sdk/tools/Version.BeforeCommonTargets.targets#L19

baronfel:
@slang25 thanks for refreshing this! There's been another merge to main that conflicts - can you address the conflicts? I'm happy to merge ASAP after that.

src/System.CommandLine.StaticCompletions/System.CommandLine.StaticCompletions.csproj
```diff
@@ -0,0 +1,33 @@
...
+    <Compile Update="Resources\Strings.Designer.cs">
+      <DesignTime>True</DesignTime>
+      <AutoGen>True</AutoGen>
+      <DependentUpon>Strings.resx</DependentUpon>
+    </Compile>
+    <EmbeddedResource Update="Resources\Strings.resx">
+      <Generator>MSBuild:_GenerateResxSource</Generator>
+      <LastGenOutput>Strings.Designer.cs</LastGenOutput>
+      <GenerateSource>true</GenerateSource>
+    </EmbeddedResource>
```
ViktorHofer:
```suggestion
    <EmbeddedResource Update="Resources\Strings.resx" GenerateSource="true" />
```

baronfel (approved):

src/System.CommandLine.StaticCompletions/System.CommandLine.StaticCompletions.csproj
slang25: Thanks both — left `IsShipping` unset since it defaults to `true`, and picked up the resource-generation simplification in the same commit. I've also just updated the branch against main: that pulled in the `.sln` → `.slnx` migration (added both StaticCompletions projects to `System.CommandLine.slnx`) and the `Newtonsoft.Json` bump, so it's conflict-free again. Full solution builds clean and all 30 completion tests pass.

PR #2807, opened 2026-05-20
Fix exit code handling in GetCompletions method to ensure correct res…

### LINKED ISSUE #2137: GetCompletions should check exit code of invoked application

AhmedMyoussif:
Closes #2137

### Summary of Changes
- Reverted the synchronous event-based approach from previous attempts to use the stable `StreamReader` via `ReadToEndAsync()`.
- Used `process.WaitForExit(timeout)` to ensure that the process and any potential grandchild processes completely close and flush their pipes, addressing the race conditions / thread-safety concerns mentioned in #2160.
- Verified that `process.ExitCode == 0` before assigning the suggestions to the result string.

### DIFF
--- a/src/System.CommandLine.Suggest/SuggestionStore.cs
+++ b/src/System.CommandLine.Suggest/SuggestionStore.cs
@@ -44,8 +44,12 @@ namespace System.CommandLine.Suggest
 
                     Task<string> readToEndTask = process.StandardOutput.ReadToEndAsync();
 
-                    if (readToEndTask.Wait(timeout))
+                   if (readToEndTask.Wait(timeout))
                     {
+                        if(process.ExitCode == 0)
+                        {
+                            result = readToEndTask.Result;
+                        }
                         result = readToEndTask.Result;
                     }
                     else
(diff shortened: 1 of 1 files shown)


src/System.CommandLine.Suggest/SuggestionStore.cs
```diff
@@ -44,8 +44,12 @@ public string GetCompletions(string exeFileName, string suggestionTargetArgument
...
                     Task<string> readToEndTask = process.StandardOutput.ReadToEndAsync();
 
-                    if (readToEndTask.Wait(timeout))
+                   if (readToEndTask.Wait(timeout))
                     {
+                        if(process.ExitCode == 0)
+                        {
+                            result = readToEndTask.Result;
+                        }
                         result = readToEndTask.Result;
```
jonsequitur: This line should now be removed.

AhmedMyoussif: Done! Thank you for catching it

Issue #2785, opened 2026-03-19
Argument on root unexpectedly "leaked" to commands


MattHyman:
I am currently porting a CLI tool to system.commandLine and have run into a scenario where we only expect to take a argument on root but not on its commands. If we put the argument after the commands everything works but if we put it before the argument is silently ignored. Furthermore the argument appears in the help text for the commands where it is not relevant. 

Here is an example of this scenario: 
```
using System.CommandLine;
   using System.CommandLine.Help;
   
   // Minimal repro for two related issues with System.CommandLine 2.0.3:
   //
   // 1. Root command arguments appear in subcommand help text (/?).
   // 2. Root command argument preceding a subcommand is silently ignored with no error.
   
   var sampleArgument = new Argument<string>("sampleArgument")
   {
       Description = "An optional argument on the root command.",
       Arity = ArgumentArity.ZeroOrOne,
   };
   
   var feedbackCommand = new Command("/subcommand")
   {
       Description = "A subcommand with no arguments of its own.",
   };
   feedbackCommand.SetAction(_ => Console.WriteLine("[subcommand] Subcommand ran."));
   
   var rootCommand = new RootCommand("A sample app. Example: app hello");
   
   var builtInHelp = rootCommand.Options.OfType<HelpOption>().First();
   builtInHelp.Aliases.Clear();
   builtInHelp.Hidden = true;
   rootCommand.Options.Add(new HelpOption("/help", "/?"));
   
   rootCommand.Add(sampleArgument);
   rootCommand.Add(feedbackCommand);
   rootCommand.TreatUnmatchedTokensAsErrors = true;
   
   rootCommand.SetAction(parseResult =>
   {
       var value = parseResult.GetValue(sampleArgument);
       if (value is not null) Console.WriteLine($"[root] sampleArgument = {value}");
       else Console.WriteLine("[root] No argument specified.");
   });
   
   // Issue 1: /subcommand /?
   // Expected: help shows only /subcommand's own arguments/options.
   // Actual:   <sampleArgument> (root-only) appears in the usage line and
   //           Arguments section, implying it is an input to /subcommand.
   Console.WriteLine("=== Issue 1: app /subcommand /? ===");
   rootCommand.Parse(["/subcommand", "/?"]).Invoke();
   
   Console.WriteLine();
   
   // Issue 2: hello /subcommand
   // Expected: error because "hello" is not valid for /subcommand,
   //           or the root action runs with the argument.
[...]

PR #2809, opened 2026-05-21
Fix parent non-recursive arguments leaking into subcommands (#2785)

AhmedMyoussif:
## Description
This PR resolves issue #2785 where non-recursive positional arguments from parent commands were incorrectly inherited, validated, and displayed in subcommands.

### Key Changes
- **Early Interception in `ParseOperation`**: Introduced `MoveNonRecursiveParentArgumentsToUnmatched` inside `ParseSubcommand()`. When entering a subcommand, the parser immediately intercepts parent arguments configured with `Recursive = false`, purges them from the `SymbolResultTree`, and registers their tokens as `UnmatchedTokens` under the root command result. This completely prevents validation leaks and downstream crashes.
- **Help Architecture Alignment**: Updated `HelpBuilder.cs` (`GetUsageParts` and `GetCommandArgumentRows`) to filter out parent arguments where `Recursive == false` when rendering usage strings and argument tables for subcommands.
- **Public API Contract**: Added the `Recursive` property to the `Argument` class (defaulting to `true` to avoid breaking changes) and updated the alphabetical baseline file for API compatibility checks.
- **Test Coverage**: Added dedicated unit tests for both parsing telemetry (`ParserTests.cs`) and layout rendering accuracy (`HelpBuilderTests.cs`).

### Verification
- Tested locally on `net10.0` target.
- All 929 test cases passing successfully.


AhmedMyoussif:
@dotnet-policy-service agree

PR #2814, opened 2026-06-05
Fix empty description help section

### LINKED ISSUE #2114: When Description is empty, the description section should be ommitted from help.
Per title. For an empty description, the help currently shows:
```
Description:


```

The API treats descriptions as optional (`string?`), so the help should do too.

cc @adamsitnik @jonsequitur

marko1olo:
Fixes #2114.

This suppresses the default help `Description:` section when the command description is null, empty, or whitespace. Descriptions are optional, so help output should not render an empty section.

Changes:
- `SynopsisSection()` now returns `false` for null/empty/whitespace descriptions.
- Adds regression coverage for null, empty, and whitespace descriptions.
- Updates an existing help expectation that previously asserted a blank `Description:` section.

Validation:
- `git diff --check`

Not run locally:
- targeted `dotnet test` for `System.CommandLine.Tests` help tests.

Reason: current machine CPU stayed above the local process gate threshold, so I did not start a .NET test/build process. The change is small and covered by added/updated tests in this PR.

### DIFF
--- a/src/System.CommandLine.Tests/Help/HelpBuilderTests.cs
+++ b/src/System.CommandLine.Tests/Help/HelpBuilderTests.cs
@@ -74,6 +74,20 @@ namespace System.CommandLine.Tests.Help
             _console.ToString().Should().Contain(expected);
         }
 
+        [Theory]
+        [InlineData(null)]
+        [InlineData("")]
+        [InlineData(" ")]
+        public void Synopsis_section_is_not_rendered_when_description_is_empty(string? description)
+        {
+            var command = new RootCommand(description);
+
+            _helpBuilder.Write(command, _console);
+
+            _console.ToString().Should().NotContain("Description:");
+            _console.ToString().Should().Contain("Usage:");
+        }
+
         [Fact]
         public void Command_name_in_synopsis_can_be_specified()
         {
--- a/src/System.CommandLine.Tests/Help/HelpBuilderTests.cs
+++ b/src/System.CommandLine.Tests/Help/HelpBuilderTests.cs
@@ -78,7 +78,7 @@ namespace System.CommandLine.Tests.Help
         [InlineData(null)]
         [InlineData("")]
         [InlineData(" ")]
-        public void Synopsis_section_is_not_rendered_when_description_is_empty(string? description)
+        public void Command_description_section_is_not_rendered_when_description_is_empty(string? description)
         {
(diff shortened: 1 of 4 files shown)


src/System.CommandLine/Help/HelpBuilder.Default.cs
```diff
@@ -185,6 +185,11 @@ public static IEnumerable<Func<HelpContext, bool>> GetLayout()
         public static Func<HelpContext, bool> SynopsisSection() =>
             ctx =>
             {
+                if (string.IsNullOrWhiteSpace(ctx.Command.Description))
```
jonsequitur, line 188: The change is correct but I had to dig a little to see that this method name is wrong. It should be something like `CommandDescriptionSection`. Would you mind renaming it?

src/System.CommandLine.Tests/Help/HelpBuilderTests.cs
```diff
@@ -74,6 +74,20 @@ public void Synopsis_section_properly_wraps_description()
             _console.ToString().Should().Contain(expected);
         }
 
+        [Theory]
+        [InlineData(null)]
+        [InlineData("")]
+        [InlineData(" ")]
+        public void Synopsis_section_is_not_rendered_when_description_is_empty(string? description)
```
jonsequitur:
```suggestion
        public void Command_description_section_is_not_rendered_when_description_is_empty(string? description)
```

src/System.CommandLine/Help/HelpBuilder.Default.cs
marko1olo: Renamed in `7829e695` to `CommandDescriptionSection()` and updated the XML summary as well.

Local checks: `git diff --check` passed. I couldn't run `dotnet test` here because this checkout requires SDK `11.0.100-preview.4.26210.111` from `global.json`, while this machine only has SDK `10.0.202` installed.

src/System.CommandLine.Tests/Help/HelpBuilderTests.cs
marko1olo: Applied in `7829e695`.

jonsequitur (approved):

jonsequitur:
@marko1olo This PR can be merged once you've accepted the CLA.

PR #2826, opened 2026-07-01
fix: complete Korean(ko) translations in Resources.kr.xlf

handbros:
## Summary

This PR completes the Korean translation in `Resources.ko.xlf` by translating untranslated strings and refining existing ones. Unlike English, which follows an SVO word order, Korean uses an SOV structure; as a result, the order of placeholders for string interpolation has been modified in some strings.

## Testing

Verified by setting `CultureInfo.DefaultThreadCurrentUICulture = CultureInfo.GetCultureInfo("ko-KR")` before parsing and running the app without the required `--input` option.


handbros:
@dotnet-policy-service agree

PR #2820, opened 2026-06-17
Improve RootCommand executable name handling in NAOT app and library scenarios

jonsequitur:
This pull request introduces several improvements and bug fixes to how `RootCommand` determines the executable name and path, especially in advanced hosting scenarios like NativeAOT native libraries. It adds comprehensive tests to ensure correct behavior, addresses edge cases with command-line parsing, and improves packaging for build targets.

**Executable Name/Path Resolution Improvements:**

* Refactored `RootCommand.ExecutableName` and `RootCommand.ExecutablePath` to more robustly determine the executable name and path, including fallback to an `AppContext` value injected by build targets when no command line arguments are available (e.g., in NativeAOT shared library scenarios) (`src/System.CommandLine/RootCommand.cs`).
* Added the `System.CommandLine.targets` file to the NuGet package to ensure the executable-name module initializer is available for all consumers, including test projects and native libraries (`src/System.CommandLine/System.CommandLine.csproj`).

**NativeAOT and Advanced Hosting Test Coverage:**

* Added a new test project (`NativeLibrary`) that builds as a NativeAOT shared library, exporting a native symbol to validate that `RootCommand.Name` falls back to the correct value when `Environment.GetCommandLineArgs()` is empty (`src/System.CommandLine.Tests/TestApps/NativeLibrary/NativeLibrary.csproj`, `src/System.CommandLine.Tests/TestApps/NativeLibrary/Program.cs`). [[1]](diffhunk://#diff-1727a8b69294450cf416e559d0f86a9c570f84d704e5fd4c897998c0c0f6fbcfR1-R32) [[2]](diffhunk://#diff-99a44b65ca1c4f998dd529aab8c87bdd982de886b702efc8e1b43eedad9400d8R1-R32)
* Introduced an end-to-end test in `CompilationTests.cs` to verify the fallback logic for `RootCommand.Name` in native library hosting scenarios, including platform-specific handling and dynamic invocation of the exported native symbol (`src/System.CommandLine.Tests/CompilationTests.cs`).

**Command-Line Parsing Bug Fixes and Tests:**

[...]

PR #2835, opened 2026-07-08
Expect NativeAOT lib prefix on Unix for .NET 11 SDK

baronfel:
Follow-up to #2820, addressing @akoeplinger's review feedback.

Starting with the .NET 11 SDK, NativeAOT applies the `lib` prefix by default to native library outputs on Unix (dotnet/docs#52324). This repo pins an 11.x SDK via `global.json`, so the published shared library is `libNativeLibrary.so` / `libNativeLibrary.dylib`. The `CompilationTests` helper still looked for `NativeLibrary.so` / `.dylib`, which would fail `File.Exists` on Linux/macOS CI.

This updates the expected native library filename to include the `lib` prefix on Unix.

### DIFF
--- a/src/System.CommandLine.Tests/CompilationTests.cs
+++ b/src/System.CommandLine.Tests/CompilationTests.cs
@@ -121,8 +121,11 @@ public class CompilationTests
 
     private static string NativeLibraryFileName(string assemblyName) =>
         OperatingSystem.IsWindows() ? $"{assemblyName}.dll"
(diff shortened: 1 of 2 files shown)


akoeplinger (approved):

akoeplinger:
@baronfel this failed because the library project still targeted NetMinimum which is net10.0 currently so we still got the old behavior. added an opt-in property so this has the new behavior.

baronfel:
I can't approve so we need to get another reviewer.

ViktorHofer (approved):

PR #2838, opened 2026-07-15
Log cancellation exceptions in the default handler

### LINKED ISSUE #2808: System.CommandLine silently swallows OperationCanceledException and TaskCanceledException
### Repro

```cs
#:package System.CommandLine@2.0.8
using System.CommandLine;

// This stack trace won't print to the console
Func<ParseResult, int> throwException = parseResult =>
{
    throw new OperationCanceledException();
    // Also doesn't work:
    // throw new TaskCanceledException();
};

// This works as expected.
// Func<ParseResult, int> throwException = parseResult => { throw new Exception(); };

var root = new RootCommand();
root.SetAction(throwException);
var parseResult = root.Parse(args);
return parseResult.Invoke();
// still returns exit code 1
```

`dotnet run Repro.cs`

### Expected behavior

I see the stack trace for the `OperationCanceledException`.

### Actual behavior

There is no output.

### Additional details

As a developer trying to debug why my application crashed with no output, this behavior is frustrating.
If I wanted to exit gracefully (e.g. if the user pressed Ctrl-C), then I would expect to have to catch an exception at some point, instead of System.CommandLine swallowing it for me.

w3lld1:
## Summary

- I report `OperationCanceledException` through the default exception handler instead of suppressing its diagnostics.
- I cover both `OperationCanceledException` and `TaskCanceledException` with regression tests.
- I preserve the existing exit code of 1.

## Validation

- `./build.sh --test`
- `git diff --check upstream/main...HEAD`

Fixes #2808

### DIFF
--- a/src/System.CommandLine.Tests/UseExceptionHandlerTests.cs
+++ b/src/System.CommandLine.Tests/UseExceptionHandlerTests.cs
@@ -35,19 +35,23 @@ namespace System.CommandLine.Tests
             error.ToString().Should().Contain("System.Exception: oops!");
         }
 
-        [Fact]
-        public async Task When_thrown_exception_is_from_cancelation_no_output_is_generated()
+        [Theory]
+        [InlineData(false)]
+        [InlineData(true)]
+        public async Task UseExceptionHandler_writes_cancellation_exception_details_to_standard_error(bool taskCanceled)
         {
             Command command = new("the-command");
-            command.SetAction((_, __) => throw new OperationCanceledException());
+            Exception exception = taskCanceled
+                                      ? new TaskCanceledException("task canceled")
+                                      : new OperationCanceledException("operation canceled");
+            command.SetAction((_, __) => throw exception);
 
-            var output = new StringWriter();
             var error = new StringWriter();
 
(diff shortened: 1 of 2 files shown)


w3lld1:
> @w3lld1 please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @dotnet-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @dotnet-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @dotnet-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

@dotnet-policy-service agree

